PR #12, opened 2024-08-28
feat: aila categoriser feature with chat ID and user ID

stefl:
## Description

- Removes a pending TODO about moving the initial lesson categorisation step into a feature
- When the user starts a chat, and there is no lesson plan yet set, we scan the initial set of messages until we are able to categorise the input as describing what lesson the user wants to create
- In this step we want to make sure we have the chat ID and the user ID so that we can track LLM token usage
- We also want to be able to mock this step out so that we don't hit OpenAI in tests
- This introduces an AilaCategorisation feature which is mockable with an example mock service and an example test


packages/core/src/models/lessonPlans.ts
```diff
@@ -65,7 +65,7 @@ export class LessonPlans {
   private _prisma: PrismaClientWithAccelerate;
   constructor(private readonly prisma: PrismaClientWithAccelerate) {
     this._prisma = prisma;
-    this._rag = new RAG(this._prisma);
+    this._rag = new RAG(this._prisma, { chatId: "none" });
```
stefl, line 68: Models are mainly used for ingest, so we don't have a chatId or a userId

packages/aila/src/core/Aila.test.ts
```diff
@@ -285,4 +286,37 @@ describe("Aila", () => {
       expect(ailaInstance.lesson.plan.title).toBe(newTitle);
     }, 20000);
   });
+
+  describe("categorisation", () => {
```
stefl, line 290: Just adding one example test here to show how we can pass in the mock categoriser during Aila initialisation

packages/aila/src/core/AilaServices.ts
```diff
@@ -23,6 +23,7 @@ export interface AilaLessonService {
   readonly hasSetInitialState: boolean;
   applyPatches(patches: string): void;
   initialise(plan: LooseLessonPlan): void;
+  setUpInitialLessonPlan(messages: Message[]): Promise<void>;
```
stefl, line 26: Moves the initial lesson plan setup to the Aila Lesson which makes more sense - resolves the TODO in the Aila class

packages/aila/src/core/Aila.ts
```diff
@@ -53,9 +57,21 @@ export class Aila implements AilaServices {
       promptBuilder: options.promptBuilder,
     });
 
-    this._lesson = new AilaLesson({ lessonPlan: options.lessonPlan ?? {} });
     this._prisma = options.prisma ?? globalPrisma;
 
+    this._lesson = new AilaLesson({
+      aila: this,
```
mantagen, line 63: Have seen some fiddly bugs arise from passing `aila: this` as a prop, as the types aren't necessarily accurate (due to a quasi race condition on instantiation)

mantagen: Not saying don't do this, just highlighting because I think you were away when it was discovered

packages/core/src/models/lessonPlans.ts
mantagen: I wondered about having a special string for this. We might want to do same for userId at some point? Perhaps not

mantagen (approved):
Looks good, couple queries

packages/aila/src/core/Aila.ts
stefl: Okay that's interesting! I see. So if you've not yet set some property on the instance before you passing it, then it could cause a problem.

packages/core/src/models/lessonPlans.ts
stefl: Yes we might want to also store something for the user ID in non-user calls. I'm expecting we look through the database at some point to see where we are missing what we need, and that "none" should mean we can see those rows. I had to pass a string here because the underlying method now throws if chatId is not supplied.

PR #15, opened 2024-08-28
fix: remove workspace colour overrides

stefl:
## Description

- Removes the colour overrides in the vscode settings so that developers can choose their own individual colour profile


simonrose121 (approved):
I'm happy with this! Didn't realise everyone was using peacock.

PR #16, opened 2024-08-28
fix: only run PR workflows in PRs

simonrose121:
## Description

- Only runs the playwright job to update the pull request if the workflow is run from a pull request.
- Only runs the workflow to set the deployment URL in PRs.

## How to test

1. In this PR, the workflows should run and update the PR description and leave a comment (playwright).
2. When merged, they shouldn't run on main, production or any branches.


mantagen (approved):
LGTM

PR #18, opened 2024-08-28
feat: detect and upgrade users with missing metadata

codeincontext:
## Description

Existing users were onboarded without us recording their region, which we need to determine their demo status. This PR takes these legacy users through a quick onboarding just to record the region and demo status

It's also a good opportunity to namespace the public metadata to the `labs` prefix. Existing users will already be migrated to this structure with a script

- Metadata structure
  - Move public user metadata to `publicMetadata.labs`
  - Add `isOnboarded` to public metadata
- Middleware
  - Add `isOnboarded` and `isDemoUser` to Clerk session claims to avoid server round trip
  - Detect users without `isOnboarded` session claim and redirect to /onboarding
  - Allow routes required for onboarding if the user needs to onboard
  - Dev: Allow routes with `x-dev-preload` even if they're protected
- Onboarding backend
  - Split demo user setup to new tRPC endpoint
- Onboarding frontend
  - Extract current onboarding form to a new AcceptTermsForm component
  - Check user metadata for `isOnboarded`
    - Show AcceptTermsForm if not onboarded
    - Just update demo status if already onboarded 
  - Regenerate token on completion to update claims

## How to test

1. Create or modify a test user with the following public metadata: `{ "labs": { "isOnboarded": true } }`
2. Log in with that user at https://oak-ai-lesson-assistant-k3tc8h51r.vercel-preview.thenational.academy
3. You go through a quick onboarding state before landing on the homepage
4. You have been assigned a demo status in your public metadata
5. You can see this demo status on /aila

## Screenshots

Current onboarding form. This shouls stay the same for new users 
CleanShot 2024-08-28 at 13 40 36@2x

For legacy users without demo status

CleanShot 2024-08-28 at 13 42 34@2x

## Checklist


apps/nextjs/src/middlewares/auth.middleware.ts
```diff
@@ -38,26 +47,89 @@ const publicRoutes = [
...
+// This allows us to warm up the chat server with live reload
+// So we get something closer to the live experience
+// Without having to wait for each page to compile as we navigate
+const isPreloadableRoute = createRouteMatcher([
+  ...publicRoutes,
+  "/api/chat",
+  "/aila",
+]);
+
+const onboardingUrl = "/onboarding";
```
mantagen, line 61: v minor but feels like this doesn't need its own variable?

apps/nextjs/src/app/onboarding/onboarding.tsx
```diff
@@ -1,128 +1,46 @@
...
-      });
-
-      if (!response?.acceptedTermsOfUse) {
-        throw new Error("Could not accept terms of use");
+  const { user } = useUser();
+  const reloadSession = useReloadSession();
+  const setDemoStatus = trpc.auth.setDemoStatus.useMutation();
+
+  const userHasAlreadyAcceptedTerms =
+    user?.publicMetadata?.["labs"]?.["isOnboarded"];
```
mantagen, line 19: is this a TS complaint that the keys might not exist?

apps/nextjs/src/app/onboarding/onboarding.tsx
```diff
@@ -1,128 +1,46 @@
...
+
+  // Edge case: Legacy users have already accepted terms but don't have a demo status
```
mantagen, line 21: Sorry being a bit slow here -- where's the condition that they don't have a demo status?

apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const acceptTerms = trpc.auth.acceptTerms.useMutation({});
+
+  const handleAcceptTermsOfUse = async () => {
+    try {
+      await setDemoStatus.mutateAsync();
+      logger.debug("Demo status set successfully");
+
+      const response = await acceptTerms.mutateAsync({
+        termsOfUse: new Date(),
+        privacyPolicy: privacyAcceptedLocal ? new Date() : false,
```
mantagen, line 42: Do we want these two `accepted_at` Dates to be identical? Can't think of a reason when it would matter but feels like it would make sense

apps/nextjs/src/app/aila/[id]/share/page.tsx
```diff
@@ -25,8 +26,10 @@ export async function generateMetadata({
 }
 
 function userCanShare(user: User) {
-  const isDemoUser = Boolean(user.publicMetadata.isDemoUser ?? "true");
-
+  if (!demoUsers.isDemoStatusSet(user)) {
+    return false;
+  }
+  const isDemoUser = Boolean(user.publicMetadata.labs.isDemoUser ?? "true");
```
Comment (stefl) [inline apps/nextjs/src/app/aila/[id]/share/page.tsx:32]: does this need to be publicMetadata.labs?.isDemoUser or is this safe at this point?

packages/api/src/router/auth.ts
```diff
@@ -47,43 +47,61 @@ export const authRouter = router({
...
           updatedUser.privateMetadata;
 
-        return { acceptedPrivacyPolicy, acceptedTermsOfUse, isDemoUser };
+        return { acceptedPrivacyPolicy, acceptedTermsOfUse };
       }
     }),
+
+  setDemoStatus: protectedProcedure.mutation(async ({ ctx }) => {
+    const { userId } = ctx.auth;
+    if (typeof userId === "string") {
```
mantagen, line 73: is it not always a string at this point in a protectedProcedure?

apps/nextjs/src/app/onboarding/onboarding.tsx
```diff
@@ -1,128 +1,46 @@
...
+  useEffect(() => {
+    async function handleDemoStatusSet() {
+      if (userHasAlreadyAcceptedTerms && !isHandlingLegacyCase.current) {
+        isHandlingLegacyCase.current = true;
+        logger.debug("User has already accepted terms");
+        await setDemoStatus.mutateAsync();
+        logger.debug("Demo status set successfully");
+        await reloadSession();
+        logger.debug("Session token refreshed successfully. Redirecting");
+        window.location.href = "/";
```
stefl, line 32: When we're doing redirects elsewhere I've been adding something like ?reason=<some string> so you can see why the redirect is happening. Would that be helpful here?

apps/nextjs/src/app/onboarding/page.tsx
```diff
@@ -1,5 +1,5 @@
 import OnBoarding from "./onboarding";
 
-export default function OnBoardingPage() {
+export default async function OnBoardingPage() {
```
stefl, line 3: Does this need an explicit "use server" so it's clear it's a server component?

apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+      if (!response?.acceptedTermsOfUse) {
+        throw new Error("Could not accept terms of use");
+      }
+      logger.debug("Terms of use accepted successfully.");
+
+      await reloadSession();
+      logger.debug("Session token refreshed successfully. Redirecting");
+
+      window.location.href = "/";
```
stefl, line 53: Again here, would a ?redirect=<reason> be helpful at all?

apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <OakBox $pt="inner-padding-m" $mv="space-between-s">
+          <CheckBox
+            label="Accept"
+            setValue={setPrivacyAcceptedLocal}
+            size="base"
+          >
+            <OakSpan>
+              Keep me updated with latest Oak AI experiments, resources and
+              other helpful content by email. You can unsubscribe at any time.
+              See our{" "}
```
stefl, line 88: The language here shifts from "me" - the user to "our" - Oak. Should this be "the"?

apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </OakLink>
+              .
+            </OakSpan>
+          </CheckBox>
+        </OakBox>
+
+        {termsAcceptedLocal ? (
+          <OakFlex $flexDirection="column" $gap="all-spacing-7">
+            <p>
+              Terms accepted, if the page does not reload please refresh and
```
stefl, line 100: Terms accepted. If...

apps/nextjs/src/hooks/useClerkDemoMetadata.ts
```diff
@@ -43,7 +68,7 @@ function getResult(
   // User is logged in and has completed onboarding
   return {
     isSet: true,
-    userType: Boolean(user.user.publicMetadata.isDemoUser) ? "Demo" : "Full",
+    userType: user.user.publicMetadata.labs.isDemoUser ? "Demo" : "Full",
```
stefl, line 71: Do we always have .labs here?

packages/core/src/models/demoUsers.ts
```diff
@@ -44,14 +61,19 @@ class DemoUsers {
     return { region, isDemoRegion };
   }
 
+  isDemoStatusSet(user: LabsUser): user is UserWithDemoStatus {
```
mantagen, line 64: No matter at all for this but I think this could be `user is LabsUser  & UserWithDemoStatus`. Probably less clear and unnecessary but just thought I'd share as at first I wasn't sure whether TS would infer that LabsUser part ... but it doesn't

apps/nextjs/src/app/onboarding/onboarding.tsx
mantagen: Ah just seen -- user won't land on this page otherwise

mantagen (approved):
LGTM! Few comments

stefl (approved):
Just a couple of questions and spotted some copy issues

apps/nextjs/src/app/aila/[id]/share/page.tsx
Comment (codeincontext) [reply apps/nextjs/src/app/aila/[id]/share/page.tsx:32]: The type of `isDemoStatusSet` is

```ts
  isDemoStatusSet(user: LabsUser): user is UserWithDemoStatus
 ```

So that's already asserted here!

File: apps/nextjs/src/app/onboarding/onboarding.tsx (hunk 3 above)
Comment (codeincontext) [reply apps/nextjs/src/app/onboarding/onboarding.tsx:21]: The case is any user who signed up before started capturing the demo status. Unless we can find a sensible representation of their region (maybe posthog?), our only option is to wait until they come back online to set it

We're soon going to have users landing on labs from OWA who don't have all of our metadata, so this concept of upgrading metadata will likely come in useful anyway

File: apps/nextjs/src/app/onboarding/onboarding.tsx (hunk 7 above)
Comment (codeincontext) [reply apps/nextjs/src/app/onboarding/onboarding.tsx:32]: This line is copied form what we're currently doing, but I like that idea and will add it

File: apps/nextjs/src/app/onboarding/page.tsx (hunk 8 above)
Comment (codeincontext) [reply apps/nextjs/src/app/onboarding/page.tsx:3]: Ah, I was originally doing the demo metadata upgrade in the server component here but we didn't really have the precedent for making calls there. I'll remove the async declaration and leave the file as-is

File: apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx (hunk 4 above)
Comment (codeincontext) [reply apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx:42]: Do you mean an identical instance, or do you mean that they should both always be set? `privacyPolicy` is a slight misnomer as it means "I want to receive marketing comms and have accepted the related privacy policy"

File: apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx (hunk 9 above)
Comment (codeincontext) [reply apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx]: Done

File: apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx (hunk 10 above)
Comment (codeincontext) [reply apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx:88]: This isn't new copy, but you are right that the grammar isn't quite right

The previous sentence has already switched to "You", so I don't think it's as bad as it looks at first glance

That said, "Keep me updated **with latest** Oak AI experiments" seems to be missing a "the"?

File: apps/nextjs/src/hooks/useClerkDemoMetadata.ts (hunk 12 above)
Comment (codeincontext) [reply apps/nextjs/src/hooks/useClerkDemoMetadata.ts:71]: We do after calling `isDemoStatusSet`. If you think that behaviour is a bit unclear and tucked away, I can change it

Typescript knows that it's an optional key on the metadata

File: apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx (hunk 4 above)
Comment (mantagen) [reply apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx:42]: Sorry i guess I meant the same instance (i.e. consent was given at exactly the same time)

File: apps/nextjs/src/middlewares/auth.middleware.ts (hunk 1 above)
Comment (codeincontext) [reply apps/nextjs/src/middlewares/auth.middleware.ts:61]: That makes sense. This originated from a clerk example

File: packages/api/src/router/auth.ts (hunk 6 above)
Comment (codeincontext) [reply packages/api/src/router/auth.ts:73]: Good point. Maybe it came as the result of a refactor. Will fix

File: packages/core/src/models/demoUsers.ts (hunk 13 above)
Comment (codeincontext) [reply packages/core/src/models/demoUsers.ts:64]: Great observation. I think this is actually an issue with the UserWithDemoStatus type, which I've now fixed

File: apps/nextjs/src/app/aila/[id]/share/page.tsx (hunk 5 above)
Comment (stefl) [reply apps/nextjs/src/app/aila/[id]/share/page.tsx:32]: I guess my concern is that we might be trusting what is coming through here, and it might be correct in the types, but I'm not sure we validate the actual data from Clerk?

File: apps/nextjs/src/hooks/useClerkDemoMetadata.ts (hunk 12 above)
Comment (stefl) [reply apps/nextjs/src/hooks/useClerkDemoMetadata.ts:71]: Okay great

File: apps/nextjs/src/app/aila/[id]/share/page.tsx (hunk 5 above)
Comment (codeincontext) [reply apps/nextjs/src/app/aila/[id]/share/page.tsx:32]: Do you mean that isDemoStatusSet could be accidentally sanitising incorrect data with a correct type?


Here's the definition:
```ts
  isDemoStatusSet(user: LabsUser): user is UserWithDemoStatus {
    const labsMetadata = user.publicMetadata.labs || {};
    return "isDemoUser" in labsMetadata && labsMetadata.isDemoUser !== null;
  }
```

apps/nextjs/src/components/Onboarding/AcceptTermsForm.tsx
codeincontext: Ok. Will update this

apps/nextjs/src/app/onboarding/onboarding.tsx
codeincontext: Yeah. It's a slightly hokey way to wait for `user` to have loaded and then dig into those fields if they don't exist. `publicMetadata` is `unknown` here, so we have to access them as fields rather than properties

PR #19, opened 2024-08-28
feat: prompt versioning

stefl:
## Description

- Adds a support for versioning the main Aila prompt
- Requires all devs to run `pnpm prompts:dev` to create rows in the prompts table in the database
- When we make any AilaGeneration we store the current version of the prompt against the generation
- Updates the `pnpm prompts` script to generate rows that relate to the Ail prompt at deployment
- Has different variants for the three variables that we pass into the prompt - useRag, interactive/generate, basedOn, since we construct different prompts in each scenario
- Because of issues getting core to pull code in about the JSON schemas from the aila package, it does not take into account if there are any changes to the lesson schema itself. We could add this at a later stage.
- Refactors the prompt into parts (in line with how the other prompts work)
- Since the format of the prompt is so different, it stores the prompt parts as a JSON array rather than in the structure the other prompts are in


packages/aila/src/features/generation/AilaGeneration.ts
```diff
@@ -17,30 +18,35 @@ export class AilaGeneration {
   private _completionTokens: number = 0;
   private _totalTokens: number = 0;
   private _systemPrompt: string = "";
-  private _promptId: string = "clnnbmzso0000vgtj13dydvs7"; // #TODO fake the prompt id
+  private _promptId: string | null = null;
```
stefl, line 21: Another TODO down!

mantagen: 😅

packages/core/src/models/promptVariants.ts
```diff
@@ -24,25 +24,26 @@ export class PromptVariants {
...
-    const hash = Md5.hashStr(template);
+  async setCurrent(variant: string, raw = false) {
+    const template = this.format(raw);
+    const { slug } = this.definition;
+    const hash = `${slug}-${variant}-${Md5.hashStr(template)}`;
+    console.log("*** setCurrent ***");
+    console.log("Hash", hash);
+    console.log("Slug", slug);
+    console.log("Variant", variant);
+    console.log("-----");
```
mantagen, line 35: Are you wanting to keep these in?

packages/aila/src/features/generation/AilaGeneration.ts
```diff
@@ -127,4 +142,42 @@ export class AilaGeneration {
...
+
+    // This key is defined in the setupPrompts script in core
+    const variantSlug = `${responseMode}-${basedOn ? "basedOn" : "notBasedOn"}-${useRag ? "rag" : "noRag"}`;
+
+    let prompt: Prompt | null = null;
+    prompt = await prisma.prompt.findFirst({
+      where: {
+        variant: variantSlug,
+        appId: appSlug,
+        slug: promptSlug,
```
mantagen, line 161: Could add an index on these columns on the prompts table?

packages/core/src/models/promptVariants.ts
```diff
@@ -24,25 +24,26 @@ export class PromptVariants {
...
     const existing = await this.prisma.prompt.findFirst({
       where: {
-        hash,
+        AND: [{ hash }, { slug }, { variant }],
```
mantagen, line 38: Perhams an index on these?

mantagen: Though ... if we're caching now ...

packages/core/src/prompts/lesson-assistant/old.ts
```diff
@@ -0,0 +1,742 @@
+// This is the original version of the prompt before
```
mantagen, line 1: When to delete?

packages/core/src/scripts/setupPrompts.ts
```diff
@@ -2,20 +2,30 @@ import { prisma } from "@oakai/db";
...
+    }
+    console.log("Lesson Planner");
     for (const k of Object.keys(lessonPlannerPrompts)) {
       const prompt =
         lessonPlannerPrompts[k as keyof typeof lessonPlannerPrompts];
       const prompts = new PromptVariants(prisma, prompt, "main");
-      await prompts.setCurrent();
+      await prompts.setCurrent("main");
     }
+    console.log("Quiz Generator");
```
mantagen, line 23: can remove

mantagen: and other logs in this file?

packages/aila/src/features/generation/AilaGeneration.ts
```diff
@@ -127,4 +142,42 @@ export class AilaGeneration {
...
+
+  private async fetchPromptId(): Promise<string | null> {
+    const appSlug = "lesson-planner";
+    const promptSlug = "generate-lesson-plan";
+    const responseMode = "interactive";
+    const basedOn = !!this._aila.lesson.plan.basedOn;
+    const useRag = this._aila.options.useRag ?? true;
+
+    // This key is defined in the setupPrompts script in core
+    const variantSlug = `${responseMode}-${basedOn ? "basedOn" : "notBasedOn"}-${useRag ? "rag" : "noRag"}`;
```
mantagen, line 154: Is there a way to share a slug generation fn?

packages/core/src/models/promptVariants.ts
```diff
@@ -90,24 +94,29 @@ export class PromptVariants {
     });
   }
 
-  format() {
+  format(raw = false) {
```
mantagen, line 97: Might be helpful to have a type annotation here, possible passing arg as an object? Thinking that seeing the invocation somewhere `format(true)` isn't clear

mantagen (approved):
Code looks good! Haven't tested locally.

Couple comments/suggestions.

In future we might want to look at using a templating language so its a little more standardised. But def not for now! 😅

packages/core/src/models/promptVariants.ts
stefl: It's quite a small table - not sure it's worth the hassle of doing a migration! And yes - I'll be adding the prompt table to the Accelerate cache

packages/core/src/models/promptVariants.ts
stefl: Removing this but it's useful to know if it is inserting when you run the script. I'll add a line only if it inserts

packages/core/src/prompts/lesson-assistant/old.ts
stefl: No, let's keep because there is a test on the diff between this prompt and the newly split up one. I'd like that in git history.

packages/core/src/scripts/setupPrompts.ts
stefl: This gives you some feedback when you run pnpm prompts:dev, so you know what it's doing. I think that's helpful?

codeincontext (approved):
Yes, this looks great. Easier to follow than the previous version, so kudos for that

PR #21, opened 2024-08-28
feat: add users to Hubspot on creation

codeincontext:
## Description

We need to send users to hubspot to be able to send them communications and track their email preferences

- Add hubspot client to core package
- Call hubspot client with user details when onboarding

Once this work is deployed I will make sure that existing users are up to date in Hubspot. I already have a script to export preferences to a CSV. If that's not useful for the team, I can adapt it to use this client

## How to test

1. Create a new user with a first and last name. Domain: https://oak-ai-lesson-assistant-j28vnq7f4.vercel-preview.thenational.academy
2. Submit the onboarding form
3. Check [hubspot (staging)](https://app-eu1.hubspot.com/contacts/24947711/objects/0-1/views/all/list)
  1. That the first and last name are there
  2. That the marketing preference is recorded: Go to "Actions" => "View all Properties" => Search for "email_consent_on_ai"

Test with the checkbox both enabled and disabled
Test with an email that already exists in hubspot, and a new account

## Checklist


packages/core/src/analytics/hubspotClient.ts
```diff
@@ -0,0 +1,63 @@
...
+  marketingAccepted,
+}: CreateHubspotCustomerInput) => {
+  let id: string | undefined;
+  try {
+    const result = await hubspotClient.crm.contacts.basicApi.getById(
+      email,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
```
mantagen, line 31: I feel like they could change the signature here!

packages/core/src/analytics/hubspotClient.ts
```diff
@@ -0,0 +1,63 @@
...
+      "email",
+    );
+    id = result.id;
+  } catch (e) {
+    if (e instanceof ApiException && e.code === 404) {
```
mantagen, line 36: What's the 404 saying here?

If it means no user found then we wouldn't want to return early, right?

mantagen: Guessing it doesn't mean that if the code is successfully creating new hubspot entities

packages/core/src/analytics/hubspotClient.ts
codeincontext: Yeah, I'm actually a little unimpressed by that package in general. For example, it doesn't export their error instances

packages/core/src/analytics/hubspotClient.ts
codeincontext: Oh, you're right. Well spotted as I thought I had tested that

mantagen (approved):
Nice one!

PR #22, opened 2024-08-28
feat: add Prisma Accelerate caching to some of our queries

stefl:
## Description

- Adds Prisma Accelerate caching to some of our queries.
- These are queries on tables which are write-once or only update during the chat.
- We could see how this performs in the Prisma Accelerate dashboard and then potentially consider some other tables which might benefit from caching.

## How to test

- Use the app to ensure that we aren't seeing any strange behaviour
- Check the Prisma Accelerate dashboard - we should see a higher percentage of cached queries if this is successful

@Stef screenshotted this on 2024-08-29 at 14 24 56@2x


apps/nextjs/src/app/api/aila-download/route.ts
```diff
@@ -115,6 +115,7 @@ async function getHandler(req: Request): Promise<Response> {
       gdriveFileId: fileId,
       userId,
     },
+    cacheStrategy: { ttl: 60 * 5, swr: 60 * 2 },
```
stefl, line 118: The pattern is the to add a TTL and SWR setting to each request that looks cacheable

mantagen (approved):
LGTM! Could also add caching to the `lessonSchema` requests I think.

Keen to get a better idea of when the data could be stale, and what the impact of that would be

codeincontext (approved):

PR #25, opened 2024-08-29
feat: pass a mocked LLM Service to Aila via the new services object

stefl:
## Description

- Adds an option under services in AilaInitialisationOptions' to pass in an LLMService
- Moves the MockLLMService into the Aila package
- Adds a test to show that we are able to mock LLM responses in tests


codeincontext (approved):
This makes sense to me

PR #27, opened 2024-08-29
fix: prompt version hash

stefl:
## Description

- Fixes an issue where there could be only one current prompt for each slug, irrespective of the variant
- Ensures there is one current prompt for each variant
- Ideally, developers should use pnpm prompts:dev to set up their prompts initially, but it tries to create the prompt variant if it is not found

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "lint": "turbo lint",
     "prompts": "turbo prompts --filter=core",
+    "prompts:dev": "turbo prompts:dev --filter=core",
     "sort-package-json": "sort-package-json \"package.json\" \"packages/*/package.json\" \"apps/*/package.json\"",
     "type-check": "turbo type-check",
     "node-version": "node -v",
--- a/packages/aila/src/features/generation/AilaGeneration.ts
+++ b/packages/aila/src/features/generation/AilaGeneration.ts
@@ -189,20 +189,25 @@ export class AilaGeneration {
         }
       }
     } else {
-      prompt = await prisma.prompt.findFirst({
+      const promptQuery = {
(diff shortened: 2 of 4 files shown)


packages/core/src/models/promptVariants.ts
```diff
@@ -83,6 +83,9 @@ export class PromptVariants {
         slug: {
           equals: slug,
         },
+        variant: {
```
stefl, line 86: This was the one-liner - it was updating all variants to not be current after creating each one. I missed this in testing.

package.json
```diff
@@ -22,6 +22,7 @@
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "lint": "turbo lint",
     "prompts": "turbo prompts --filter=core",
+    "prompts:dev": "turbo prompts:dev --filter=core",
```
stefl, line 25: When you run this it should fill up your prompts table with all prompts and variants

codeincontext (approved):

packages/aila/src/features/generation/AilaGeneration.ts
```diff
@@ -189,20 +189,25 @@ export class AilaGeneration {
...
-      const prompts = new PromptVariants(prisma, ailaGenerate, promptSlug);
-      const created = await prompts.setCurrent(variantSlug, true);
-      promptId = created?.id;
+      // // If the prompt does not exist for this variant, we can try to generate it
+      try {
+        const prompts = new PromptVariants(prisma, ailaGenerate, promptSlug);
+        const created = await prompts.setCurrent(variantSlug, true);
+        promptId = created?.id;
+      } catch (e) {
+        console.error("Error creating prompt", e);
```
stefl, line 209: This alerts below with instructions to run pnpm prompts:dev

PR #28, opened 2024-08-29
chore: skip Github workflows that depend on PR ID when PR doesn't exist

codeincontext:
## Description

Two of our Github workflows are triggered by successful deployments, and then search for a PR with that SHA. They both fail if a branch is pushed before the PR is created

- Skip GH actions that rely on the PR number if it's missing


simonrose121 (approved):
Nice, LGTM!

PR #34, opened 2024-08-30
fix: replace Inngest app/generation.requested with an async worker function

codeincontext:
## Description

The quiz designer uses inngest functions. We're transitioning away from inngest and it isn't set up as a Vercel integraiton on the new Vercel project

- Extract generation inngest function to a file
- Wrap and call generation code
- Use async background pattern

## How to test

- Go to the quiz designer
- Generations should work after ~10 seconds instead of hanging


packages/core/src/workers/generations/requestGeneration.ts
```diff
@@ -0,0 +1,420 @@
...
+  } catch (e) {
+    await onFailure({
+      error: e as Error,
+      event: { data: { event: { data } } },
+      logger: baseLogger,
+    });
+  }
+}
+
+async function invoke({ data, user }: InvokeArgs) {
```
codeincontext, line 55: This function itself is unchanged

packages/core/src/workers/generations/requestGeneration.ts
```diff
@@ -0,0 +1,420 @@
...
+  error: Error;
+  event: {
+    data: {
+      event: { data: z.infer<(typeof requestGenerationSchema)["data"]> };
+    };
+  };
+  logger: Logger;
+};
+
+async function onFailure({ error, event, logger }: OnFailureArgs) {
```
codeincontext, line 395: This function is unchanged

packages/core/src/workers/generations/requestGeneration.ts
```diff
@@ -0,0 +1,420 @@
...
+  try {
+    return await invoke({ data, user });
+  } catch (e) {
+    await onFailure({
+      error: e as Error,
+      event: { data: { event: { data } } },
+      logger: baseLogger,
+    });
+  }
+}
```
codeincontext, line 53: This is new

packages/core/src/workers/generations/requestGeneration.ts
```diff
@@ -0,0 +1,446 @@
...
+   * ```
+   *
+   * See `waitUntil` documentation in
+   * [Cloudflare](https://developers.cloudflare.com/workers/runtime-apis/handlers/fetch/#contextwaituntil)
+   * and [Vercel](https://vercel.com/docs/functions/edge-middleware/middleware-api#waituntil)
+   * for more details.
+   * ```
+   */
+  pending: Promise<void>;
+};
```
codeincontext, line 60: This pattern is inspired by @upstash/ratelimit

packages/core/src/workers/generations/requestGeneration.ts
```diff
@@ -0,0 +1,446 @@
...
+        ModerationType.NOT_ENGLISH,
+        {
+          moderationMeta: { englishScores },
+        },
+      );
+      // Return early and don't process generation further
+      return;
+    }
+
+    //logger.info("Running step: Detect profanity");
```
tomwisecodes, line 212: Think this was commented out but shall we delete this while we are at it?

tomwisecodes (approved):
One small comment looks like the correct way to manage this

packages/core/src/workers/generations/requestGeneration.ts
codeincontext: Done

PR #38, opened 2024-09-02
fix: update aila chat e2e test to handle 10 stages

simonrose121:
## Description

- Fix playwright tests failing because we've reduced the stages from 12 to 10.

## How to test

1. Change the base branch to production, which should trigger all the playwright tests.
2. If it passes, then switch the base back to main before merging.


apps/nextjs/tests-e2e/tests/aila-chat.test.ts
```diff
@@ -84,14 +84,14 @@ test.describe("Authenticated", { tag: "@authenticated" }, () => {
       await page.waitForURL(/\/aila\/.+/);
       await waitForGeneration(page);
 
-      for (let i = 0; i < 12; i++) {
+      for (let i = 0; i < 10; i++) {
```
simonrose121, line 87: Will tidy this up if it works.

PR #39, opened 2024-09-02
fix: large chat history

mantagen:
## Description

- Fixes chat history for users with a long chat history
- Cause: https://www.prisma.io/docs/accelerate/limitations#response-size-limit
  - 5mb limit was being exceeded because so much data (every chat 'output') was being retrieved from database

https://www.notion.so/oaknationalacademy/Fix-s-chat-history-7ea2831b3b954a10ab38a3d3b0f9a72b?pvs=4

## How to test

- history should still work
- ... once merged @hannah-bethclark should check history


simonrose121 (approved):
LGTM!

PR #61, opened 2024-09-04
feat: launch

tomwisecodes:
This removes the pages for comparative judgment and the image alt text generator.

This also removes the feature flag opening aila up to the public and launching the project 🫨


apps/nextjs/src/app/home-page.tsx
```diff
@@ -162,26 +160,19 @@ export default function HomePage({ featureFlag }) {
                 resources.
               </OakP>
 
-              {!userId || !featureFlag ? (
```
codeincontext, line 165: This looks like a functional change if the userId doesn't immediately load. I don't know if that can happen

apps/nextjs/src/components/Footer.tsx
```diff
@@ -73,9 +70,6 @@ const Footer = ({ featureFlag }: FooterProps) => {
                           }
                         }}
                         href={tool.href}
-                        disabled={
```
codeincontext, line 76: What about `!tool.href`? Is href a required field now?

apps/nextjs/src/app/home-page.tsx
tomwisecodes: This was to do with creating a button saying coming soon or not weather the user was logged in

apps/nextjs/src/components/Footer.tsx
tomwisecodes: The types allowed it to build so I assumed this was fine

PR #70, opened 2024-09-04
fix(share): fix share bug

mantagen:
## Description

- adds `isShared` to Aila package, so that it persists, fixing the share 404 bug
- adds authorisation step in `chatHandler()`

## Issue(s)

https://www.notion.so/oaknationalacademy/Share-Page-404-dd392621286e40279204d563960f2e75?pvs=4

## How to test

1. Go to the site
2. Create a lesson
3. Share it
4. Go back to lesson and change it
5. Check the share page ... should still be shared!


apps/nextjs/scripts/aila-cli.ts
```diff
@@ -12,7 +12,7 @@ const cliPlugin: AilaPlugin = {
 };
 
 const aila = new Aila({
-  chat: { id: "my chat id", userId: undefined },
+  chat: { id: "my chat id", userId: undefined, isShared: false },
```
stefl, line 15: Looks good. Maybe we should be storing the date it was shared?

apps/nextjs/src/app/api/chat/route.test.ts
```diff
@@ -1,21 +1,40 @@
...
+};
+
 describe("Chat API Route", () => {
   let testConfig: Config;
   let mockLLMService: MockLLMService;
   let mockChatCategoriser: MockCategoriser;
   beforeEach(() => {
     mockTracer.reset();
     jest.clearAllMocks();
+    jest
```
stefl, line 35: Not sure about this - if we don't want persistence in this test, we should pass usePersistence: false

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -26,6 +26,41 @@ export async function GET() {
   return new Response("Chat API is working", { status: 200 });
 }
 
+async function authoriseAndGetChat({
```
stefl, line 29: I'm trying to move all of this kind of logic out of the controller and into the Aila library. Could we make this logic happen there instead?

tomwisecodes (approved):
LGTM

PR #74, opened 2024-09-05
chore: add lesson plan content to rag error logs

mantagen:
## Description

- adds lesson plan content to the error so we see what the content was that failed to parse


packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
```diff
@@ -23,6 +23,11 @@ export async function fetchLessonPlanContentById(
   const parsedPlan = tryWithErrorReporting(
     () => LessonPlanSchemaWhilstStreaming.parse(lessonPlanRecord.content),
     "Failed to parse lesson plan content",
+    undefined,
+    undefined,
```
mantagen, line 27: This is one of the reasons I nearly always use objects as a single argument to functions. Extending the signature beyond option args is horrible otherwise.

stefl (approved):

PR #76, opened 2024-09-05
feat: update to latest "ai" package and openai

stefl:
## Description

- We want to use Open AI's [new Structured Outputs feature](https://platform.openai.com/docs/guides/structured-outputs)
- To do that, we need to update to the latest OpenAI package
- But we were blocked from doing that because we use both it and Vercel's "ai" package and there is a difference in the streaming protocol between the two versions
- This makes a relatively minimal change to how we send text back from the /chat endpoint by formatting each chunk with `0:${JSON.stringify(chunk)}\n`
- This means we are now on latest "ai" and "openai"
- From here we could then use the Sructured Outputs generation, which in turn should remove this issue where we are seeing JSON and incorrect lesson output on the left hand side of the UI
- This leaves us with Langchain at this very old version, so we should look at updating it separately

The changes in this PR are incorporated into https://github.com/oaknational/oak-ai-lesson-assistant/pull/91 so we may not need to merge this if we approve that.


apps/nextjs/src/components/AppComponents/Chat/chat-list.tsx
```diff
@@ -81,6 +86,35 @@ export function ChatList({
...
+        ailaStreamingStatus={ailaStreamingStatus}
+      />
+
+      {isDemoLocked && <DemoLimitMessage id={id} />}
+      <div ref={chatEndRef} />
+    </div>
+  );
+}
+
+export const ChatMessagesDisplay = ({
```
stefl, line 103: I've moved this here so that when you use the React Developer Tools plugin you can see the messages that are being rendered

packages/core/src/llm/helicone.ts
```diff
@@ -0,0 +1,45 @@
+export type HeliconeChatMeta = {
```
stefl, line 1: Move the helicone headers into a separate file

packages/core/src/llm/openai.ts
```diff
@@ -71,6 +31,16 @@ function createOpenAIClient(props: CreateOpenAIClientProps) {
   return new OpenAI(openAiFields);
 }
 
+function createOpenAIProvider(props: CreateOpenAIClientProps): OpenAIProvider {
```
stefl, line 34: "ai" now give us an OpenAIProvider, so we export both this and the original OpenAI

packages/aila/src/core/llm/OpenAIService.ts
```diff
@@ -1,50 +1,58 @@
...
-      // Handle the ChatCompletion case here if necessary
-      throw new Error("Received ChatCompletion instead of a Stream.");
-    } else {
-      throw new Error("Failed to create chat completion stream.");
-    }
+
+    return stream.getReader();
+  }
+
+  async createChatCompletionObjectStream(params: {
```
stefl, line 39: This would allow us to stream an object as well as text

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -179,7 +179,14 @@ async function generateChatStream(
     async () => {
       invariant(aila, "Aila instance is required");
       const result = await aila.generate({ abortController });
-      return result;
+      const transformStream = new TransformStream({
+        transform(chunk, controller) {
+          const formattedChunk = `0:${JSON.stringify(chunk)}\n`;
```
mantagen, line 184: Does `0:` here mean it's the one and only of potentially multiple simultaneous streams?

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -90,13 +91,12 @@ export class AilaChat implements AilaChatService {
...
 
-  public async appendChunk(value?: Uint8Array) {
+  public async appendChunk(value?: string) {
     invariant(this._chunks, "Chunks not initialised");
     if (!value) {
       return;
     }
-    const decoded = new TextDecoder().decode(value);
-    this._chunks.push(decoded);
+    this._chunks.push(value);
```
mantagen, line 99: nicer

mantagen (approved):
Code looks good! I would say let's give this a thorough test _on this branch_ before merging.

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: No the stream can handle different data sources simultaneously. So if we wanted different async writers to push chunks to the stream, then we could do with a different prefix. I think the prefix of 0 is for raw text.

stefl:
This is superceded by https://github.com/oaknational/oak-ai-lesson-assistant/pull/91 which includes everything in this PR

PR #78, opened 2024-09-05
fix: mobile download buttons were blocked by demo account banner

tomwisecodes:
## Issue(s)

The download and share bar were dropped by the demo account banner for users on demo accounts.

Fixes #

## How to test

1. Go to a chat in development on mobile
2. You should be able to access the share and download buttons

### DIFF
--- a/apps/nextjs/src/components/ContextProviders/Demo.tsx
+++ b/apps/nextjs/src/components/ContextProviders/Demo.tsx
@@ -51,14 +51,14 @@ export function DemoProvider({ children }: Readonly<DemoProviderProps>) {
     () =>
       isDemoUser
         ? {
-            isDemoUser,
+            isDemoUser: true,
             appSessionsRemaining,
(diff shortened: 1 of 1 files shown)


apps/nextjs/src/components/ContextProviders/Demo.tsx
```diff
@@ -51,14 +51,14 @@ export function DemoProvider({ children }: Readonly<DemoProviderProps>) {
...
-            isDemoUser,
+            isDemoUser: true,
             appSessionsRemaining,
             appSessionsPerMonth: DEMO_APP_SESSIONS_PER_30D,
             contactHref: "mailto:help@thenational.academy",
             isSharingEnabled,
           }
         : {
-            isDemoUser,
+            isDemoUser: true,
```
mantagen, line 61: But this is setting `isDemoUser` true when it's not?

tomwisecodes: wow this is the wrong file !

PR #79, opened 2024-09-05
fix: downgrade moderation if all categories triggered at same time

codeincontext:
Sometimes the moderation prompt returns all categories for a benign "sensitive" input. Until we can address the root cause, detect this case

CleanShot 2024-09-05 at 12 53 44@2x


mantagen (approved):
LGTM!

This assumes no lesson plan will be so toxic to include all. I think this is a fair assumption, but perhaps we should have an alert when this does happen so we can manually verify.

codeincontext:
> LGTM!
> 
> This assumes no lesson plan will be so toxic to include all. I think this is a fair assumption, but perhaps we should have an alert when this does happen so we can manually verify.

Sorry, I missed this message. That's a good idea. Let's look at it as a follow up

PR #89, opened 2024-09-05
feat give user feedback when download email is sent

tomwisecodes:
## Description

Currently on mobile if a user requests an email of the google docs there is no feedback that it sent. This PR changes that so that there is a fix if it has sent. Other wise allows the user to retry after delivering them an error.

### DIFF
--- a/apps/nextjs/src/components/DownloadButton.tsx
+++ b/apps/nextjs/src/components/DownloadButton.tsx
@@ -46,8 +46,11 @@ export const DownloadButton = ({
 
   const { icon, ext, analyticsResourceType } = getExportsConfig(exportsType);
 
-  const { mutateAsync, isLoading } =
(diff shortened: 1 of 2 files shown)


apps/nextjs/src/components/DownloadButton.tsx
```diff
@@ -46,8 +46,11 @@ export const DownloadButton = ({
 
   const { icon, ext, analyticsResourceType } = getExportsConfig(exportsType);
 
-  const { mutateAsync, isLoading } =
-    trpc.exports.sendUserExportLink.useMutation();
+  const {
+    data: sendUserExportLinkSent,
```
codeincontext, line 50: Could you try using `isSuccess` instead? That should avoid the need to convert to a boolean

codeincontext (approved):

PR #91, opened 2024-09-05
feat: structured outputs for Aila

stefl:
## Description

- Introduces structured outputs and schema definitions for the main Aila prompting process
- Alters the response to be a JSON object from the LLM rather than using record separators for each patch
- Builds on my previous PR which upgrades to the latest "ai" and "openai" packages
- Alters the concept of a message part to:
  - Include an ID
  - Include a type
  - Not append isPartial to the actual message part document
  - Instead, put the message part document under a "document" key
  - This means we don't have as much complexity around validating the types of the part documents and can reason about whether they are partial separately from the actual contents of what they represent

When we use the prompt, we have two modes.
- “generate” does a one-shot lesson plan generation and we want it to respond with a full lesson plan that meets the main lesson plan schema.
- “interactive” should send patches back where the patches meet a patch schema, with the intention of building the plan to the overall schema.

It looks like at some point we lost track of the “generate” mode, and “interactive” became the default. So when we tried to generate lesson plans in the ingest process, it was responding with patches instead of with the lesson plan structure. This is why we had some corrupted "lessonPlan" rows (which we use in RAG) that we had to delete.

So I’ve moved the parts about the lesson plan schema and the patch schema to the appropriate places in the prompt so they are included in the right modes. This potentially fixes one-shot lesson plan generation but I’ve not tested that yet.

To check in QA:
- Does it respond with the correct number of items for the various constraints still
- Does the length of the text in various places still match the spec
- Does it handle an existing lesson not generated using the new process. Eg. if it loads and nothing is displayed in the left hand side, or if you try to add a new message to an existing chat, does it go wrong?


.vscode/settings.json
```diff
@@ -163,6 +166,8 @@
       "rootPath": "packages/aila"
     }
   ],
+  "jest.runMode": "on-demand",
```
stefl, line 169: I wonder if we need this in user settings somewhere? Without this my mac grinds to a halt because on every save it seems to run the entire test suite!

apps/nextjs/package.json
```diff
@@ -107,7 +107,7 @@
     "unique-names-generator": "^4.7.1",
     "untruncate-json": "^0.0.1",
     "usehooks-ts": "^2.9.1",
-    "zod": "3.23.5",
+    "zod": "3.23.8",
```
stefl, line 110: Latest zod is a peer dependency of openai

packages/aila/src/protocol/parseMessageRow.test.ts
```diff
@@ -0,0 +1,67 @@
+import invariant from "tiny-invariant";
```
stefl, line 1: This is a test for the bug that I was working on today.

Basically, the Zod schema was too relaxed, and if you do [ZodSchema].safeParse it would return data = {} instead of data = {a validated cycle}.

Very strange and unexpected behavour. I've left a comment on the relevant type that was causing the problem, but I've had to leave it in place.

.vscode/settings.json
codeincontext: Running the whole test quite automatically doesn't sound good. Half-related to this: I think we should decide on some principles for our jest tests to keep them slim. For example I'd be quite keen to ensure that no external requests are made by code under test

.vscode/settings.json
```diff
@@ -179,5 +184,11 @@
   "yaml.schemas": {
     "file:///c%3A/Users/Simon/.vscode/extensions/atlassian.atlascode-2.8.6/resources/schemas/pipelines-schema.json": "bitbucket-pipelines.yml",
     "https://www.artillery.io/schema.json": []
+  },
+  "testing.openTesting": "neverOpen",
+  "jest.outputConfig": {
+    "revealOn": "run",
+    "revealWithFocus": "none",
+    "clearOnRun": "none"
```
codeincontext, line 192: These look like they might be personal settings rather than project ones?

apps/nextjs/src/components/AppComponents/Chat/chat-list.tsx
```diff
@@ -81,6 +86,35 @@ export function ChatList({
...
     <div className="relative flex w-full flex-col " onScroll={handleScroll}>
+      <ChatMessagesDisplay
+        messages={messages}
+        id={id}
+        lastModeration={lastModeration}
+        persistedModerations={persistedModerations}
+        ailaStreamingStatus={ailaStreamingStatus}
+      />
+
+      {isDemoLocked && <DemoLimitMessage id={id} />}
```
codeincontext, line 97: @mikeritson-oak We should test that the demo locking functionality still works as expected

apps/nextjs/src/components/AppComponents/Chat/chat-message/protocol.ts
```diff
@@ -1,100 +1,20 @@
...
+export function isModeration(
+  part: MessagePartDocument,
+): part is ModerationDocument {
   return part.type === "moderation";
 }
-export function isAccountLocked(part: MessagePart): part is ActionDocument {
+export function isAccountLocked(
+  part: MessagePartDocument,
+): part is ActionDocument {
   return part.type === "action" && part.action === "SHOW_ACCOUNT_LOCKED";
```
codeincontext, line 19: We should test account locking as well

packages/aila/src/core/llm/LLMService.ts
```diff
@@ -6,5 +8,13 @@ export interface LLMService {
...
     temperature: number;
-  }): Promise<ReadableStreamDefaultReader<Uint8Array | undefined>>;
+  }): Promise<ReadableStreamDefaultReader<string>>;
+  createChatCompletionObjectStream(params: {
+    model: string;
+    schema: ZodSchema;
+    schemaName: string;
+    messages: Message[];
+    temperature: number;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
codeincontext, line 18: Is there still an explicit any here?

packages/aila/src/core/llm/MockLLMService.ts
```diff
@@ -3,20 +3,43 @@ import { LLMService } from "./LLMService";
...
+    ReadableStreamDefaultReader<string>
   > {
-    const encoder = new TextEncoder();
     const responseChunks = this.responseChunks;
     const stream = new ReadableStream({
       async start(controller) {
         for (const chunk of responseChunks) {
-          controller.enqueue(encoder.encode(chunk));
+          controller.enqueue(chunk);
+          await new Promise((resolve) => setTimeout(resolve, 4));
```
codeincontext, line 28: I imagine there's a good reason for this delay, so it would be useful to put it in here

If the mock service is used in unit tests it would be best not to add any extra delays, so having the reason listed would be useful

packages/aila/src/core/llm/MockLLMService.ts
```diff
@@ -3,20 +3,43 @@ import { LLMService } from "./LLMService";
...
+        }
+        controller.close();
+      },
+    });
+    return stream.getReader();
+  }
+  async createChatCompletionObjectStream(): Promise<
```
codeincontext, line 35: We now have two paths for completions: with and without objects. It's not clear to me from the code I've seen so far if we're in a transition period, if we need both, etc

packages/aila/src/protocol/jsonPatchProtocol.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const result = parseMessageParts(content);
+
+    expect(result).toHaveLength(6);
+    expect(result[0]?.document?.type).toBe("patch");
+    expect(result[1]?.document?.type).toBe("patch");
+    expect(result[2]?.document?.type).toBe("text");
+    expect(result[3]?.document?.type).toBe("id");
+    expect(result[4]?.document?.type).toBe("moderation");
+    expect(result[5]?.document?.type).toBe("state");
```
codeincontext, line 27:
```suggestion
   const resultTypes = result.map(r => r?.document?[0]?.document?.type)
    expect(resultTypes).toBe(["patch", "text"...]);
```

Another way to approach this (may not be any better!)

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -116,28 +178,41 @@ export const PatchKeywords = z.object({
...
+  op: z.literal("remove"),
+  path: z.string(),
+});
+
 // Potentially remove this once stable
 export const JsonPatchAddSchema = z.object({
   op: z.literal("add"),
   path: z.string(),
-  value: z.unknown(), // TODO - Allows any value type
+  value: z.union([z.string(), z.object({}), z.number(), z.array(z.string())]), // TODO - Allows any value type
```
codeincontext, line 203: Do we still need this TODO?

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -167,6 +259,28 @@ export const PatchDocumentSchema = z.object({
   value: JsonPatchValueSchema,
 });
 
+// This is the schema that we send when requesting Structured Outputs
+// From the LLM. We are limited to not having min, max and we also
+// have constraints about ensuring each object has a type attribute
+// as its first attribute in the definition.
+// Note, that unless/until we migrate all past messages we won't be
+// able to use this to validate the patches currently in the database
+// because they would be missing a "type" attribute.
```
codeincontext, line 268: This is frustrating, but maybe not avoidable. We just need to be careful when updating the types

Maybe there's a future version where can wrap the current types with the LLM types, but it sounds like the better approach is to migrate the current data

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -282,108 +411,305 @@ export const JsonPatchDocumentJsonSchema = zodToJsonSchema(
...
+  | "prompt"
+  | "text"
+  | "action"
+  | "bad"
+  | "unknown"
+  | "id";
+
+export const MessagePartDocumentSchemaByType: {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [K in MessagePartType]: z.ZodSchema<any>;
```
codeincontext, line 458: Does `unknown` work here? That will often resolve that lint error

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -282,108 +411,305 @@ export const JsonPatchDocumentJsonSchema = zodToJsonSchema(
...
+// * A moderation messagePart
+// * An ID messagePart
+// * A state messagePart
+// (Potentially more)
+// So we split the message into rows and parse each one.
+// The message is streaming in, so we evaluate each row to see if it is
+// valid JSON. If it is not, then we assume that the row is streaming
+// and set an isPartial boolean on the resulting object.
+// This helps us not to re-process past messageParts after they have
+// fully streamed in.
```
codeincontext, line 568: It's useful to have all the jsonPatchProtocol work in the same file for the PR as it makes a useful diff. I think we should follow up with a PR to split out the schema definition from the utility functions. Comments like this are really useful and at the moment get a bit lost in the file

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -475,3 +802,8 @@ export function parseJsonSafely(jsonStr: string, logging: boolean = false) {
   // Return null if no valid JSON could be extracted
   return null;
 }
+
+// Handles parsing the message parts that are part of a message
+// Each row is separated by the record separator character
+// Rows can either be a valid document OR an object that looks like this:
+// {"type":"llmMessage","patches":[{...},{...}]},"prompt":{"type":"text","message":"Some message"}}
```
codeincontext, line 809: Should this go at the top of the file?

packages/aila/src/protocol/parseMessageRow.test.ts
codeincontext: I think it could be useful context to put this in the test file itself?

packages/aila/src/protocol/schema.ts
```diff
@@ -38,6 +38,19 @@ export const MisconceptionSchema = z.object({
     .describe("Not to be used. Only included here for legacy support."), // There is some confusion here about what the LLM generates
 });
 
+export const MisconceptionSchemaWithoutLength = z.object({
```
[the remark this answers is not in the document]
codeincontext: It would be useful to have a comment for each of these reminding us why they exist (we can't have min/max). In other places we've suffixed with `ForLLM` which does a similar thing

I wonder if it's worth writing a function to take a zod schema and replace the parts which aren't supported by the LLM automatically. That would need to be a follow up regardless

packages/core/src/llm/openai.ts
```diff
@@ -71,6 +31,16 @@ function createOpenAIClient(props: CreateOpenAIClientProps) {
   return new OpenAI(openAiFields);
 }
 
+function createOpenAIProvider(props: CreateOpenAIClientProps): OpenAIProvider {
```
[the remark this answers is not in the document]
codeincontext: It's annoying that the library is called "AI" as it makes this hard to name. What about something like `createVercelOpenAiClient`? Or provider would also be fine

packages/core/src/prompts/lesson-assistant/parts/interactive.ts
```diff
@@ -53,13 +59,9 @@ Remove one item from an array:
 Replace a value
 { "op": "replace", "path": "/misconceptions/0/misconception", "value": "A renamed misconception" }
 
-FORMATTING
+The schema of the overall lesson plan is provided below. The changes you make should match the appropriate schema definition for the section you are adding or replacing.
```
codeincontext, line 62: It says "below". I assume that `lessonPlanJsonSchema` is appended below this prompt part?

packages/aila/src/protocol/jsonPatchProtocol.ts
stefl: Good spot - I must have moved the method and forgot the comment

.vscode/settings.json
stefl: Agree! I'll move these settings to a personal file for now

packages/core/src/llm/openai.ts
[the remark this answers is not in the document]
stefl: Okay let's go with that

codeincontext (approved):

.vscode/settings.json
stefl: I've moved these to a personal file now

PR #102, opened 2024-09-09
fix(moderation prompt): lower false positives in toxic category

mantagen:
## Description

I tried a few variants locally (prompt/model combinations), and this was the best I found. (note: gpt-4o-mini is no good for this!).

- updates open-ai from 4.52.0 to 4.58.1 (allowing for json_schema mode) 
- moderation agent now uses "json_schema" mode
- adds a 'note' to the Toxic moderation category description
- amends 'instruction' setting of the prompt
- makes justification non-optional
- filter categories by parent category score

Later will explore some more in-depth changes, but that would come with more risk associated, so want to do that when there's a dataset (and more time) to give the confidence required.

## Issue(s)

https://www.notion.so/oaknationalacademy/Fix-moderation-bug-where-all-categories-get-flagged-3c93a1e819bb438f971187cb8be653bc

## How to test

1. Moderation should generally work (e.g. test some sensitive and toxic material)

### DIFF
--- a/apps/nextjs/package.json
+++ b/apps/nextjs/package.json
@@ -80,7 +80,7 @@
     "languagedetect": "^2.0.0",
     "next": "14.2.5",
     "object-hash": "^3.0.0",
-    "openai": "^4.52.0",
+    "openai": "^4.58.1",
     "p-limit": "^6.1.0",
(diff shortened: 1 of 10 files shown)


codeincontext (approved):
Makes sense to me 👍

PR #104, opened 2024-09-10
chore: back to lesson from help

tomwisecodes:
## Description

Fixes: 
https://www.notion.so/oaknationalacademy/Back-from-help-8493abd0c1dc437989f4792f485efb83

### DIFF
--- a/apps/nextjs/src/app/aila/help/index.tsx
+++ b/apps/nextjs/src/app/aila/help/index.tsx
@@ -3,6 +3,7 @@
 import { useRef } from "react";
 
 import { OakLink } from "@oaknational/oak-components";
+import { useParams } from "next/navigation";
 
 import { Header } from "@/components/AppComponents/Chat/header";
 import GetInTouchBox from "@/components/AppComponents/GetInTouchBox";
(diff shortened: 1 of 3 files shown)


apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -24,6 +25,8 @@ const Help = () => {
       window.scrollTo({ top: y, behavior: "smooth" });
     }
   };
+  // Check weather the url has a param called ailaId
```
codeincontext, line 28:
```suggestion
  // Check whether the url has a param called ailaId
```

codeincontext (changes requested):
I got a 404 on the help page. I wonder if it should use a query param rather than baking into the URL?

codeincontext:
Sidebar stuck fix

apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -94,7 +97,10 @@ const Help = () => {
 
           <div className="mb-40 max-w-[800px]">
             <div className="mb-20 mt-10 sm:hidden">
-              <OakLink href="/aila" iconName="chevron-left">
+              <OakLink
+                href={ailaId ? `/aila/${ailaId}` : `/aila`}
```
codeincontext, line 101: Is this one still out of date?

tomwisecodes: no this one is correct

codeincontext (approved):

PR #107, opened 2024-09-10
Fix/sidebar getting stuck at 0

tomwisecodes:
## Description

Fixes:
https://www.notion.so/oaknationalacademy/Sidebar-getting-stuck-at-0-10-sections-e3d486efcb334db58af9439778d0c31c


codeincontext:
I got a 404 on the help page. I wonder if it should use a query param rather than baking into the URL?
CleanShot 2024-09-10 at 13 19 42@2x

codeincontext (approved):

PR #114, opened 2024-09-11
feat: check for exsisting downloads on download page mount

tomwisecodes:
## Description

Fixes the below issue by checking if there is an existing snap shot of the lesson (and therefore a download) on page mount.

Fixes:

https://www.notion.so/oaknationalacademy/Return-To-Downloads-ebbb8afc900148a79e102da9a9510c63


mantagen (commented):
Couple comments.

I put comments on one of the hooks, but as it repeated it will apply to all

PR #132, opened 2024-09-13
feat: local db setup scripts

tomwisecodes:
## Description

This addresses issues we have had with moving data around. 

- Adds a variable seed script from a live env and a relation / null value checking script.
- Updates the readme set up for DB in line with new changes.


README.md
```diff
@@ -80,68 +79,38 @@ To run `psql`, ssh into the box using:
...
 
 ```shell
-pnpm run docker-reset
-pnpm run db-restore-from:prd
+pnpm run db-push
 ```
 
-where `:prd` can be either `:prd` or `:stg` (the Doppler environments).
-
-### With Homebrew
```
tomwisecodes, line 111: Removing this now we know the docker setup works and is clear

packages/db/scripts/import-export/tables.txt
```diff
@@ -0,0 +1,38 @@
...
+re_generations
+generation_user_flags
+shared_content
+lesson_snapshots
+lesson_exports
+qd_snapshots
+moderations
+qd_exports
+lesson_export_downloads
+qd_export_downloads
```
tomwisecodes, line 38: I did want a more dynamic way of doing this however it was so much more reliable to just have a clear ordered list.

tomwisecodes:
This could include an adjustment to the port that docker is running on as well however would need team wide buy in to avoid causing many issues.

simonrose121:
@mantagen I've picked up this PR in @tomwisecodes's absence, but I'm having issues with seeding a local schema.

We have data in a few tables in both staging and production that doesn't match the Prisma schema, so when I try to validate the data (one of Tom's additional steps) it fails. I think this is where migrations have been applied but the pre-existing data doesn't meet the new schema requirements. It's often non-null columns that have null values.

My initial idea was to reduce the tables we seed to just the necessary ones for local dev, but that's still proving to be an issue. Or we adjust the validation step to ignore non nullables, but that doesn't feel right.

What do you reckon?

simonrose121:
> @mantagen I've picked up this PR in @tomwisecodes's absence, but I'm having issues with seeding a local schema.
> 
> We have data in a few tables in both staging and production that doesn't match the Prisma schema, so when I try to validate the data (one of Tom's additional scripts) it fails. I think this is where migrations have been applied but the pre-existing data doesn't meet the new schema requirements. It's often non-null columns that have null values.
> 
> My initial idea was to reduce the tables we seed to just the necessary ones for local dev, but that's still proving to be an issue. Or we adjust the validation step to ignore non nullables, but that was the main reason Tom implemented this in the first place!
> 
> What do you reckon?

Updated to just use tables required for local dev with RAG.

mantagen (approved):
Nice1!

packages/db/package.json
```diff
@@ -9,6 +9,10 @@
...
     "db-dump:stg": "docker exec oak-ai-beta pg_dump $(doppler secrets get -c stg --plain DATABASE_ADMIN_URL | sed -e 's/?sslmode=required//g') --no-owner --exclude-table-data public.generations --exclude-table-data public.snippets --exclude-table-data public.downloads",
     "db-dump:stg:homebrew": "pg_dump $(doppler secrets get -c stg --plain DATABASE_ADMIN_URL | sed -e 's/?sslmode=required/?sslcert=./certs/db_stg.pem/g') --no-owner --exclude-table-data public.generations --exclude-table-data public.snippets --exclude-table-data public.downloads",
+    "db-seed-local-from:stg": "doppler run --config stg -- bash -c 'source scripts/import-export/db_seed.sh stg dev'",
+    "db-seed-local-from:prd": "doppler run --config prd -- bash -c 'source scripts/import-export/db_seed.sh prd dev'",
+    "db-seed-stg-from:local": "doppler run --config stg -- bash -c 'source scripts/import-export/db_seed.sh dev stg'",
+    "db-seed-prd-from:local": "doppler run --config prd -- bash -c 'source scripts/import-export/db_seed.sh dev prd'",
```
mantagen, line 15: Little bit weary about having this script just casually there!

simonrose121: Same, the original idea behind this PR was to do with pushing the locally ingested data up to live environments. If your ingest PRs actually solve this problem then we could remove those and have this functionality focusing on seeding dev environments from live?

PR #133, opened 2024-09-13
Add a toggle to enable/disable structured outputs

stefl:
## Description

- This adds NEXT_PUBLIC_STRUCTURED_OUTPUTS_ENABLED as an environment variable to switch between structured outputs or standard output

### DIFF
--- a/packages/aila/src/core/llm/OpenAIService.ts
+++ b/packages/aila/src/core/llm/OpenAIService.ts
@@ -7,6 +7,8 @@ import { ZodSchema } from "zod";
 import { Message } from "../chat";
 import { LLMService } from "./LLMService";
 
+const STRUCTURED_OUTPUTS_ENABLED =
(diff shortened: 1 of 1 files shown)


packages/core/src/prompts/lesson-assistant/parts/interactive.ts
```diff
@@ -1,12 +1,22 @@
 import { TemplateProps } from "..";
 
+const STRUCTURED_OUTPUTS_ENABLED =
+  process.env.NEXT_PUBLIC_STRUCTURED_OUTPUTS_ENABLED === "true" ? true : false;
```
stefl, line 4: I'm not so happy about this, because we aren't adding another prompt variant. But that felt like too big a change for just now.

Note that if we're not using structured outputs we need to re-add the language about sending back JSON patch documents for each change rather than a single object with patches and a user prompt.

simonrose121 (approved):
LGTM!

PR #134, opened 2024-09-14
feat(llm): o1 mini

johnrobeds:
## Description

Provides preview of o1-mini


packages/aila/src/constants.ts
```diff
@@ -1,6 +1,6 @@
 import OpenAI from "openai";
 
-export const DEFAULT_MODEL: OpenAI.Chat.ChatModel = "gpt-4o-2024-08-06";
+export const DEFAULT_MODEL: OpenAI.Chat.ChatModel = "o1-mini";
```
stefl, line 3: Does o1-mini have structured outputs?

johnrobeds: No, but... https://cookbook.openai.com/examples/o1/using_chained_calls_for_o1_structured_outputs

PR #135, opened 2024-09-15
feat: use TRPC for state and remove the spinner flash

stefl:
## Description

- If the server has any kind of failure after a message has been sent, then it will fail to send the "state" message
- This results in some odd behaviour on the front end
- We now have TRPC for loading the initial chat, so... let's get rid of the state command use TRPC to fetch the chat after each message has arrived
- This moves the TRPC call inside the ChatProvider, which can then refetch the query
- For performance we change the tempLessonPlan to use a ref instead of state, which improves performance during streaming
- Plus, by doing so we have more control and can get rid of the flash where the lessonPlan is replaced by the new state, which currently makes each section flip back to the loading state
- The end result is that the front end doesn't jump and show spinners after each message, plus we can guarantee after each message that the front end has a matching lesson plan state to what the server has

## How to test

1. Go through the chat flow
2. Keep an eye on the sections as they appear on the right
3. You should not see the 500ms spinner animation after the message to the user appears on the left
4. Reloading the chat you should see the same state
5. In the Network panel you should see TRPC calls to fetch the chat after each message has completed


apps/nextjs/src/app/aila/page-contents.tsx
```diff
@@ -2,41 +2,17 @@
...
-          key={`chat-${id}`}
-          id={id}
-          initialMessages={initialMessages}
-          initialLessonPlan={initialLessonPlan}
-          initialModerations={initialModerations}
-          startingMessage={startingMessage}
-        >
-          <Chat isShared={isShared} />
+        <ChatProvider key={`chat-${id}`} id={id}>
+          <Chat />
```
stefl, line 15: Chat now handles the TRPC calls to the server, so we don't need to pass these in at this level now.

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -45,7 +46,7 @@ const DropDownSection = ({
   useEffect(() => {
     if (value === null || value === undefined || !value) {
       setStatus("empty");
-    } else if (!valuesAreEqual(value, prevValue)) {
+    } else if (!equals(value, prevValue)) {
```
stefl, line 49: Use ramda equals instead of stringifying JSON for comparison

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -1,46 +1,46 @@
...
-} from "../AppComponents/Chat/Chat/utils";
+import { findMessageIdFromContent } from "../AppComponents/Chat/Chat/utils";
 import {
   isAccountLocked,
   isModeration,
 } from "../AppComponents/Chat/chat-message/protocol";
 
 export type ChatContextProps = {
   id: string;
+  chat: AilaPersistedChat | undefined;
```
stefl, line 43: The chat context now provides the chat

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
```diff
@@ -42,7 +42,7 @@ function patchHasBeenApplied(
 export const useTemporaryLessonPlanWithStreamingEdits = ({
   lessonPlan,
   messages,
-  isStreaming,
+  //isStreaming, // unused now?
```
stefl, line 45: I'm not sure if we need this prop now

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
```diff
@@ -55,90 +55,101 @@ export const useTemporaryLessonPlanWithStreamingEdits = ({
...
+  const appliedPatchesRef = useRef<PatchDocumentWithHash[]>([]);
 
+  // Update the ref when lessonPlan changes
   useEffect(() => {
-    if (!isStreaming && lessonPlan) {
-      if (!equals(lessonPlan, tempLessonPlan)) {
-        console.log("Reset the temp lesson plan to the lesson plan");
-        setTempLessonPlan(deepClone(lessonPlan));
-      }
+    if (lessonPlan && !equals(lessonPlan, tempLessonPlanRef.current)) {
```
stefl, line 63: Doing this means that the lessonPlan object is stable, which results in the main component tree not doing a full re-render when we change the values of keys

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
```diff
@@ -55,90 +55,101 @@ export const useTemporaryLessonPlanWithStreamingEdits = ({
   partialPatches: PatchDocument[];
 } => {
   const throttledAssistantMessages = useThrottle(messages, 100);
-  const [tempLessonPlan, setTempLessonPlan] = useState<LooseLessonPlan>({
-    ...lessonPlan,
-  });
-  const [appliedPatches, setAppliedPatches] = useState<PatchDocumentWithHash[]>(
-    [],
-  );
+  const tempLessonPlanRef = useRef<LooseLessonPlan>({});
```
stefl, line 58: Move the tempLessonPlan to a ref instead of using state

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
mantagen: If commenting doesn't complain, delete!

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -294,21 +266,20 @@ export function ChatProvider({
...
-      lessonPlanTracking.onStreamFinished({
-        prevLesson: lessonPlan,
-        nextLesson: state,
-        messages,
-      });
+      // lessonPlanTracking.onStreamFinished({
+      //   prevLesson: lessonPlan,
+      //   nextLesson: state,
+      //   messages,
+      // }); // #TODO reinstate analytics
```
mantagen, line 275: What's stopping this from being uncommented?

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -329,17 +300,17 @@ export function ChatProvider({
   useEffect(() => {
     if (toxicModeration) {
       setMessages([]);
-      setLessonPlan({});
+      //setLessonPlan({}); #TODO what to do here
```
mantagen, line 303: We need to wipe local state (but not wipe it in the db)

mantagen (commented):
Looks good! Couple comments

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -158,17 +149,21 @@ export function ChatProvider({
...
       id,
-      lessonPlan,
+      lessonPlan: chat?.lessonPlan,
       options: {
         useRag: true,
         temperature: 0.7,
       },
     },
+    fetch(input: RequestInfo | URL, init?: RequestInit | undefined) {
+      lessonPlanSnapshot.current = chat?.lessonPlan ?? {};
```
stefl, line 164: Each fetch, we snapshot the lesson plan for analytics

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -294,30 +270,29 @@ export function ChatProvider({
...
     if (!hasFinished || !messages) return;
-    const state = findLatestServerSideState(messages);
-    if (!state) return;
-    setLessonPlanWithLogging(state, "hasFinished");
+    trpcUtils.chat.appSessions.getChat.invalidate({ id });
     if (shouldTrackStreamFinished.current) {
       lessonPlanTracking.onStreamFinished({
-        prevLesson: lessonPlan,
-        nextLesson: state,
+        prevLesson: lessonPlanSnapshot.current,
```
stefl, line 276: Use the snapshot stored during fetch

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
stefl: Resolved this with an override

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
stefl: Updated the logic to use the isStreaming prop to reduce additional processing when not streaming

mantagen: I think `isStreaming` is just derived from `isLoading` (from the `ai` hook), but a render behind -- but not worries for now

mantagen (approved):

PR #145, opened 2024-09-18
fix: message ids, flag and modify tables

mantagen:
## Description

- fix issue where moderations and lesson snapshots were referencing user messages (which differed to the message ids referenced by in app feedbacks)
  - after this work goes in i'll put in a migration to fix the old instances of the issue
- adds columns to modifications table so that we're storing
  - 'text' when action is 'other'
  - the path and value of the section being modified
- adds columns to the flag table so that we're storing
  - the path and value of the section being flagged


packages/api/src/router/chatFeedback.ts
```diff
@@ -0,0 +1,124 @@
+import * as Sentry from "@sentry/nextjs";
```
mantagen, line 1: Mostly moves the file into its own router

packages/core/src/functions/slack/getExternalFacingUrl.ts
```diff
@@ -1,8 +1,10 @@
 export function getExternalFacingUrl() {
+  const vercelEnv = process.env.VERCEL_ENV;
   const vercelUrl = process.env.VERCEL_URL;
+  const vercelProductionUrl = process.env.VERCEL_PROJECT_PRODUCTION_URL;
 
-  if (vercelUrl?.includes("production")) {
-    return "labs.thenational.academy";
+  if (vercelEnv === "production") {
```
mantagen, line 6: Realised this was still returning a preview URL before

packages/db/prisma/schema.prisma
```diff
@@ -144,11 +146,14 @@ model AilaUserModification {
...
 
-  userId     String                     @map("user_id")
-  chatId     String                     @map("chat_id")
-  messageId  String                     @map("message_id")
-  textForMod String                     @map("text_for_mod")
-  action     AilaUserModificationAction @map("action")
+  userId          String                     @map("user_id")
+  chatId          String                     @map("chat_id")
+  messageId       String                     @map("message_id")
+  textForMod      String?                    @map("text_for_mod")
```
mantagen, line 152: Made this nullable, will remove in a subsequent migration, as it's not storing meaningful data

packages/db/prisma/migrations/20240918123254_modifications_value_and_other_text/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "aila_user_modifications" ADD COLUMN     "action_other_text" TEXT,
+ADD COLUMN     "section_path" TEXT,
+ADD COLUMN     "section_value" JSONB,
+ALTER COLUMN "text_for_mod" DROP NOT NULL;
```
tomwisecodes, line 5: 🚨  run migrations stg reminder

packages/db/prisma/schema.prisma
tomwisecodes: why not just remove now?

tomwisecodes (approved):

PR #146, opened 2024-09-19
fix: next router skipping signed-in check

carlmw:
currently when a user signs into OWA and subsequently clicks the "Sign in" link in Aila (I.e they had the page open in another tab) they're presented with the error

"You're currently in single session mode. You can only be signed into one account at a time."

This appears to be caused by navigating to the page through the Next router. Replacing it with a plain anchor appears to trigger the expected signed-in check and redirect the user back to the app where they are signed in.

## How to test

1. When signed out
2. Open labs
3. Open /sign-in in another tab
4. Sign in
5. Return to the first tab and click "Sign in"
6. Try to sign in
7. You should be redirected to the home page

## Screenshots

How it used to look:

https://github.com/user-attachments/assets/a9f11e3d-0ebd-4817-a50d-d7fa5f24a52d

How it should now look:

https://github.com/user-attachments/assets/be04a9ba-6622-423e-92d7-bc1f9924b1c1

## Checklist

### DIFF
--- a/apps/nextjs/src/components/HeaderAuth.tsx
+++ b/apps/nextjs/src/components/HeaderAuth.tsx
@@ -1,6 +1,5 @@
 import { UserButton, useAuth } from "@clerk/nextjs";
 import { OakP } from "@oaknational/oak-components";
-import Link from "next/link";
 
 const HeaderAuth = () => {
   const { isSignedIn } = useAuth();
@@ -9,9 +8,9 @@ const HeaderAuth = () => {
     <>
       {isSignedIn && <UserButton />}
       {!isSignedIn && (
-        <Link href="/sign-in">
+        <a href="/sign-in">
           <OakP $font="body-2">Sign in</OakP>
-        </Link>
+        </a>
       )}
     </>
   );


codeincontext:
Thanks for the PR, @carlmw. Have you had a look at the use of the clerk `SignIn` component on the sign in page? While I'm sure that as hard refresh does fix the problem, it's possible that we can change the logic on that page to avoid the need for a sledgehammer-style workaround

For example, maybe we can conditionally render the `SignIn` component if we can detect this situation. On the other hand, if your team is in a rush and needs a quick fix, then we can push this change through and optimise it later instead

carlmw:
Okay I can see what's going on. It's the shared layout. `<SignIn>` relies on `<ClerkProvider>` to know if there is an active session. Since this is in the layout it doesn't update when navigating through the router. It's stale when we land on `/sign-in` and the logic inside `<SignIn>` that will automatically redirect when a user is already signed-in isn't triggered. 

This issue can be reproduced entirely in labs.

1. When signed out
2. Open production labs
3. Open `/sign-in` in another tab
4. Sign in
5. Return to the first tab and click "Sign in"
6. Try to sign in
7. "You're currently in single session mode. You can only be signed into one account at a time."

I'm going to step away from this and we can raise a separate issue in labs. 

My take is:

a. Make the link a proper anchor
b. Rejig the layouts so that `/sign-in` has it's own instance of `<ClerkProvider>` which should fetch the session whenever navigating to `/sign-in` (I think)
c. Find some method in Clerk that re-fetches (I found `useClerk().client.reload()` but that didn't seem to do what I was expecting and isn't documented)

codeincontext:
Okay, thank you for the time you've spent investigating this and the steps to reproduce. Leave it with us and we'll keep you informed about the solution we take, as it might be useful to you as well

codeincontext:
> @codeincontext this may be redundant if #208 resolves the issue?

I don't think it's the same issue, but we can try these reproduction steps and find out. I have a hunch that these issues stem from lower down in Clerk

codeincontext:
Some more context on this: It looks like Clerk can detect a log in when a background tab becomes active, but only within the same subdomain. I'm discussing with Clerk but I agree that a hard load of the login page is the simplest fix

https://github.com/user-attachments/assets/2afde4f1-70d7-4bf6-869a-ec546606a17a

codeincontext (approved):

PR #147, opened 2024-09-19
feat: allow the user to interact while moderation is in progress

stefl:
## Description

- This makes some front end logic changes to make it possible for the user to interact with the chat and see the final message from Aila during the moderation step
- Adds a "comment" document from the server to say that moderation is starting (we already filter these out so it seemed the easiest way)
- If the chat streaming status is "Moderation" or "Idle" then the user is able to interact with both the quick actions and the message box
- If the user interacts with them, then it disables the component, waits for the streaming state to be "Idle" and then does the appropriate action.

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
+++ b/apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
@@ -1,4 +1,4 @@
-import { useMemo } from "react";
+import { useMemo, useEffect, useRef } from "react";
 
 import { Message } from "ai";
 
@@ -16,8 +16,11 @@ export const useAilaStreamingStatus = ({
   isLoading: boolean;
   messages: Message[];
 }): AilaStreamingStatus => {
+  const prevStatusRef = useRef<AilaStreamingStatus | null>(null);
+
   const possibleTypePatch = `"type":"patch"`;
   const possibleTypePrompt = `"type":"prompt"`;
+  const possibleTypeModeration = `STARTING_MODERATION`;
 
   const ailaStreamingStatus = useMemo<AilaStreamingStatus>(() => {
     const lastMessage = messages[messages.length - 1];
@@ -29,6 +32,8 @@ export const useAilaStreamingStatus = ({
       if (!lastMessage) return "Loading";
       if (lastMessage.role === "user") {
         return "RequestMade";
+      } else if (lastMessage.content.includes(possibleTypeModeration)) {
+        return "Moderating";
       } else if (
         lastMessage.role === "assistant" &&
         lastSectionOfStreamingContent?.includes(possibleTypePatch)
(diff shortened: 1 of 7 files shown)


apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
```diff
@@ -16,46 +16,39 @@ export const useAilaStreamingStatus = ({
...
-        lastMessage.role === "assistant" &&
-        lastSectionOfStreamingContent?.includes(possibleTypePatch)
-      ) {
-        return "StreamingLessonPlan";
-      } else if (
-        lastMessage.content.includes(possibleTypePatch) &&
-        lastMessage.content.includes(possibleTypePrompt) &&
-        lastSectionOfStreamingContent !== ""
+        content.includes(`"type":"prompt"`) ||
+        content.includes('"type":"prompt"')
```
codeincontext, line 35: These two strings look the same to me. Am I missing something?

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
```diff
@@ -16,46 +16,39 @@ export const useAilaStreamingStatus = ({
...
-  // @todo lets build a logger
-  if (process.env.NODE_ENV !== "development") {
-    console.log("ailaStreamingStatus", ailaStreamingStatus);
-  }
+  useEffect(() => {
+    if (prevStatusRef.current !== ailaStreamingStatus) {
+      console.log("ailaStreamingStatus changed:", ailaStreamingStatus);
+      prevStatusRef.current = ailaStreamingStatus;
+    }
+  }, [ailaStreamingStatus]);
```
codeincontext, line 51: It looks like this could be replaced with a `useMemo`

apps/nextjs/src/components/AppComponents/Chat/prompt-form.tsx
```diff
@@ -13,56 +13,101 @@ import { useLessonPlanTracking } from "@/lib/analytics/lessonPlanTrackingContext
...
   }, []);
+
   const sidebar = useSidebar();
+
+  const queueSubmission = useCallback(
+    (value: string) => {
+      setQueuedUserInput(value);
+      timeoutRef.current = window.setTimeout(() => {
+        setQueuedUserInput(null);
+      }, 10000);
```
codeincontext, line 53: The goal of the timeoutRef isn't very clear to me. What would the user see if it times out? Maybe a commend would be useful to explain it and/or the reason for 10000ms

codeincontext (commented):
This approach looks sensible, though I haven't tested it myself. Just a couple of questions

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
stefl: Ah - this was where I was originally testing for escaped strings too \"type\" - looks like that reverted somehow

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
stefl: Hmm - are you supposed to have side-effects with useMemo?

apps/nextjs/src/components/AppComponents/Chat/prompt-form.tsx
codeincontext: Would the message still be in the input field, or would it have disappeared in that case?

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
codeincontext: Oh, I see. `prevStatusRef` isn't used for anything, it's just to allow this effect to log. Sorry, I thought the goal was to derive one value from another

If you are ok with an initial log message, maybe you don't need a ref at all:

```ts
  useEffect(() => {
    console.log("ailaStreamingStatus set:", ailaStreamingStatus);
  }, [ailaStreamingStatus]);
```

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
stefl: Looks like the linter removed these on save. Resolved.

codeincontext (approved):

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
stefl: Good catch, thanks

apps/nextjs/src/components/AppComponents/Chat/prompt-form.tsx
stefl: I believe I've set it to still be displayed, but the input box should then be disabled, then the moderation completes, then it empties.

codeincontext (approved):

PR #153, opened 2024-09-24
fix: in app feedback snags

tomwisecodes:
[## Description
](https://www.notion.so/oaknationalacademy/In-app-feedback-snags-41fb8cca4d454b16851537dba899fc6f)


apps/nextjs/src/components/AppComponents/Chat/drop-down-section/modify-button.tsx
```diff
@@ -136,10 +157,12 @@ function handleLabelText({
   text: string;
   section: string;
 }): string {
+  console.log("section", section);
```
JBR90, line 160: 🔥

JBR90 (approved):

PR #155, opened 2024-09-24
fix: dialog window bug

tomwisecodes:
## Description
fixes:
https://www.notion.so/oaknationalacademy/Feedback-Form-Sizing-10426cc4e1b18022addddbdf27ec49f9?pvs=23


stefl (approved):
Is this okay on mobile if the viewport changes size due to scrolling?

PR #157, opened 2024-09-24
fix: change plan parts embeddings to 256 vectprs

mantagen:
## Description

- the new embeddings for lesson_plan_parts are 256 in length, this requires us to drop the column and recreate
- this migration hasn't been run on staging or production yet
- need to run it before uploading the new ingest data


packages/core/src/models/lessonPlans.ts
```diff
@@ -335,75 +335,4 @@ ${p.content}`,
       WHERE id = ${id}`;
     return result;
   }
-
-  async search(
```
mantagen, line 339: This was unused, and the table 'lesson_plans' no longer has embeddings in it, so this will not work.

codeincontext (approved):

PR #164, opened 2024-09-26
chore: wrap auth middleware errors

codeincontext:
Our middleware-related sentry errors don't give us much context:

`Error in nextMiddleware`
=> `Unexpected token [redacted]... is not valid JSON`

I'm hoping that wrapping the different places that might throw an error might help to narrow down the origin of the errors


mantagen (approved):
I think we should implement this catch and rethrow with `{ cause }` as our default pattern. I also think when we rethrow it should always be a custom error class, but one step at a time!

codeincontext:
> I think we should implement this catch and rethrow with `{ cause }` as our default pattern. I also think when we rethrow it should always be a custom error class, but one step at a time!

I do 100% agree with you on that, though I'm hoping against hope that this might not need to be a permanent part of the middleware

PR #165, opened 2024-09-27
fix: stop interaction except in moderation or idle

stefl:
## Description

- Adjusts the in-moderation interaction feature so that both the quick buttons component and the prompt form are managed by the ChatProvider
- Ensures matching logic between all related components about when they can be interacted with

## To test
- Start a chat
- During the moderation step when the left hand message first appears, click Continue.
- Notice you can't type in the box or click Continue again
- Do the same by typing in the box
- Wait until it is interactive again
- Type in the box again
- Notice the placeholder is the text you typed (it's queued)
- Notice it submits after a few seconds

### DIFF
--- a/apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
+++ b/apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
@@ -1,5 +1,6 @@
 import React, {
   createContext,
+  useCallback,
   useContext,
   useEffect,
   useMemo,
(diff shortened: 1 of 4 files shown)


apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -245,6 +249,50 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
       messageHashes,
     });
 
+  // Handle queued user actions and messages
```
codeincontext, line 252: Ideally, rather than adding to this function this logic could be in its own hook called by the provider. I know that there's time pressure for this release though

codeincontext (approved):
Tested locally with an artificial delay in moderation and it works as expected

PR #166, opened 2024-09-27
fix: refetch chat on mount to fix the blank chat issue on back navigation

stefl:
## Description

- Addresses an issue where the chat is blank when navigating back from the downloads page

## How to test

1. Create a complete chat
2. Go to the download page
3. Press the back to lesson link at the top left
4. See that the chat has loaded

### DIFF
--- a/apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
+++ b/apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
@@ -42,7 +42,7 @@ function patchHasBeenApplied(
 export const useTemporaryLessonPlanWithStreamingEdits = ({
   lessonPlan,
   messages,
-  //isStreaming,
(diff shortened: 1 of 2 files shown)


apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
```diff
@@ -90,19 +90,20 @@ export const useTemporaryLessonPlanWithStreamingEdits = ({
...
+  //       return workingLessonPlan;
+  //     }
+  //     const newLessonPlan: LooseLessonPlan | undefined = applyLessonPlanPatch(
+  //       { ...workingLessonPlan },
+  //       patch,
+  //     );
+  //     return newLessonPlan ?? workingLessonPlan;
+  //   },
+  //   [isStreaming],
+  // );
```
tomwisecodes, line 107: Shall we delete commented out code while we arnt using it?

PR #169, opened 2024-09-30
build: use release token to merge RC PRs so that the release workflow…

codeincontext:
Our accept_release_candidate workflow merges the RC pull request. That push should then be picked up by the release workflow. Because we're using the default github_token to merge, [github won't trigger the push event](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow)

This PR uses `RELEASE_GITHUB_TOKEN` to merge the PR instead, which should trigger the push


simonrose121 (approved):
LGTM!

PR #170, opened 2024-09-30
test: allow one e2e retry in CI

codeincontext:
End to end tests can be flaky. While the ideal is to not have flaky tests, the reality is that failing tests on PRs lead developers to expect or ignore a failing check

This PR allows a retry in CI in case a failure is just flake


simonrose121 (approved):
It's working!

Should we merge the downloads e2e fix one into this so they're passing before merge?

PR #178, opened 2024-10-01
fix: feature flag "download all" button

codeincontext:
The "download all" button is blocking our latest release. Rather than try to fix it further on main, let's disable it and fix it out-of-band

The feature hook wasn't being used, so it seems like a good opportunity to simplify it. Our analytics setup has changed and I don't think that we need to identify here any more

The features has been created in posthog staging and production

### After

CleanShot 2024-10-01 at 16 35 12@2x

### DIFF
--- a/apps/nextjs/src/utils/useClientSideFeatureFlag.ts
+++ b/apps/nextjs/src/utils/useClientSideFeatureFlag.ts
@@ -8,8 +8,11 @@ export function useClientSideFeatureFlag(flag: string): boolean {
   const [featureEnabled, setFeatureEnabled] = useState<boolean | undefined>();
 
   useEffect(() => {
+    const isDebug = process.env.NEXT_PUBLIC_POSTHOG_DEBUG === "true";
+
     return client.onFeatureFlags(() => {
-      setFeatureEnabled(client.isFeatureEnabled(flag));
+      const isEnabled = isDebug ? true : client.isFeatureEnabled(flag);
+      setFeatureEnabled(isEnabled);
     });
   }, [client, flag]);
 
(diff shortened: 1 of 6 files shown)


apps/nextjs/src/utils/useClientSideFeatureFlag.ts
```diff
@@ -1,57 +1,17 @@
...
-        setIsCheckComplete(true);
-      }
-    }
-
-    checkFeatureFlagStatus();
-  }, [user, isLoaded, featureFlagId, posthogAiBetaClient]);
+    return client.onFeatureFlags(() => {
+      setFeatureEnabled(client.isFeatureEnabled(flag));
+    });
+  }, [client, flag]);
```
codeincontext, line 14: This hook is a pretty close copy of the [official posthog hook ](https://github.com/PostHog/posthog-js/blob/26150b197dc1a2f5241b8d29ee7a359331a1a0a4/react/src/hooks/useFeatureFlagEnabled.ts)

mantagen: Hmm wonder why we aren't using there's

mantagen: Actually tbf there's it seems we don't control which posthog instance it uses

codeincontext: If we use theirs it would mean adding their provider and providing `client` to it. But only one of the clients which could be confusing. The actual code in their library is pretty minimal though

mantagen (approved):

PR #179, opened 2024-10-01
test: use @clerk/testing-esque sign-in helper

codeincontext:
## Description

- This PR aligns our e2e test auth with clerk's best practices
- As of `v1.3.7`,  @clerk/testing has a helpful `signIn` helper. This feels like a good replacement for our magic link auth solution: it's approved for this purpose and should be a little quicker
  - I implemented `clerk.signIn` and when I pushed to a preview I later found out that it triggers our `unsafe-eval` CSP rule
  - The helper itself is actually pretty small. I swapped out the [waitForFunction](https://github.com/clerk/javascript/blob/1a1247ce08b8a3e6d4435548a344cd859cf3b007/packages/testing/src/playwright/helpers.ts#L79-L82) calls which broke our CSP with an `evaluate` call, and copied over the [signIn helper](https://github.com/clerk/javascript/blob/1a1247ce08b8a3e6d4435548a344cd859cf3b007/packages/testing/src/common/helpers-utils.ts)
- We no longer need to call `setupClerkTestingToken` for tests calling `prepareUser`
- Clerk recommend calling `clerkSetup` once in a setup file, so I've set that up
- Added `+clerk_test` to our test persona email addresses, as that allows clerk's magic 424242 email code to be accepted


mantagen (approved):
Yep, nice!

PR #180, opened 2024-10-02
perf: add fts index to snippets table

simonrose121:
## Description

- Adds a full-text search index to the snippets table that we had on the old production DB.
- @remy was reporting performance hits using the new DBs, we identified this as the probable culprit.
- The snippets table we're applying this too is pretty hefty, but as we're not writing to it it shouldn't be an issue.

## Issue(s)

Fixes #ENG-985

## How to test

1. Once applied to production, test the transcripts API route on the Open API.


remy (approved):
😍 ta

mantagen (approved):

PR #181, opened 2024-10-02
test: add @common-auth strategy to e2e tests

codeincontext:
## Description

Our older e2e tests sign in by manually typing test credentials into the sign in form. We have a new approach which generates test users and signed them in through the clerk API. This PR brings those older tests in line with our new approach

- Keep the existing `@authenticated` setup step, but rename it to `@common-auth` as other tests can still authenticate individually
  - Call `prepareUser` from the common auth setup step
  - Save the chatId from prepareUser to a cookie, so individual tests can retrieve it and test against a completed chat
- Create a new auth.test.ts test file which goes through a clerk login manually
- Remove unnecessary `test.describe` wrappers as they add noise the test UI
- Set maximum concurrent workers to 5 in development. I found testing this that pressing the big green button in testflight can spam the dev server a bit too much

NOTE: This PR is branched off #179, so the base will change when that gets merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ pnpm test-e2e
 
 Our Playwright tests are organised with tags:
 
-- `@authenticated`: Tests with this tag use a pre-created test user, reused without cleanup between tests.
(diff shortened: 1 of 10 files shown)


mantagen (approved):
LGTM

PR #183, opened 2024-10-03
chore: update CompletedLessonPlanSchema so doesn't use min() or max()

mantagen:
**So that the ingest service PR isn't too unwieldy, I'm trying to break it into smaller PRs**

## Description

- This one updates the CompletedLessonPlan schema  so that it doesn't use Zod min() and max(), which aren't supported by Structured Output

### DIFF
--- a/packages/aila/src/protocol/schema.ts
+++ b/packages/aila/src/protocol/schema.ts
@@ -341,13 +341,36 @@ export const KeywordsOptionalSchema = z.array(KeywordOptionalSchema);
 export type KeywordOptional = z.infer<typeof KeywordOptionalSchema>;
 export type Keyword = z.infer<typeof KeywordSchema>;
 
+function minMaxText({
+  min,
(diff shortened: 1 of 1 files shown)


packages/aila/src/protocol/schema.ts
```diff
@@ -341,13 +341,36 @@ export const KeywordsOptionalSchema = z.array(KeywordOptionalSchema);
...
+  max,
+  entity = "elements",
+}: {
+  min?: number;
+  max?: number;
+  entity?: "elements" | "characters";
+}) {
+  if (typeof min !== "number" && typeof max !== "number") {
+    throw new Error("min or max must be provided");
+  }
```
codeincontext, line 355: Can we skip this runtime check by making the type a `min | max` union?

codeincontext (approved):
Nicely done. Looks good to me

PR #189, opened 2024-10-03
test: add an e2e test for the sidebar

codeincontext:
The most minimal test possible to check that you can access a chat from the sidebar


mantagen (approved):
Love it

PR #190, opened 2024-10-03
chore: openai batch api helpers

mantagen:
## Description

- Adds [OpenAI batch API](https://platform.openai.com/docs/guides/batch) helper functions in preparation for ingest service


codeincontext (approved):
Makes sense to me 👍

PR #191, opened 2024-10-03
chore: ingest helper functions for different steps

mantagen:
## Description

- zod schema and types
- helper functions for chunking and embedding steps


packages/ingest/src/utils/getDataHash.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { IngestError } from "../IngestError";
+
+export function getDataHash(data: unknown) {
+  try {
+    const str = JSON.stringify(data);
+    const hash = crypto.createHash("sha256").update(str).digest("hex");
+
+    return hash;
+  } catch (cause) {
```
codeincontext, line 11: Sneaky to name the error `cause`. I like that a lot!

PR #197, opened 2024-10-04
test: add e2e test for mobile chat flow

codeincontext:
Okiedokie. Here's a first mobile-specific e2e test

First hurdle: I needed to rejig the playwright config. Now any test with the `@mobile-common-auth` tag will be in a mobile viewport (and the common user persona)

Next hurdle: This test runs in webkit to get closer to an iPhone browser. Webkit applies the `upgrade-insecure-requests` CSP rule on localhost, which tries to load site resources with HTTPS. That needed a CSP change

Next hurdle: When I rejigged test auth, we lost the call to accept cookie consent by default

Next thing: I noticed that we were an `isFinished` helper as if it would assert, but actually it returns true or false. That's my bad. I've now added `assertFinished`

Finally: Added the mobile chat test and IT WORKS

There's a lot more we can add later to check that the right things are being rendered, that panels are scrolled to the right places, etc. But it's a start


apps/nextjs/playwright.config.ts
```diff
@@ -4,32 +4,43 @@ export default defineConfig({
...
+    // Setup projects: Run before other tests
     {
-      name: "setup",
+      name: "global-setup",
       testMatch: "global.setup.ts",
     },
     {
-      name: "common-auth",
-      testMatch: "common-auth.setup.ts",
-      dependencies: ["setup"],
```
codeincontext, line 14: Renamed to `common-persona-setup`

apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
```diff
@@ -55,33 +55,28 @@ test(
...
       await continueChat(page);
       await waitForGeneration(page, generationTimeout);
       await expectSectionsComplete(page, 10);
-      await page.waitForTimeout(500);
 
       setFixture("roman-britain-5");
       await continueChat(page);
       await waitForGeneration(page, generationTimeout);
       await expectSectionsComplete(page, 10);
-      await page.waitForTimeout(500);
```
codeincontext, line 82: It seems to be doing ok without these. Let's see

mantagen (commented):
🎉

mantagen (approved):

PR #194, opened 2024-10-03
chore: add ingest db-helpers

mantagen:
## Description

- updates to `prisma.schema` including
  - adding 'multiSchema' with schema 'ingest' in addition to the default 'public'
  - adding all the required tables to the 'ingest' schema 
- migration file for above changes
- adds 'db-helpers' directory as a with various database helper functions

PR #199, opened 2024-10-07
chore: final ingest helpers

mantagen:
## Description

- final set of ingest helpers before the 'steps' functions PR (which should see it all come together)
- contains changes within https://github.com/oaknational/oak-ai-lesson-assistant/pull/194 so that should go in first and then i'll merge main into this for a smaller/easier review


packages/ingest/src/import-lessons/importLessons.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    await prisma.ingestLesson.createMany({
+      data,
+    });
+
+    console.log(`Imported ${data.length} lessons`);
+
+    if (true) {
+      break;
+    }
```
codeincontext, line 67: It looks like you had this in testing, but now it's causing some dead code below?

PR #200, opened 2024-10-07
feat: ingest scripts

mantagen:
## Description

This adds the ingest 'steps' scripts, these are:
- start (creates the ingest and pulls in lessons from the Oak API)
- captions (pull in and store captions) 
- start lesson plan generation (with open ai batch api)
- sync lesson plan generation (check status and parse/store lesson plans on completion)
- chunk (break lesson plans in to parts)
- start embedding (batch api)
- sync embedding (status, parse, store)

These functions will eventually either be steps in a pipeline that runs on a cronjob, or instead to support a simple ingest dashboard where non-devs can control ingests.

For the time being, to run one of the scripts above, from `/packages/ingest/src` run:
```
pnpm ingest [command] [?ingest_id]
```
where `ingest_id` is optional and defaults to the id of the latest ingest.


packages/ingest/src/import-lessons/importLessons.ts
```diff
@@ -1,71 +0,0 @@
-import { prisma } from "@oakai/db";
```
mantagen, line 1: This file as been superseded, and is no longer used

packages/ingest/src/steps/0-start.ts
```diff
@@ -0,0 +1,58 @@
+import { PrismaClientWithAccelerate } from "@oakai/db";
+
+import { IngestError } from "../IngestError";
+import { IngestConfig } from "../config/ingestConfig";
+import { createIngestRecord } from "../db-helpers/createIngestRecord";
+import { importLessonsFromCSV } from "../import-lessons/importLessonsFromCSV";
+import { importLessonsFromOakDB } from "../import-lessons/importLessonsFromOakDB";
+
+const config: IngestConfig = {
```
mantagen, line 9: This config will be passed as an argument

packages/ingest/src/steps/1-captions.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const lessons = await loadLessonsAndUpdateState({
+    prisma,
+    ingestId,
+    prevStep,
+    currentStep,
+  });
+
+  if (ingest.config.sourcePartsToInclude === "title-subject-key-stage") {
+    console.log("Skipping captions fetch for title-subject-key-stage ingest");
```
codeincontext, line 37: It would be nice to have a follow up task to see if it's a good idea to move these logs to the new logger. It might not make sense if this package is only run in isolation of our other packages

codeincontext (approved):

packages/ingest/src/steps/1-captions.ts
mantagen: 💯

PR #202, opened 2024-10-08
fix: add embedding indexes to improve slow vector similarity queries

stefl:
## Description

- We are seeing slow queries related to PGVector cosine distance queries
- This adds one migration to Snippets to add a lookup for key stage and subject
- Adds an SQL script which we would have to run manually / in another step to add the embedding indexes
- However, this will likely mean we end up with Prisma drift and we will need to come up with a solution in a follow-up

We need to use ivff indexes for the embedding columns
- Prisma doesn’t support it
- Prisma doesn’t have a feature to ignore columns or indexes when comparing schemas to do migrations
- So… I’ve had to apply these indexes manually on staging and production to fix the immediate issue
- This now means we will have the dreaded “you have schema drift, delete your entire database” prisma issue
  - Apparently ways round this might include:
  - Drop the index with a pre-script, run migrations, then create them, for every migration
  - Stop using Prisma for migrations and use something else
  - Move the embeddings piece to a separate Postgres schema which is not managed by Prisma, and then join against the main Postgres schema, more like a search index

## Issue(s)

Fixes an issue with slow queries potentially causing high database load and failed queries for users


packages/db/prisma/migrations/20241008111057_add_key_stage_subject_index_to_snippets/migration.sql
```diff
@@ -0,0 +1,19 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `captions_id` on the `ingest_lesson` table. All the data in the column will be lost.
+  - You are about to drop the column `lesson_plan_id` on the `ingest_lesson` table. All the data in the column will be lost.
+
+*/
+-- DropIndex
+DROP INDEX "ingest"."ingest_lesson_captions_id_key";
```
stefl, line 9: @mantagen I'm not sure what this is. It's auto-generated by Prisma

packages/db/prisma/migrations/20241008111057_add_key_stage_subject_index_to_snippets/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+
+-- AlterTable
+ALTER TABLE "ingest"."ingest_lesson" DROP COLUMN "captions_id",
```
stefl, line 12: Again - not sure why this is in the migration

packages/db/prisma/migrations/20241008111057_add_key_stage_subject_index_to_snippets/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+DROP COLUMN "lesson_plan_id";
+
+-- CreateIndex
+CREATE INDEX "lesson_summaries_key_stage_id_subject_id_idx" ON "public"."lesson_summaries"("key_stage_id", "subject_id");
```
stefl, line 16: This is the index that we need for key stage and subject

packages/db/scripts/embeddings-indexes/add_embedding_indexes.sql
```diff
@@ -0,0 +1,19 @@
+-- Add vector index for quiz_questions table
```
stefl, line 1: Running this should add the ivfflat indexes that optimise cosine similarity queries. We would need to run this manually or whenever we run Prisma migrate. But I think this would lead to getting schema drift warnings.

mantagen (approved):
Makes sense. Need to come up with a strategy here

stefl:
@mantagen I think we paused this when you were doing related work. Safe to merge now?

packages/core/src/rag/index.ts
```diff
@@ -5,15 +5,9 @@ import {
   LessonPlanPart,
   PrismaClientWithAccelerate,
   Subject,
+  prisma as globalPrisma,
```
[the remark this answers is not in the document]
stefl: This is the fallback so that if you don't pass it in it uses the global prisma

stefl:
Closing this because the embedding columns have been removed, making most of this PR redundant. We just need to remember we need to have this manually run `CREATE INDEX IF NOT EXISTS vector_table_name_embedding_idx 
ON "table_name" USING ivfflat (embedding vector_cosine_ops);`

PR #205, opened 2024-10-08
chore: ingest helpers final changes

mantagen:
## Description

- a final round of helper changes before the pipeline step scripts

### DIFF
--- a/packages/ingest/src/embedding/handleEmbeddingBatchSuccess.ts
+++ b/packages/ingest/src/embedding/handleEmbeddingBatchSuccess.ts
@@ -1,6 +1,7 @@
 import { PrismaClientWithAccelerate } from "@oakai/db";
 
 import { IngestError } from "../IngestError";
+import { createErrorRecord } from "../db-helpers/createErrorRecord";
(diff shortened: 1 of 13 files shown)


packages/ingest/src/embedding/handleEmbeddingBatchSuccess.ts
```diff
@@ -0,0 +1,96 @@
...
+        lessonIdsFailed.add(lessonId);
+        continue;
+      }
+    } catch (error) {
+      if (error instanceof IngestError && error.lessonId) {
+        lessonId = error.lessonId;
+      }
+      if (lessonId) {
+        lessonIdsFailed.add(lessonId);
+      }
```
codeincontext, line 57: Do we want to track the error here? Maybe push to an array of errors to log at the end?

codeincontext (approved):

packages/ingest/src/embedding/handleEmbeddingBatchSuccess.ts
mantagen: Thanks yes was def thinking of this -- glad you brought it up. Will make this change before merging

PR #208, opened 2024-10-08
fix: hook error on sign-in page

stefl:
## Description

- This attempts to fix an issue where the sign in page was rendering with a different number of hooks
- If the user is logged in and visits /sign-in it redirects to /

## Issue(s)

Fixes [Sentry issue 3875705](https://oak-national-academy.sentry.io/issues/3875705/?query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=0)

### DIFF
--- a/apps/nextjs/src/app/sign-in/[[...sign-in]]/page.tsx
+++ b/apps/nextjs/src/app/sign-in/[[...sign-in]]/page.tsx
@@ -1,4 +1,9 @@
-import { SignIn } from "@clerk/nextjs";
+"use client";
+
+import { useEffect } from "react";
+
+import { SignIn, SignedIn, SignedOut, useUser } from "@clerk/nextjs";
+import { useRouter } from "next/navigation";
 
 import SignUpSignInLayout from "@/components/SignUpSignInLayout";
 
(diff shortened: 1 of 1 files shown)


codeincontext:
We updated the version of Clerk on October 10th, but it doesn't look like it's fixed the issue. It's frustrating as the change here implies that it's a bug in Clerk's package

Have you been able to replicate this locally, or is this fix a "try it and see" situation? I'd recommend adding a comment to explain why it's here and when it can be removed

stefl:
> We updated the version of Clerk on October 10th, but it doesn't look like it's fixed the issue. It's frustrating as the change here implies that it's a bug in Clerk's package
> 
> Have you been able to replicate this locally, or is this fix a "try it and see" situation? I'd recommend adding a comment to explain why it's here and when it can be removed

This is a fix and see if it works for sure. If we want to try it and see if it removes the Sentry error then we could try it, otherwise happy to close.

codeincontext:
Ok, why don't we merge and if the error continues we can decide whether to revert it 👍

PR #211, opened 2024-10-10
fix: findLast is not available for all browsers (WIP)

stefl:
## Description

- Fixes an issue where some browsers do not have findLast because it's a relatively new JS method

## Issue(s)

Fixes [9006645](https://oak-national-academy.sentry.io/issues/9006645/?project=4507089597300816&query=release%3Av1.9.1&referrer=release-issue-stream)


codeincontext (approved):
I looked into this. I was expecting our tsconfig to be polyfilling findLast, and couldn't work out why it wasn't. I think this is the right thing to do

mantagen (approved):

PR #213, opened 2024-10-10
fix: reduce middleware syntax error logs, send to logs, tests for CSP in each env

stefl:
## Description

- Reduces the Syntax Error logging we currently have in Sentry, but sends them to Datadog so we still have the data
- Adds more information on the request
- Typescript was complaining in the CSP file when I added tests to test middleware behaviour
- Slightly refactored the CSP file so that we can pass in config rather than using environment variables
- Added tests to ensure the CSP logic is correct


apps/nextjs/src/middleware.ts
```diff
@@ -1,19 +1,86 @@
...
+      : new Error("Error in nextMiddleware", { cause: error });
+
+  Sentry.captureException(wrappedError, {
+    extra: {
+      requestInfo,
+      eventInfo,
+      bodyText,
+      errorType: "OtherError",
+    },
+  });
```
codeincontext, line 83: This looks smart. Let's extract it somewhere and co-locate with the test

apps/nextjs/src/middlewares/auth.middleware.ts
```diff
@@ -149,7 +152,7 @@ export async function authMiddleware(
       return response;
     }
   } catch (error) {
-    console.error("Error in authMiddleware", error);
+    console.error({ event: "middleware.auth.error", error });
```
stefl, line 155: If we log JSON then Datadog is able to query it

apps/nextjs/src/middlewares/csp.ts
```diff
@@ -154,26 +154,27 @@ const buildCspHeaders = (nonce: string) => {
 
   const cspString = Object.keys(baseCsp)
     .map((policy) => {
-      const value = [...baseCsp[policy]];
+      const value = [...(baseCsp[policy as keyof typeof baseCsp] || [])];
```
stefl, line 157: I suspect all of this is redundant now @codeincontext ?

apps/nextjs/src/middleware.ts
stefl: Agreed - I'll extract it to a utility function

apps/nextjs/src/middlewares/csp.ts
codeincontext: I think something is being conflated here, as the map is per declaration key (like `img-src`), and now `addPolicyIfExists` is doing its own looping for the same thing. The naming of `addPolicyIfExists` suggests that there's something that might not exist, but it isn't taking a key for a declaration name

It doesn't seem to relate to the middleware logs we're trying to solve, so maybe we could address it in a followup PR so we can get the logging fix in today's RC?

codeincontext: It turns out that when this is imported from a test then it starts to get TS errors. The approach here does look correct

apps/nextjs/src/middleware.ts
```diff
@@ -1,31 +1,68 @@
...
     return NextResponse.json({ error: "Bad Request" }, { status: 400 });
   }
 
-  const wrappedError = new Error("Error in nextMiddleware", { cause: error });
-  Sentry.captureException(wrappedError, { extra });
-
   return NextResponse.json({ error: "Internal Server Error" }, { status: 500 });
 }
 
+const cspConfig: CspConfig = {
```
stefl, line 31: Moving this to a config allows us to test the CSP generation without relying on ENV being set.

apps/nextjs/src/middlewares/csp.ts
```diff
@@ -1,56 +1,70 @@
 import { NextRequest, NextResponse } from "next/server";
 
-const sentryEnv = process.env.NEXT_PUBLIC_SENTRY_ENV;
-const sentryRelease = process.env.NEXT_PUBLIC_APP_VERSION;
-const sentryReportUri = `${process.env.SENTRY_REPORT_URI}&sentry_environment=${sentryEnv}&sentry_release=${sentryRelease}`;
+function generateNonce(): string {
```
stefl, line 3: I'm not too happy about this. Because it's an edge function, and we're running the test in node we need a way to generate the nonce in both environments.

apps/nextjs/src/lib/errors/server/middlewareErrorLogging.ts
```diff
@@ -0,0 +1,71 @@
+import * as Sentry from "@sentry/nextjs";
```
codeincontext, line 1: It potentially feels like quite a disconnected place to put this file, and I wonder if it should sit alongside the middleware?

stefl: Happy to move it!

apps/nextjs/src/middlewares/test/snapshots/policies.development.test.txt
stefl, line 1: These are snapshots so that we can review if there are any changes to how the CSP is generated

apps/nextjs/src/middlewares/test/snapshots/policies.development.test.txt
```diff
@@ -0,0 +1,14 @@
+default-src 'self'
+media-src 'self' 'self' https://*.mux.com https://stream.mux.com blob:
```
stefl, line 2: @codeincontext is this a bug? Should there be as string after blob:?

codeincontext: Just a bug! Just an oddity

CleanShot 2024-10-14 at 12 25 00@2x

apps/nextjs/src/middlewares/csp.test.ts
```diff
@@ -94,30 +39,55 @@ const originalRequire = jest.requireActual<NodeRequire>("module");
...
       const generatedPolicies = generatePoliciesForEnvironment(env);
-
-      if (regenerateSnapshots) {
-        writePoliciesToFile(env, generatedPolicies);
-        console.log(`Regenerated snapshot for ${env} environment`);
-      } else {
-        const savedPolicies = readPoliciesFromFile(env);
-        expect(generatedPolicies).toBe(savedPolicies);
-      }
+      expect(generatedPolicies).toMatchSnapshot();
```
codeincontext, line 90: So nice!

PR #216, opened 2024-10-10
chore: refactor chat provider

codeincontext:
Please review commit-by-commit

I'm tracking down a bug in the UI, and as a first start I thought that it would help to see the inner workings of the chat provider more clearly

By the way: I accidentally merged and subsequently reverted an earlier version of this PR


apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -113,26 +113,12 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
     refetch: refetchModerations,
-  } = trpc.chat.appSessions.getModerations.useQuery(
-    { id },
-    {
-      refetchOnMount: true,
-      refetchOnWindowFocus: true,
-      staleTime: 0,
-    },
-  );
+  } = trpc.chat.appSessions.getModerations.useQuery({ id });
```
stefl, line 235: The refetch on mount option I think was what fixed the issue with back navigation when returning to the chat after the download page.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -328,9 +327,18 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
       invokeActionMessages(response.content);
 
       trpcUtils.chat.appSessions.getChat.invalidate({ id });
+      clearHashCache();
```
stefl, line 327: Is the onFinish handler reliable? Do we need to have similar logic in onError?

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
stefl: You can't actually see this error in local development. You'll need to deploy to Vercel to see it in action, so could you check it doesn't reintroduce the problem?

stefl (approved):
Could you check that this doesn't reintroduce the back navigation from downloads problem?

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
codeincontext: Interesting, I hadn't considered that

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
codeincontext: Sorry, I thought I had replied to this!

I looked at the tanstack-query docs which said that these three properties were already the defaults, so I did remove them without checking that. I'll have a test now and will see if I can get to the bottom of it

mikeritson-oak (changes requested):
I spoke to @codeincontext about this and it's waiting on other PRs and work to move along before this can progress.

PR #139, opened 2024-09-17
fix: revert error to previous protocol

mantagen:
## Description

- reverts https://github.com/oaknational/oak-ai-lesson-assistant/pull/122
- fixes [issue described in this slack thread](https://oaknationalacademy.slack.com/archives/C05RYE3JL4S/p1726559097143399)

PR #218, opened 2024-10-11
fix: revert "fix: revert error to previous protocol (#139)

codeincontext:
This reverts commit e053b895503d5a29522404ce14e13239b6db7852.

We reverted the protocol changes for error messages like the friendly rate limit message in the Chat. Now that we're using the latest protocol with structured outputs, we need the new version again

### DIFF
--- a/apps/nextjs/src/utils/testHelpers/consumeStream.ts
+++ b/apps/nextjs/src/utils/testHelpers/consumeStream.ts
@@ -14,3 +14,12 @@ export async function consumeStream(
 
   return result;
 }
+
+export function extractStreamMessage(streamedText: string) {
+  const content = streamedText.match(/0:"(.*)"/);
+  if (!content?.[1]) {
+    throw new Error("No message found in streamed text");
+  }
+  const strippedContent = content[1].replace(/\\"/g, '"');
+  return JSON.parse(strippedContent);
+}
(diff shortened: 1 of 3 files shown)


apps/nextjs/src/utils/testHelpers/consumeStream.ts
```diff
@@ -14,3 +14,12 @@ export async function consumeStream(
 
   return result;
 }
+
+export function extractStreamMessage(streamedText: string) {
+  const content = streamedText.match(/0:"(.*)"/);
```
stefl, line 19: Could this potentially accidentally match an escaped "? 

Should it be `/0:"(.*)"^/` ?

stefl (approved):
Just a comment on whether the regex needs to match to the end of the line

apps/nextjs/src/utils/testHelpers/consumeStream.ts
codeincontext: I *think* it would be greedy so it should capture the largest possible string, but I agree that it's ambiguous. I can match for the start as well

codeincontext: My attempt at this wasn't working, and I'm having some local troubles running the test quite. I'll leave it for now given that this is test code rather the production code, and that it seemed to be running ok on main until it was reverted

PR #219, opened 2024-10-11
test: add e2e test for rate limiting

codeincontext:
Adds a test to check that user rate limits are incremented for each message, and that the correct rate limit message is shown to the user

This means:
- A new persona for a user with 119/120 usage of their rate limit
- Setting the rate limit on personas
- A test to send one message successfully, and then trigger their rate limit

Gotcha: I found that @upstash/ratelimit instances will cache in memory when a given ratelimit is exceeded. To reset this in-memory cache we need to import the same ratelimit instance in the test helper


apps/nextjs/src/components/AppComponents/Chat/chat-message/index.tsx
```diff
@@ -190,6 +190,9 @@ function MessageWrapper({
   children,
   type,
 }: MessageWrapperProps) {
+  const testId = errorType
```
stefl, line 193: Should this be === ?

stefl (approved):

apps/nextjs/src/components/AppComponents/Chat/chat-message/index.tsx
codeincontext: The behaviour I'm looking for is: If there's an error type, use `chat-message-wrapper-error-{errorType}`. Otherwise use `chat-message-wrapper-{type}`

PR #220, opened 2024-10-11
test: add a delay after generation to let romans test settle

codeincontext:
Our end-to-end test for the aila chat with fixtures is regularly failing. It got worse after removing an arbitrary delay before clicking the continue button

Problem:
  - The e2e tests do some things quicker than a typical user, for example the moderation step is a lot quicker. This could cause race conditions which we don't normally see
  - This doesn't mean that there isn't a bug, it could just be a rare one
  - We want tests to be snappy, but also not spend too much time optimising our code for them

Solution:
  - Use a small delay before clicking the continue button
  - If we know that the test run is a retry, assume that the error could have been this race condition and increase the delay before clicking continue

I think we could also look for a different element to indicate that the response is still streaming. I think that the "enqueue while moderating" will be active in these cases, but wouldn't be if we used a different element to determine a streaming response

### DIFF
--- a/apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
+++ b/apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
@@ -1,5 +1,5 @@
 import { setupClerkTestingToken } from "@clerk/testing/playwright";
-import { test, expect } from "@playwright/test";
+import { test, expect, Page } from "@playwright/test";
 
 import { TEST_BASE_URL } from "../../config/config";
 import { bypassVercelProtection } from "../../helpers/vercel";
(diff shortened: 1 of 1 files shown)


apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
```diff
@@ -18,73 +18,75 @@ import {
 // const FIXTURE_MODE = "record" as FixtureMode;
 const FIXTURE_MODE = "replay" as FixtureMode;
 
-test.describe(() => {
-  // NOTE(2024-10-10): This test is flaky, with the "10 of 10" check often returning 6-8 sections
-  // Remove extra retries when it becomes more stable
-  if (process.env.CI === "true") {
-    test.describe.configure({ retries: 4 });
-  }
```
codeincontext, line 26: This is the code that's changed. The changes below are just indentation

apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
```diff
@@ -18,73 +18,75 @@ import {
...
+  { tag: "@common-auth" },
+  async ({ page }, testInfo) => {
+    const generationTimeout = FIXTURE_MODE === "record" ? 75000 : 50000;
+    test.setTimeout(generationTimeout * 5);
+
+    // The chat UI has a race condition when you submit a message too quickly after the previous response
+    // This is a temporary fix to fix test flake
+    async function letUiSettle() {
+      return await page.waitForTimeout(testInfo.retry === 0 ? 500 : 6000);
+    }
```
codeincontext, line 32: This is the new solution to the problem. If the last test run failed, wait 6 seconds before clicking `continue` again

This means that we'll get a quick test run if the code under test works well, and if it doesn't we'll see that there was a retry in the test summary

stefl (approved):

stefl:
Should we add a data-test property that explicitly states the chat state so that we can test for it?

codeincontext:
> Should we add a data-test property that explicitly states the chat streaming state so that we can test for it?

I like that. We could expose the streaming state value so we can hook into whatever state we need "natively" rather than using UI as an analog

PR #204, opened 2024-10-08
refactor: demo account components to oak components - AI-460

JBR90:
## Description

- Refactors demo components to use Oak Components (excluding model)
- Refactors header to use Oak components 
- Fixes margin bug on help page (margin was too small with extra demo banner height)

These changes have had some subtle margin / font size changes due to Oak Components spacing and font type constraints.

## Issue(s)

Fixes #

## How to test

1. Go to https://oak-ai-lesson-assistant-61gyyikpr.vercel-preview.thenational.academy
Toggle isDemoUser true in clerk
Check banner desktop / mobile , check links are working , check pop up for first demo lesson / used all demo lessons up

## Screenshots

How it used to look (delete if n/a):
image
image
image
image

How it should now look:
image
image
image
image

## Checklist

- [ ] Manually tested across browsers / devices
- [ ] Considered impact on accessibility
- [ ] Does this PR update a package with a breaking change

PR #221, opened 2024-10-11
fix: revert Oak components for demo UI (#204)

codeincontext:
This reverts commit 9f7a3d6c0c8e0b5d432a33f338271fb8f7035280, reversing changes made to 6737c050f0a1e844fa974df6cac0d586a57c1e01.

We found a bug in a mobile dialog which we don't want to bring into the RC


mikeritson-oak (approved):
Some minor issues spotted with mobile layout and demo accounts, I will document these for review.

PR #224, opened 2024-10-14
build: deploy playwright reports to Vercel

codeincontext:
## Description

- Playwright results are viewable in an interactive report. You can see an overview of the run, and the detailed replay of each test including steps, assertions, failures, JS console, network, etc
- This report is static HTML and only works properly when served over HTTP(S)
- We've looked at various ways to host these alongside our test runs, and it looks like just pushing to vercel is the simplest approach. We already have user access etc set up there


mantagen (approved):
Fab!!

PR #228, opened 2024-10-15
chore: allow import from csv, and teething fixes

mantagen:
## Description

- allow import lessons from CSV
- add a small amount of testing
- a few fixes to teething issues i discovered when running a full ingest

### DIFF
--- a/apps/nextjs/.gitignore
+++ b/apps/nextjs/.gitignore
@@ -9,3 +9,6 @@ tests-e2e/.auth
 .env.sentry-build-plugin
 
 *storybook.log
+
+# Playwright artifacts
+playwright-report
--- a/packages/aila/package.json
+++ b/packages/aila/package.json
@@ -4,7 +4,6 @@
   "license": "MIT",
   "main": "./index.ts",
   "types": "./index.ts",
-  "type": "module",
   "scripts": {
     "check": "tsc --noEmit",
     "clean": "rm -rf .turbo node_modules",
--- a/packages/ingest/src/import-lessons/graphql/query.ts
+++ b/packages/ingest/src/import-lessons/graphql/query.ts
@@ -6,7 +6,11 @@ export const query = gql`
       limit: $limit
       offset: $offset
       distinct_on: lessonSlug
-      where: { videoTitle: { _is_null: false }, isLegacy: { _eq: false } }
+      where: {
+        videoTitle: { _is_null: false }
+        isLegacy: { _eq: false }
+        subjectSlug: { _neq: "english" }
+      }
     ) {
       oakLessonId: lessonId
       lessonSlug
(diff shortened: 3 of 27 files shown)


apps/nextjs/.gitignore
```diff
@@ -8,4 +8,7 @@ tests-e2e/.auth
 # Sentry Config File
 .env.sentry-build-plugin
 
-*storybook.log
\ No newline at end of file
+*storybook.log
+
+# Playwright artifacts
+playwright-report
```
codeincontext, line 14: 👍

packages/aila/package.json
```diff
@@ -4,7 +4,6 @@
   "license": "MIT",
   "main": "./index.ts",
   "types": "./index.ts",
-  "type": "module",
```
codeincontext, line 7: It might be worth pulling in @stefl as I know he had some trouble here getting the imports working across the CLI, tests, storybook, whatever tools we use

So long as the github checks pass I'm satisfied though

codeincontext (approved):

packages/ingest/src/import-lessons/graphql/query.ts
```diff
@@ -6,7 +6,11 @@ export const query = gql`
       limit: $limit
       offset: $offset
       distinct_on: lessonSlug
-      where: { videoTitle: { _is_null: false }, isLegacy: { _eq: false } }
+      where: {
+        videoTitle: { _is_null: false }
+        isLegacy: { _eq: false }
+        subjectSlug: { _neq: "english" }
```
stefl, line 12: This might be a gotcha. Needs a comment?

packages/aila/package.json
stefl: As long as we don't have to rename everything .mjs and start importing "something.js" then I'm happy too!

packages/ingest/src/openai-batches/batchLineCompletion.ts
```diff
@@ -1,30 +1,32 @@
...
   userPrompt: string;
-  responseSchema: z.Schema;
-  responseSchemaName: string;
+  responseFormat: ReturnType<typeof zodResponseFormat>;
 }) {
   return {
-    custom_id: lineId,
+    custom_id: customId,
     method: "POST",
     url: "/v1/chat/completions",
```
stefl, line 21: Is this correct? Don't we want the beta endpoint for structured outputs? Or are you not using it here?

stefl (approved):
A couple of questions. Do you want structured outputs in this? Might want to check that is the correct URL if so

packages/ingest/src/import-lessons/graphql/query.ts
mantagen: Good point. Eventually it should be a parameter so that whoever is kicking off an ingest can control filters on say subject or key stage

mantagen: I meant to take it out

PR #234, opened 2024-10-15
fix: refactor lesson snapshots for improved efficiency

mantagen:
## Description

- store a lesson snapshot after each generation
- this snapshot will then be referenced by both moderation and export services
- also gives us a clearer data flow (i.e. a snapshot after every assistant message)


apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+          chatId,
+          data,
+          messageId,
+        });
+        setChecked(true);
+      } catch (cause) {
+        const error = new Error("Failed to check for existing export", {
+          cause,
+        });
+        Sentry.captureException(error, {
```
stefl, line 36: Does this need some user feedback?

apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
```diff
@@ -0,0 +1,44 @@
+import { useCallback, useState } from "react";
+
+import * as Sentry from "@sentry/react";
+
+export const useExistenceCheck = <T>({
```
stefl, line 5: Potentially name this with a more explicit definition of what the existence check is for?

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -233,13 +233,11 @@ export class AilaChat implements AilaChatService {
 
   private accumulatedText() {
     const accumulated = this._chunks?.join("");
-    return accumulated;
+    return accumulated ?? "";
```
stefl, line 236: Does this have any unexpected effects? I could imagine an ||= somewhere!

packages/aila/src/features/snapshotStore/AilaSnapshotStore.ts
```diff
@@ -0,0 +1,52 @@
...
+    messageId,
+    lessonPlan,
+    trigger,
+  }: {
+    messageId: string;
+    lessonPlan: LooseLessonPlan;
+    trigger: LessonSnapshotTrigger;
+  }) {
+    invariant(this._aila.userId, "userId is required for saving snapshots");
+
```
stefl, line 42: Think you can remove these?

stefl (approved):

apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
mantagen: Yes it should, although I haven't changed functionality here, just refactored this out to reduce duplication. Willtake a look to see if there's a simple way to add user feedback but not wanting to get pulled into widening the scope of this

apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
mantagen: Yep sure. It's scoped within the exports folder, but agree it could have a better name

packages/aila/src/features/snapshotStore/AilaSnapshotStore.ts
mantagen: userId can be undefined at this point I believe?

packages/aila/src/core/chat/AilaChat.ts
mantagen: No effects, everywhere just checks is result is truthy.

apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
mantagen: Actually just seen that this doesn't actually block the user. It's a background check, and the user hits download and would see an error if there was a subsequent error

apps/nextjs/src/components/ExportsDialogs/useExistenceCheck.ts
mantagen: Done

PR #235, opened 2024-10-16
fix: attempt to allow cloudinary images more explicitly

stefl:
## Description

- In some cases we see Cloudinary images being blocked by CSP, so this attempts to add a more specific line to see if it resolves it

## Issue(s)

[549118](https://oak-national-academy.sentry.io/issues/549118/events/6e932464d02f434aacea7ea8c4680c61)


codeincontext:
If you enable `STRICT_CSP` locally then you should be able to test this locally

mantagen (commented):
I don't think this will make a difference as it's covered by the whole domain on the above line?

codeincontext:
Closing for now as we don't have a strong conviction that this will solve the problem, and if it was important we would have updated it by now

PR #242, opened 2024-10-16
fix(sec): bump next to 14.2.18

johnrobeds:
## Description

Update next and next/env to 14.2.18


stefl:
The version bump needs to be applied three places - this is in just one of the packages I think?

PR #244, opened 2024-10-16
chore: centralised, filterable logging

codeincontext:
A number of our existing dependencies like prisma use the [debug](https://www.npmjs.com/package/debug) library. This lets you set a namespace for a group of logs, and then filter which logs you want to see through the `DEBUG` env var. You can also change the filtered groups programmatically  

This should help us:
- Cut through the log noise in local dev
- Selectively log in production when needed
- See slow areas in the code

Example log:
CleanShot 2024-10-16 at 17 04 13@2x

Prisma supports the debug library, but then it shows debug-level information which we don't want. I've added a workaround for that

I've set DEBUG in doppler to :

Dev:
`ai:*,-ai:middleware:auth,-ai:aila:moderation:response,+ai:db`

_The `+ai:db` is superfluous, but personally I'm preferring -ai:db a lot of the time so it makes it easy to switch_

Staging/production
`ai:*,-ai:middleware:auth,-ai:aila:moderation:response,-ai:db`

We have an old logger for structured logging with pino. It's used in very few places. If we need to, we can set `baseDebug.log = structuredLogger.info`, though the format changes for the worse in the default configuration. There's also a [pino-debug](https://github.com/pinojs/pino-debug) package which seems to offer this


packages/logger/index.ts
```diff
@@ -1,27 +1,42 @@
...
+  | "feedback"
+  | "fixtures"
+  | "generation"
+  | "inngest"
+  | "judgements"
+  | "qd"
+  | "rag"
+  | "rate-limiting"
+  | "ui"
+  | "testing";
```
codeincontext, line 35: This list is really the crux of making this logging useful. I haven't given myself a lot of time to split this perfectly, so I'm open to suggestions about how we shake out the different parts of the project

For example, we're treating aila as a package with its own namespace, but we also have categories like `analytics` and `rag` on the main namespace

I'm very happy for these to evolve over time

packages/logger/legacyLogger.ts
```diff
@@ -0,0 +1,27 @@
...
+    },
+  },
+  // Pino's default key of `time` is ignored by datadog, so explicitly set
+  // `timestamp`
+  timestamp: () => `,"timestamp":"${new Date(Date.now()).toISOString()}"`,
+});
+
+export type Logger = typeof logger;
+
+export default logger;
```
codeincontext, line 27: This is just copied from index.ts, and is still being used (for now)

stefl (approved):

PR #247, opened 2024-10-18
fix: chat streaming and loading states (WIP)

stefl:
## Description

There's a lot in here. I suggest we break up into separate PRs once we've reviewed the overall approach.

For reference on the problems this is looking to address, see [this Notion page](https://www.notion.so/oaknationalacademy/Main-Structured-Output-10626cc4e1b180aaa99ed1ed2fdcbe5e?pvs=4).

- This addresses various problems we have had with the chat experience, prompt, data loading and lesson plan completion

UI/UX:
- Fixes some small issues with the chat forms where images were lazy-loaded (use hidden/inline-block instead of mounting/unmounting)
- Adds a new queuing scrolling manager that brings the completed items into view as soon as they are ready. The current behaviour distributes this logic amongst the DropDopSection components and it's unpredictable how they behave.
- Adjusts the behaviour of the Continue button so that we can use it to fill in any missing sections and make it clear to the LLM which sections we want to generate next
- Allows the front end to rescue from scenarios where there are missing sections. The "Continue" button now includes the text of the next section group section names or any missing sections which need to be generated. This should mean that you can always complete a lesson if it seems to get stuck or miss a section.
- Rather than "Working on it…" it displays the section that is being streamed so there is more feedback to the user

Code structure:
- Replaces the tempLessonPlan logic with a LessonPlanManager
- Improves type-safety by asserting the section keys we are using are valid keys in the lesson plan, rather than just plain strings

Streaming / persistence:
- Fixes detection of parts while content is streaming in
- Reduces how often we have to Zod parse and patch the incoming message patches down to ~once per valid message part
- Adds a "status":"complete" to each generated message part. We use this to detect when the actual content of the message part has finished streaming in.
- Adds an "iteration" property which increments each time the chat is persisted. This allows us not to overwrite the chat with a past version if for some reason we have a stale cache for a request or there is a race condition

Prompt:
- Alters the prompt based on feedback from Open AI to put your cacheable / most important instructions at the top of the prompt. This now filters out steps that have previous been completed so that you aren't repeatedly asked about basedOn etc.
- Fixes the basedOn problem (or at least makes it a lot better!) where Aila automatically picks a lesson to base the desired lesson on, but without the user choosing from the list of options
- Fixes the issue where there are sometimes long text responses
- Reorders the Aila prompt so that we have instructions about the intended task at the top and then at the end constraints about how to respond to users. This seems to fix issues around text length / inclusion of summaries etc.
- Adds in HB's recent prompt work to add definitions for missing sections
- Puts more emphasis on the Additional Materials and now generates content rather than "None"

Testing / local dev:
- Adjusts the tests to check that content comes back in section groups rather than individually
- Adds more data-test attributes so that we can check the streaming state and state of the lesson plan in tests
- Rewrites the Romans test to ensure that the chat comes back in groups and validates that the section count increases over time
- Adds a display for local dev to show you what state the chat is in and which section is streaming
- Removes some CSP warnings in local development / test

Todo

[ ] Make sure we can still load existing chats
[ ] Update Playwright tests to ensure the new flow works as expected in E2E tests
[ ] Find out the spec for when lesson plan sections should be popped down - there's some logic about mobile that I haven't quite understood with this
[ ] Find out the spec (or change it) for which sections should be displayed in groups. This seems to be buggy / incorrect logic at the moment

## Issue(s)

Fixes testing issues in [this Notion ticket](https://www.notion.so/oaknationalacademy/Main-Structured-Output-10626cc4e1b180aaa99ed1ed2fdcbe5e?pvs=4).

## How to test

1. Start a chat
2. While each section streams in, note that you get a spinner, the content appears once ready, and the scroll position updates
3. You'll see a new status display at the top left. This is only shown on local development and staging so you don't need to give feedback on it. It displays the current chat iteration which increments each time the chat is persisted, as well as the streaming status
4. Note that there is no longer a white flash where the lesson plan gets emptied and then reset
5. Note that as sections stream in you are shown each one in turn but only when it has text, rather than seeing a spinner and then text for each one
6. Scroll down while a long section is loading in and you'll see the working section with a spinner
7. Check that responses with and without choosing a basedOn work as expected
8. Check that sections are generated in groups as per the design. Eg. cycles all together rather than one at a time
9. Check that a "basedOn" is not automatically set based on the initial input
10. Check the language to make sure it is in line with the spec for chat responses being brief


apps/nextjs/next.config.js
```diff
@@ -54,6 +55,28 @@ const getConfig = async (phase) => {
        * @see https://docs.sentry.io/platforms/javascript/guides/nextjs/migration/v7-to-v8/#updated-sdk-initialization
        */
       instrumentationHook: true,
+      turbo: {
```
stefl, line 58: Mirrors our webpack setup for these aliases if we are using turbopack

apps/nextjs/next.config.js
```diff
@@ -138,37 +161,39 @@ const getConfig = async (phase) => {
 
 module.exports = getConfig;
 
-module.exports = withSentryConfig(module.exports, {
-  // For all available options, see:
-  // https://github.com/getsentry/sentry-webpack-plugin#options
+if (!process.env.TURBOPACK) {
```
stefl, line 164: Turbopack has trouble with sentry in hot reloading mode apparently, but this doesn't resolve it for me

apps/nextjs/package.json
```diff
@@ -8,7 +8,8 @@
     "check": "tsc --noEmit",
     "clean": "rm -rf .next .turbo node_modules",
     "dev": "concurrently \"pnpm dev-server\" \"node scripts/local-dev.mjs\"",
-    "dev-server": "pnpm with-env next dev --port 2525 | pino-pretty -C",
+    "dev-turbo": "pnpm with-env node scripts/increase-listeners.js next dev --port 2525 --turbo | pino-pretty -C",
```
stefl, line 11: Add a separate option for turbopack. Also increase the listeners to remove an error where we have too many

apps/nextjs/scripts/increase-listeners.js
```diff
@@ -0,0 +1,7 @@
+// Increase the limit of max listeners
```
stefl, line 1: Allows us to have more listeners since the project has lots of files that are touched each build. Fixes a "too many listeners" error in local dev

apps/nextjs/src/app/api/chat/route.ts
```diff
@@ -10,4 +10,10 @@ async function postHandler(req: NextRequest): Promise<Response> {
   return handleChatPostRequest(req, config);
 }
 
+async function getHandler(): Promise<Response> {
```
stefl, line 13: Fixes the /api/chat 405 error in local dev

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
```diff
@@ -15,36 +15,64 @@ export const useAilaStreamingStatus = ({
...
+        if (lastMessage.role === "user") {
+          status = "RequestMade";
+        } else if (content.includes(moderationStart)) {
+          status = "Moderating";
+        } else if (content.includes(`"type":"text"`)) {
+          status = "StreamingChatResponse";
+        } else if (content.includes(chatStart)) {
+          status = "StreamingLessonPlan";
+          // Extract the slug of the currently streaming section, if there is one
+          const pathMatches = [
```
stefl, line 51: Find the most recent "path" so we can deduce which section is streaming without having to parse the JSON object itself.

apps/nextjs/src/components/AppComponents/Chat/Chat/hooks/useAilaStreamingStatus.ts
```diff
@@ -15,36 +15,64 @@ export const useAilaStreamingStatus = ({
...
-    console.log("ailaStreamingStatus set:", ailaStreamingStatus);
-  }, [ailaStreamingStatus]);
+    console.log("ailaStreamingStatus set:", status);
+    if (streamingSection) {
+      console.log("streamingSection:", streamingSection);
+    }
+  }, [status, streamingSection]);
 
-  return ailaStreamingStatus;
+  return { status, streamingSection, streamingSections };
```
stefl, line 77: Adds the slug of the section that is streaming, as well as all sections that have or are streaming

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -62,6 +65,24 @@ const organiseSections = [
   { trigger: "additionalMaterials", dependants: ["additionalMaterials"] },
 ];
 
+function getSectionsToDisplay(
```
stefl, line 68: Pull out this logic so we know which sections will be available. I'm not convinced on this - won't this mean we sometimes see empty sections?

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -73,8 +94,22 @@ export const LessonPlanDisplay = ({
...
+  const {
+    lessonPlan: incomingLessonPlan,
+    ailaStreamingStatus,
+    lastModeration,
+    streamingSections,
+  } = chat;
+
+  useEffect(() => {
+    if (Object.keys(incomingLessonPlan).length > 0) {
+      console.log("Update with incoming lesson plan", incomingLessonPlan);
```
stefl, line 109: Just some logging - we could remove or have a flag? What's the status on having a logger for local dev?

apps/nextjs/src/components/AppComponents/Chat/chat-list.tsx
```diff
@@ -131,7 +131,7 @@ export const ChatMessagesDisplay = ({
   ailaStreamingStatus: AilaStreamingStatus;
   demo: DemoContextProps;
 }) => {
-  const { lessonPlan, isStreaming } = useLessonChat();
+  const { lessonPlan, isStreaming, streamingSection } = useLessonChat();
```
stefl, line 134: We now have the currently streaming section from the provider

apps/nextjs/src/components/AppComponents/Chat/chat-start-form.tsx
```diff
@@ -59,19 +59,16 @@ export function ChatStartForm({
         />
         <div className="absolute bottom-10 right-10 top-10 flex items-center justify-center ">
           <Tooltip>
+            <LoadingWheel visible={isSubmitting} />
```
stefl, line 62: Fixes a blank state where the loading wheel icon has not lazy loaded

apps/nextjs/src/components/AppComponents/Chat/chat-start.tsx
```diff
@@ -16,7 +16,7 @@ import { trpc } from "@/utils/trpc";
 import { useDialog } from "../DialogContext";
 import ChatPanelDisclaimer from "./chat-panel-disclaimer";
 import { ChatStartForm } from "./chat-start-form";
-import EmptyScreenAccordion from "./empty-screen-accordian";
+import EmptyScreenAccordion from "./empty-screen-accordion";
```
stefl, line 19: Typo

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -22,31 +23,42 @@ const DropDownSection = ({
 }: {
   objectKey: string;
   sectionRefs: Record<string, React.MutableRefObject<HTMLDivElement | null>>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  value: any;
+  value: LessonPlanSectionWhileStreaming;
```
stefl, line 26: Removes an "any" since we now have a type for the different sections

apps/nextjs/src/components/AppComponents/Chat/empty-screen-accordion.tsx
```diff
@@ -157,11 +157,12 @@ const AccordionItem = React.forwardRef<HTMLDivElement, AccordionItemProps>(
 AccordionItem.displayName = "AccordionItem";
 
 const AccordionTrigger = React.forwardRef<
-  HTMLDivElement,
+  HTMLButtonElement,
   AccordionTriggerProps
->(({ children, ...props }) => (
+>(({ children, ...props }, forwardedRef) => (
```
stefl, line 162: Fixes a missing forwardRef error

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -391,54 +393,54 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
       overrideLessonPlan,
       queuedUserAction,
       queueUserAction,
-      executeQueuedAction,
+      reload,
+      setInput,
+      stop,
+      streamingSection,
+      streamingSections,
+      toxicModeration,
```
stefl, line 443: Sorted alphabetically

apps/nextjs/src/lib/lessonPlan/LessonPlanManager.ts
```diff
@@ -0,0 +1,70 @@
+import {
```
stefl, line 1: New LessonPlanManager class

apps/nextjs/src/lib/lessonPlan/LessonPlanManager.ts
```diff
@@ -0,0 +1,70 @@
...
+  public applyPatchesFromMessage(message: Message): void {
+    if (message.role === "assistant") {
+      console.log("Applying patches from message", message.content);
+      const { validPatches } = extractPatchesFromMessage(message);
+      this.applyLocalPatches(validPatches);
+    }
+  }
+
+  public updateFromServer(serverLessonPlan: LooseLessonPlan): void {
+    // Update existing keys and add new ones
```
stefl, line 33: This means we are maintaining a consistent reference to the lessonPlan which should mean fewer full re-renders

apps/nextjs/src/middlewares/csp.test.ts
```diff
@@ -2,42 +2,10 @@ import { NextRequest } from "next/server";
...
-    for (let i = 0; i < array.length; i++) {
-      array[i] = i % 256; // Fill with predictable values
-    }
-    return array;
-  }),
-  subtle: {},
-  randomUUID: jest.fn(() => "mocked-uuid"),
-};
+// Mock the uuid module
+jest.mock("uuid", () => ({
```
stefl, line 6: Turbopack doesn't seem to support "crypto" so I've switched it out for a uuid

packages/aila/src/core/chat/PatchEnqueuer.ts
```diff
@@ -61,6 +61,7 @@ export class PatchEnqueuer {
       type: "patch",
       reasoning: "generated",
       value: { op: "add", path, value },
+      status: "complete",
```
stefl, line 64: This adds a final "complete" message on each patch. We might want this to be something else, but it's just a marker to say that the value is all there during streaming.

packages/aila/src/protocol/jsonPatchProtocol.ts
```diff
@@ -527,11 +535,25 @@ function tryParseJson(str: string): {
     }
     return { parsed, isPartial: false };
   } catch (e) {
-    // If parsing fails, assume it's partial
-    return { parsed: null, isPartial: true };
+    try {
```
stefl, line 538: This now allows us to use process patches that are part of the llmMessage even though the llmMessage itself is still streaming in

packages/aila/src/protocol/schema.ts
```diff
@@ -646,3 +649,13 @@ export const chatSchemaWithMissingMessageIds = z
 export type AilaPersistedChatWithMissingMessageIds = z.infer<
   typeof chatSchemaWithMissingMessageIds
 >;
+
+export type LessonPlanSectionWhileStreaming =
```
stefl, line 653: Add a type to represent all the different sections that could be part of a lesson plan

packages/aila/tsconfig.json
```diff
@@ -13,7 +13,11 @@
     "moduleResolution": "bundler",
     "esModuleInterop": true,
     "types": ["jest"],
-    "typeRoots": ["./../../node_modules/@types", "./src/types"],
+    "typeRoots": [
+      "./../../node_modules/@types",
+      "./src/types",
+      "./src/utils/language/*.d.ts"
```
stefl, line 19: Remove the direct import of the d.ts file for americanisms - this was causing a turbopack issue

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
codeincontext: Have a look at #244!

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -109,7 +144,24 @@ export const LessonPlanDisplay = ({
...
+
+  useEffect(() => {
+    if (
+      JSON.stringify(sectionsToDisplay) !==
+      JSON.stringify(prevSectionsToDisplay)
+    ) {
+      console.log("Render LessonPlanDisplay with sections", sectionsToDisplay);
+      setPrevSectionsToDisplay(sectionsToDisplay);
+    }
+  }, [sectionsToDisplay, prevSectionsToDisplay]);
```
codeincontext, line 161: If planning to keep this, it would be worth putting it in a hook in this file. It doesn't look like we need to use `prevSectionsToDisplay` in the render itself

I think `prevSectionsToDisplay` would be better as a ref, to avoid a second render

It's maybe a bit misleading to log "Render", as there are other reasons it would render. It's more that the sections have changed

apps/nextjs/src/components/AppComponents/Chat/chat-list.tsx
```diff
@@ -191,7 +193,9 @@ export const ChatMessagesDisplay = ({
                   ? {
                       id: "working-on-it-initial",
                       role: "assistant",
-                      content: "Working on it…",
+                      content: streamingSection
+                        ? `Editing ${streamingSection}…`
+                        : "Working on it…",
```
codeincontext, line 198: This could be really nice

apps/nextjs/src/components/AppComponents/Chat/chat-start-form.tsx
codeincontext: Ah, that's a nice approach

apps/nextjs/src/components/AppComponents/Chat/empty-screen-accordion.tsx
codeincontext: Ah lovely!

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -156,6 +148,10 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
   const hasAppendedInitialMessage = useRef<boolean>(false);
 
   const lessonPlanSnapshot = useRef<LooseLessonPlan>({});
+  const lessonPlanManager = useRef(new LessonPlanManager({}));
+  const [lessonPlan, setLessonPlan] = useState(
+    lessonPlanManager.current.getLessonPlan(),
+  );
```
codeincontext, line 154: 🎉

apps/nextjs/src/hooks/useTemporaryLessonPlanWithStreamingEdits.ts
```diff
@@ -23,7 +23,8 @@ const hashFor = (obj: object, messageHashes) => {
 };
```
codeincontext, line 23: Can we delete this file now?

codeincontext (commented):
Reviewed up to the LessonPlanManager class. I'll do that in a second pass

apps/nextjs/src/lib/lessonPlan/LessonPlanManager.ts
codeincontext: I do think that this can make the rendering less predictable and I think we can solve the rerendering problem with smarter use of react techniques like context and state selectors

This mutation loop is already in our code so I think it's fine to keep it here for this PR at the moment

apps/nextjs/src/middleware.ts
```diff
@@ -51,7 +51,7 @@ const nextMiddleware: NextMiddleware = async (request, event) => {
 
   try {
     response = await authMiddleware(request, event);
-    response = addCspHeaders(response, request, cspConfig);
+    response = await addCspHeaders(response, request, cspConfig);
```
codeincontext, line 54: It doesn't look like this returns a promise? I don't ever see the eslint rule about missing an await on a promise. If we get that working, we would be alerted the function started to be async and needed to be awaited

ps: Obviously this is for another PR when this one is stable and ready to merge

packages/aila/src/core/chat/PatchEnqueuer.ts
codeincontext: I might need some more context here. Can the status be any other value? How does this differ from waiting for the closing brace and a valid object? ie: isPartial=false from tryParseJson

It it just that it aligns with the way the frontend is looking up statuses?

packages/aila/src/protocol/jsonPatchProtocol.ts
codeincontext: I assume that this will make the parsed value change on every streamed token. Would this result in a new render, or does it go into some logic that condenses it down into a value we care about?

.eslintrc.cjs
```diff
@@ -13,6 +13,6 @@ module.exports = {
   plugins: ["@typescript-eslint"],
   extends: ["plugin:@typescript-eslint/recommended"],
   rules: {
-    "no-console": "error"
+    "no-console": "warn"
```
stefl, line 16: We still have lots of console.log in the codebase so I had to switch this back to warn

apps/nextjs/package.json
```diff
@@ -18,6 +19,8 @@
     "test": "pnpm with-env jest --colors --config jest.config.js",
     "test-e2e": "pnpm with-env playwright test",
     "test-e2e-ui": "pnpm with-env playwright test --ui",
+    "test-e2e-ui-serve": "pnpm build && pnpm start --port 4848 --turbo",
```
stefl, line 22: Attempting to run E2E tests on a built version of the app so that they are less flaky

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -41,7 +48,10 @@ const displayStyles = cva(
   "relative flex flex-col space-y-10 px-14 pb-28 opacity-100 sm:px-24 ",
 );
 
-const organiseSections = [
+const organiseSections: {
```
stefl, line 51: I'm still very unconvinced by this method. It assumes that the LLM always responds in chunks that match what it is expecting. If it sends back three sections when you're expecting four, you end up with a broken section that never loads and you can't interact with. I would vote for removing this and just using the sections that are present in the lesson plan.

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -62,54 +72,113 @@ const organiseSections = [
...
-  const chat = useLessonChat();
-  const { lessonPlan, ailaStreamingStatus, lastModeration } = chat;
+  const [openSections, setOpenSections] = useState<Record<string, boolean>>({});
+  const [sectionsToDisplay, setSectionsToDisplay] = useState<LessonPlanKeys[]>(
+    [],
+  );
 
-  const [userHasCancelledAutoScroll, setUserHasCancelledAutoScroll] =
-    useState(false);
+  const handleSetIsOpen = (section: string, isOpen: boolean) => {
```
stefl, line 119: Move all of the "isOpen" management to this level

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -62,54 +72,113 @@ const organiseSections = [
...
-      const timer = setTimeout(() => {
-        setUserHasCancelledAutoScroll(false);
-      }, 1000);
-      return () => clearTimeout(timer);
-    }
+    const newSectionsToDisplay = getSectionsToDisplay(
+      lessonPlanSectionKeys,
+      streamingSections,
+    );
+    setSectionsToDisplay((prevSectionsToDisplay) => {
```
stefl, line 151: Reduces some thrash about which sections to display. But if we were to just use the keys that are on the lesson or are streaming in, we could remove all this.

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanMapToMarkDown.tsx
```diff
@@ -4,6 +4,8 @@ import { LooseLessonPlan } from "@oakai/aila/src/protocol/schema";
 import { sectionToMarkdown } from "@oakai/aila/src/protocol/sectionToMarkdown";
 import { lessonSectionTitlesAndMiniDescriptions } from "data/lessonSectionTitlesAndMiniDescriptions";
 
+import { allSectionsInOrder } from "@/lib/lessonPlan/sectionsInOrder";
```
stefl, line 7: Uses the shared definition rather than having this repeated in multiple places.

apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
```diff
@@ -20,16 +20,18 @@ const ChatLhsHeader = ({
...
   return (
     <>
       <div className="mt-6 hidden items-center justify-end gap-5 sm:flex">
         {process.env.NEXT_PUBLIC_ENVIRONMENT !== "production" && (
-          <div
-            className="flex-grow text-left text-xs"
-            data-testid="chat-aila-streaming-status"
-          >
-            {chat.ailaStreamingStatus}
+          <div className="flex flex-grow flex-row space-x-4 text-left text-xs">
```
stefl, line 27: Adds some additional display for test / dev

apps/nextjs/src/components/AppComponents/Chat/chat-right-hand-side-lesson.tsx
```diff
@@ -159,6 +158,7 @@ const ChatRightHandSideLesson = ({
         </ChatButton>
       </span>
       <div ref={endOfDocRef} />
+      <div className="h-[50vh]"></div>
```
stefl, line 161: Adding this enables us to scroll newly inserted elements to the top of the scrolling container

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -1,118 +1,82 @@
...
-import { scrollToRef } from "@/utils/scrollToRef";
 
 import Skeleton from "../../common/Skeleton";
+import { AilaStreamingStatus } from "../Chat/hooks/useAilaStreamingStatus";
 import ChatSection from "./chat-section";
 
 const DropDownSection = ({
-  objectKey,
-  sectionRefs,
+  section,
```
stefl, line 19: Much of the logic from this component is moved to its parent

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -1,118 +1,82 @@
...
       $pv="inner-padding-xl"
       ref={sectionRef}
     >
-      <OakFlex $gap="all-spacing-2">
-        <OakBox>
-          {status === "empty" && <div className="w-14"></div>}
-          {status === "isStreaming" && <LoadingWheel />}
-          {status === "isLoaded" && <Icon icon="tick" size="sm" />}
-        </OakBox>
+      <FullWidthButton onClick={() => setIsOpen(section, !isOpen)}>
```
stefl, line 58: Make the icons clickable as well as the text

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -1,118 +1,82 @@
...
+        <OakFlex $gap="all-spacing-2">
+          <OakBox>
+            {!isStreaming && !isLoaded && <div className="w-14"></div>}
```
stefl, line 61: We no longer need to use state for its status - it is calculated each render

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/modify-button.tsx
```diff
@@ -65,13 +66,32 @@ const ModifyButton = ({
...
+    return String(value);
+  };
   const recordUserModifySectionContent = async () => {
     if (selectedRadio && lastAssistantMessage) {
       const payload = {
         chatId: id,
         messageId: lastAssistantMessage.id,
         sectionPath,
-        sectionValue,
+        sectionValue: prepareSectionValue(sectionValue),
```
stefl, line 94: After making some type improvements, this was complaining about potentially sending numbers, which is not supported, so we have to convert the value first.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -65,6 +67,14 @@ export type ChatContextProps = {
   queuedUserAction: string | null;
   queueUserAction: (action: string) => void;
   executeQueuedAction: () => Promise<void>;
+  streamingSection: LessonPlanKeys | undefined;
```
stefl, line 70: We now have more information on what is streaming

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -65,6 +67,14 @@ export type ChatContextProps = {
...
+  streamingSections: LessonPlanKeys[] | undefined;
+  streamingSectionCompleted: LessonPlanKeys | undefined;
+  sectionRefs: Record<
```
stefl, line 73: The provider now manages the refs to the sections that are displayed. We use this to manage the scroll behaviour.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -159,6 +159,11 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
   const hasAppendedInitialMessage = useRef<boolean>(false);
 
   const lessonPlanSnapshot = useRef<LooseLessonPlan>({});
+  const { lessonPlanManager, lessonPlan } = useLessonPlanManager();
+
+  const [currentIteration, setCurrentIteration] = useState<number | undefined>(
```
stefl, line 164: Each time we save the chat, we increment this number. We can then use it to compare the current lesson plan with what has come down the wire and only update it if it has incremented.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -343,33 +351,34 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
    *  Update the lesson plan if the chat has finished updating
    *  Fetch the state from the last "state" command in the most recent assistant message
    */
   useEffect(() => {
     if (!hasFinished || !messages) return;
-    trpcUtils.chat.appSessions.getChat.invalidate({ id });
+    const timeout = setTimeout(() => {
+      // Delay fetching the lesson plan to ensure the UI has updated
+      trpcUtils.chat.appSessions.getChat.invalidate({ id });
+    }, 10000);
```
stefl, line 363: If we're not careful, we update the chat before patches are applied.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -383,7 +392,37 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
+    status: ailaStreamingStatus,
+    streamingSection,
+    streamingSections,
+  } = useAilaStreamingStatus({ isLoading, messages });
+
+  const [streamingSectionCompleted, setStreamingSectionCompleted] = useState<
+    LessonPlanKeys | undefined
+  >(undefined);
+
+  const { sectionRefs, setSectionRef } = useLessonPlanScrollManagement(
```
stefl, line 405: This is the new scroll management hook

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -383,7 +392,37 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
+    streamingSection,
+    streamingSectionCompleted,
+  );
+
+  const workingMessage = useRef<Message | undefined>(undefined);
```
stefl, line 410: We have to snapshot the message so that we don't get a render loop

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -394,61 +433,63 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
 
   const value: ChatContextProps = useMemo(
     () => ({
+      ailaStreamingStatus,
```
stefl, line 436: Sorted alphabetically

apps/nextjs/src/hooks/useLessonPlanScrollManagement.ts
```diff
@@ -0,0 +1,155 @@
+import { useCallback, useEffect, useRef, useState } from "react";
```
stefl, line 1: A new hook that handles scrolling behaviour during streaming.

As new sections are completed it scrolls them into view using a queue.

apps/nextjs/tests-e2e/config/config.ts
```diff
@@ -5,7 +5,8 @@ dotenv.config({ path: path.join(process.cwd(), ".env") });
 
 export const TEST_USER_EMAIL = process.env.TEST_USER_EMAIL;
 export const TEST_USER_PASSWORD = process.env.TEST_USER_PASSWORD;
+export const PORT = process.env.PORT ?? 2525;
 export const TEST_BASE_URL =
-  process.env.TEST_BASE_URL || "http://localhost:2525";
+  process.env.TEST_BASE_URL || `http://localhost:${PORT}`;
```
stefl, line 10: Allow us to specify the port where the server is running. Eg. run the server using a built version of the app rather than the dev server

apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
```diff
@@ -59,32 +59,51 @@ test(
...
-      await waitForGeneration(page, generationTimeout);
-      await expectSectionsComplete(page, 10);
-      await letUiSettle();
-
-      setFixture("roman-britain-5");
-      await continueChat(page);
-      await waitForGeneration(page, generationTimeout);
-      await expectSectionsComplete(page, 10);
+
+      let prevSectionsComplete = 0;
```
stefl, line 63: This was proving quite brittle because the LLM sometimes asks again if you want to do something, or only generates one section instead of three.

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -40,6 +40,8 @@ export class AilaChat implements AilaChatService {
   private readonly _patchEnqueuer: PatchEnqueuer;
   private readonly _llmService: LLMService;
   private readonly _promptBuilder: AilaPromptBuilder;
+  private _iteration: number | undefined;
```
stefl, line 43: We now have an "iteration" number on the chat which increments each time we save it

packages/aila/src/core/chat/PatchEnqueuer.ts
stefl: This is due to the object being nested within the llmMessage, so it was hard to validate that it was a complete object. The best idea I had was to validate the overall object, untruncate it if it failed to validate, then find the substring where the object starts and then validate that. But that felt somewhat brittle. This way we can check that this field is present and we know then that the object is complete. It's like a checksum. It doesn't have any other potential value, so we could pick something else.

packages/aila/src/features/persistence/AilaPersistence.ts
```diff
@@ -48,7 +56,9 @@ export abstract class AilaPersistence {
       subject,
       keyStage,
       topic,
-      createdAt: Date.now(),
+      createdAt: createdAt ? createdAt.getTime() : Date.now(),
+      updatedAt: Date.now(),
+      iteration: iteration ? iteration + 1 : 1,
```
stefl, line 61: Increment the iteration field. Add updatedAt. Make createdAt stable.

packages/aila/src/protocol/jsonPatchProtocol.ts
stefl: Yes, I've now added logic that reduces the parsing down significantly - we only parse once a section is detected as complete.

packages/aila/src/protocol/schema.ts
```diff
@@ -501,24 +504,27 @@ export const LessonPlanSchemaWhilstStreaming = LessonPlanSchema;
...
-  | "misconceptions"
-  | "keywords"
-  | "basedOn"
-  | "starterQuiz"
-  | "exitQuiz"
-  | "cycle1"
-  | "cycle2"
-  | "cycle3"
-  | "additionalMaterials";
+export const LessonPlanKeysSchema = z.enum([
```
stefl, line 507: Make this a Zod schema too

apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
```diff
@@ -20,16 +20,18 @@ const ChatLhsHeader = ({
...
+            <div data-testid="chat-iteration">{chat.chat?.iteration}</div>
+            <div data-testid="chat-aila-streaming-status">
+              {chat.ailaStreamingStatus}
```
stefl, line 30: Add some additional output for testing and local dev

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -158,28 +204,19 @@ export const LessonPlanDisplay = ({
...
-                    userHasCancelledAutoScroll={userHasCancelledAutoScroll}
-                    documentContainerRef={documentContainerRef}
-                    showLessonMobile={showLessonMobile}
-                  />
-                );
-              }
-            })
-          );
-        })}
+        {allSectionsInOrder.map((section) => (
```
stefl, line 207: Rather than selectively rendering each section, we render them all with a visible attribute. This reduces thrash with the DIV refs so that when we try to scroll to each section during streaming, the DIV ref is not null. This comment is the result of several hours of work!

apps/nextjs/src/components/AppComponents/Chat/chat-list.tsx
```diff
@@ -131,13 +135,17 @@ export const ChatMessagesDisplay = ({
...
-  const { lessonPlan, isStreaming } = useLessonChat();
+  const { lessonPlan, isStreaming, streamingSection } = useLessonChat();
   const { setDialogWindow } = useDialog();
   const { totalSections, totalSectionsComplete } = useProgressForDownloads({
     lessonPlan,
     isStreaming,
   });
 
+  const workingOnItMessage = streamingSection
+    ? `Editing ${camelCaseToTitleCase(streamingSection)}…`
```
stefl, line 146: Display the section that is streaming

apps/nextjs/src/components/AppComponents/Chat/chat-message/index.tsx
```diff
@@ -160,9 +160,7 @@ export function ChatMessage({
               ailaStreamingStatus === "Moderating" ? (
                 <MemoizedReactMarkdownWithStyles markdown={"Finishing up..."} />
               ) : (
-                <MemoizedReactMarkdownWithStyles
-                  markdown={"Working on it ..."}
-                />
+                <MemoizedReactMarkdownWithStyles markdown={message.content} />
```
stefl, line 163: This was a bug - setting this text higher up had no effect

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.tsx
```diff
@@ -1,118 +1,86 @@
...
     <DropDownSectionWrapper
+      className={visible ? "block" : "hidden"}
       $borderColor="black"
       $bb="border-solid-m"
       $pv="inner-padding-xl"
       ref={sectionRef}
+      data-testid={`chat-section-${section}`}
+      data-test="lesson-plan-section"
+      data-test-section-key={section}
+      data-test-section-complete={isLoaded ? "true" : "false"}
```
stefl, line 60: Add more data-test attributes

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -286,6 +296,19 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
+  const generateContinueAction = useCallback(() => {
+    const sectionsToGenerate = nextSectionsToGenerate(lessonPlan);
+    if (sectionsToGenerate.length === 0) {
+      return "Continue - check the lesson plan for consistency and correctness";
+    } else {
+      const sentenceCaseSections = sectionsToGenerate
+        .map((section) => camelCaseToTitleCase(section))
+        .join(", ")
+        .replace(/, ([^,]*)$/, " and $1");
+      return `Continue. Generate the ${sentenceCaseSections} section${sectionsToGenerate.length > 1 ? "s" : ""}.`;
```
stefl, line 308: Doing this means that the front end should be more likely to rescue from missing sections. It will automatically trigger generation of the correct next section group OR try to fill in if something is missing.

apps/nextjs/src/hooks/useLessonPlanScrollManagement.ts
```diff
@@ -0,0 +1,155 @@
...
+    LessonPlanKeys | undefined
+  >(undefined);
+
+  const [userHasCancelledAutoScroll, setUserHasCancelledAutoScroll] =
+    useState(false);
+
+  const sectionRefs = useRef<
+    Record<LessonPlanKeys, React.MutableRefObject<HTMLDivElement | null>>
+  >({
+    title: { current: null },
```
stefl, line 22: Not so happy about having these hard-coded like this, but the lesson structure seems stable, so I didn't try to make this more complex than it needed to be

apps/nextjs/src/hooks/useLessonPlanScrollManagement.ts
```diff
@@ -0,0 +1,155 @@
...
+    (section: LessonPlanKeys, el: HTMLDivElement | null) => {
+      sectionRefs.current[section].current = el;
+    },
+    [],
+  );
+
+  const scrollQueueRef = useRef<LessonPlanKeys[]>([]);
+  const activeScrollRef = useRef<LessonPlanKeys | null>(null);
+
+  const processScrollQueue = useCallback(() => {
```
stefl, line 51: Adds queueing behaviour so that during streaming we don't end up with scrolling thrash

apps/nextjs/src/lib/lessonPlan/organiseSections.ts
```diff
@@ -0,0 +1,25 @@
+import { LessonPlanKeys } from "@oakai/aila/src/protocol/schema";
+
+export const organiseSections: {
```
stefl, line 3: I'm still unhappy with this. I don't think it's wise to assume that the LLM will always respond in groups, but this matches the current expected behaviour.

apps/nextjs/tests-e2e/tests/aila-chat/full-romans.mobile.test.ts
```diff
@@ -15,8 +15,8 @@ import {
 // --------
 // CHANGE "replay" TO "record" TO RECORD A NEW FIXTURE
 // --------
-// const FIXTURE_MODE = "record" as FixtureMode;
-const FIXTURE_MODE = "replay" as FixtureMode;
+const FIXTURE_MODE = "record" as FixtureMode;
```
stefl, line 18: This test is broken for me on local dev. It tries to connect using the "https://" scheme

packages/core/src/utils/camelCaseConversion.ts
```diff
@@ -0,0 +1,13 @@
+export function camelCaseToSentenceCase(str: string) {
```
stefl, line 1: I'd hoped to make this generally available but we have some circular dependencies

packages/core/src/prompts/lesson-assistant/parts/promptingTheUser.ts
```diff
@@ -0,0 +1,40 @@
+export const promptingTheUser =
```
stefl, line 1: New section in the prompt at the end which specifies the response format.

packages/core/src/prompts/lesson-assistant/parts/lessonComplete.ts
```diff
@@ -2,13 +2,7 @@ export const lessonComplete = () => `ONCE THE LESSON IS COMPLETE
...
 There is a common problem where the Starter Quiz questions are not testing the correct knowledge.
 Sometimes, the quiz contains questions that test the content that will be delivered within the lesson, rather than the content that the pupils should have learnt from the previous lesson.
-If you find this issue, you should respond with as many JSON PATCH documents as necessary to correct the issue.
-The lesson plan also needs to match the JSON Schema that is supplied.
-If it does not, you should respond with as many JSON PATCH documents to correct the issues with the data structure as needed to get it to be in the correct format.
-Also, for every cycle, make sure that all of the parts of the cycle have values.
-If they do not, generate instructions to set the missing sections of the cycle.
-For instance, for each cycle, ensure that it has at least two checks for understanding, as per the specification.`;
+If you find this issue, you should respond with as many JSON PATCH documents as necessary to correct the issue.`;
```
stefl, line 8: Remove unnecessary parts of the prompt now we have structured outputs

packages/core/src/prompts/lesson-assistant/parts/interactingWithTheUser.ts
```diff
@@ -1,144 +1,202 @@
...
-In most scenarios you will be provided with title, keyStage, subject, topic (optionally)in the lesson plan.
-If they are not present, ask the user for them.
-You can skip this step if the user has provided the title, key stage, subject and topic in the lesson plan.`,
-
-    hasRelevantLessons
-      ? `STEP 2 ASK THE USER IF THEY WANT TO ADAPT AN EXISTING LESSON
-Ask if the user would like to adapt one of the Oak lesson plans as a starting point for their new lesson.
+    relevantLessonPlans && relevantLessonPlans.length > 3; // TODO This is a string, not an array!
+
+  const steps: (LessonConstructionStep | undefined)[] = [
```
stefl, line 26: We now construct the steps for creating the lesson programatically. We unshift this and only pass the next-most instruction rather than all instructions. This seems to remove the issue where it is repeatedly talking about basedOn lessons.

packages/core/src/prompts/lesson-assistant/parts/endingTheInteraction.ts
```diff
@@ -1,17 +1,5 @@
 export const endingTheInteraction = () => `ENDING THE INTERACTION
 Once you have sent back all of the edits that you need to make to fulfil the request from the user, you should respond with an additional message with your next question for the user.
-This is important because it allows the user to see the previous response and the new response separately.
-Everything you send to the user should be in the format of a set of JSON documents.
-Do not send text before or after the set of JSON documents.
 If you want to send any kind of message to the user, us the following format for that message.
```
stefl, line 3: Much of this was no longer needed because of Structured Outputs

packages/core/src/prompts/lesson-assistant/parts/body.ts
```diff
@@ -8,21 +8,22 @@ A well-thought-out lesson plan should:
     keyStage ? ` at Key Stage ${keyStage}` : ""
   }.
 * Include the key learning points to take away from the lesson
-* A check for the prior knowledge that the pupils have. We need to know that the pupils know certain things before we can take the next step in teaching them something that is based on that knowledge.
+* Identify and check for pupils' prior knowledge during a starter quiz. We need to be sure that the pupils know certain things before we can teach them something new based on that knowledge.
```
stefl, line 11: Merging HB's recent work so that the Additional Materials does not default to "None" (see the Notion doc).

packages/core/src/prompts/lesson-assistant/index.ts
```diff
@@ -57,20 +58,21 @@ export const getPromptParts = (props: TemplateProps): TemplatePart[] => {
...
       : undefined,
+    body,
+    currentLessonPlan,
+    props.useRag ? rag : undefined,
+    props.baseLessonPlan ? basedOn : undefined,
     americanToBritishSection,
     languageAndVoice,
     props.isUsingStructuredOutput ? undefined : schema,
     response,
+    props.responseMode === "interactive" ? promptingTheUser : undefined,
```
stefl, line 75: This re-orders the prompt sections and breaks the response out, putting it last. This seems to fix a lot of the isues about the response text being long / incorrect

packages/aila/src/protocol/schema.ts
```diff
@@ -10,7 +10,9 @@ export const BasedOnSchema = z
       ),
     title: z.string().describe("The human-readable title of the lesson."),
   })
-  .describe("A reference to a lesson plan that this lesson is based on.");
+  .describe(
```
stefl, line 13: This was subtle. But adding this here seemed to fix the issue whereby it would automatically pick a basedOn without user input.

packages/aila/src/core/prompt/builders/AilaLessonPromptBuilder.ts
```diff
@@ -95,10 +95,7 @@ export class AilaLessonPromptBuilder extends AilaPromptBuilder {
       2,
     );
 
-    console.log(
-      "Got RAG content, length:",
-      stringifiedRelevantLessonPlans.length,
-    );
+    log.info("Got RAG content, length:", stringifiedRelevantLessonPlans.length);
```
stefl, line 98: One to pick up next. We are handing around a string here!

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -214,14 +224,18 @@ export class AilaChat implements AilaChatService {
...
+      await this._patchEnqueuer.enqueueMessage(message);
+    }
   }
 
   public async enqueuePatch(
     path: string,
     value: string | string[] | number | object,
   ) {
-    await this._patchEnqueuer.enqueuePatch(path, value);
+    if (this._patchEnqueuer) {
```
stefl, line 236: Fixes a server error where if the chat closes mid-stream, the server crashes

apps/nextjs/tests-e2e/tests/aila-chat/full-romans.test.ts
```diff
@@ -59,33 +66,109 @@ test(
...
+          { timeout: generationTimeout },
+        );
+        // Update the previous iteration
+        const chatIteration = await page.textContent(
+          '[data-testid="chat-iteration"]',
+        );
+        prevIteration = parseInt(chatIteration || "0", 10);
+
+        // Get the current sections complete
+        const currSectionsComplete = await getSectionsComplete(page);
```
stefl, line 111: This is more complex logic for checking the chat flow. Probably we move it to helpers?

packages/aila/src/core/chat/PatchEnqueuer.ts
codeincontext: Ok. Maybe we could say something like that on the type for future reference?

.eslintrc.cjs
codeincontext: Ah, interesting. I thought I'd caught them all on main. Though I was originally going to use warn so I'm happy either way

apps/nextjs/package.json
codeincontext: That could be very cool

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
codeincontext: It does seem slightly brittle to me. It's almost like you'd want the LLM to return with the names of the sections it's about to stream upfront in order for it to work nicely

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
codeincontext: These kind of comments would be useful in the code itself so we know why they're there when we next try to tidy it

I actually don't quite understand the render loop issue?

PR #248, opened 2024-10-18
feat: add dev-turbo mode to use turbopack

stefl:
## Description

- Adds a new `dev-turbo` pnpm command that uses Turbopack instead of Webpack in local development
- This speeds up compilation time and once we are happy with it, we could potentially make this the main way that we run the dev server
- There may be a remaining problem with the Error Boundary during hot reloading. If you encounter an issue saying that the module factory has gone away, you need to kill and restart the server.


apps/nextjs/src/middlewares/csp.ts
```diff
@@ -1,21 +1,14 @@
 import { NextRequest, NextResponse } from "next/server";
+import { v4 as uuidv4 } from "uuid";
```
stefl, line 2: There's an incompatibility somewhere with Turbopack not having access to "crypto". So instead it uses a UUID.

apps/nextjs/scripts/increase-listeners.js
```diff
@@ -0,0 +1,7 @@
+// Increase the limit of max listeners
+require("events").EventEmitter.defaultMaxListeners = 20;
+
+// Run the original command
+require("child_process").spawn(process.argv[2], process.argv.slice(3), {
+  stdio: "inherit",
+});
```
codeincontext, line 7: It would be useful to have context for why this is needed, and under what conditions we would be able to remove it

codeincontext (approved):

packages/aila/src/utils/language/findAmericanisms.ts
```diff
@@ -1,8 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
```
stefl, line 1: I couldn't get Typescript to find this file, and it's a bit of a distraction so using a /// here for now

PR #249, opened 2024-10-18
test: check streaming status directly

codeincontext:
In #247 @stefl introduced a useful element on screen to show the streaming state. We can use this for more reliable e2e testing

This PR just adds that early and updates the test helpers

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
@@ -2,6 +2,8 @@ import React from "react";
 
 import { useRouter } from "#next/navigation";
 
+import { useLessonChat } from "@/components/ContextProviders/ChatProvider";
+
 import AiIcon from "../../AiIcon";
 import ChatButton from "./ui/chat-button";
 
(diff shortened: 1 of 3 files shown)


stefl (approved):

apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
```diff
@@ -17,10 +19,19 @@ const ChatLhsHeader = ({
...
     <>
       <div className="mt-6 hidden items-center justify-end gap-5 sm:flex">
+        {process.env.NEXT_PUBLIC_ENVIRONMENT !== "production" && (
+          <div
+            className="flex-grow text-left text-xs"
+            data-testid="chat-aila-streaming-status"
+          >
+            {chat.ailaStreamingStatus}
+          </div>
+        )}
```
JBR90, line 34: Are we trying to move away from tailwind ? If so maybe this could be with oak components

JBR90 (approved):

apps/nextjs/src/components/AppComponents/Chat/chat-lhs-header.tsx
codeincontext: That's a good point, though in this case I grabbed it from another PR and don't want to cause a conflict there

PR #259, opened 2024-10-24
fix: reduce turbo errors on local dev, enable running e2e in built mode

stefl:
## Description
Incremental improvements to try to reduce turbo errors and test timeouts.

- Removes Sentry wrapper for turbo mode
- Adds some next.config settings for turbo mode
- Adds a script to run e2e on a built server to reduce timeouts in tests

Turbo mode:
- There seems to be a problem with Sentry whereby when you run the server in turbo mode, if you make changes, you'll see an error bounday issue about modules no longer being available in the factory
- The guidance from Next is to not uses the withSentry wrapper in turbo mode, and to add some settings to next.config
- This seems to have reduced the issue for me but not fixed it entirely
-
Running the e2e server in built mode could mean that:
- We can run a separate server for e2e tests locally, and continue working on the dev server while the tests are running
- Tests can run in parallel rather than one at a time


apps/nextjs/package.json
```diff
@@ -7,9 +7,10 @@
     "build:dev": "pnpm with-env next build",
     "check": "tsc --noEmit",
     "clean": "rm -rf .next .turbo node_modules",
-    "dev": "concurrently \"pnpm dev-server\" \"node scripts/local-dev.mjs\"",
-    "dev-turbo": "pnpm with-env node scripts/increase-listeners.js next dev --port 2525 --turbo | pino-pretty -C",
-    "dev-server": "pnpm with-env node scripts/increase-listeners.js next dev --port 2525 | pino-pretty -C",
+    "dev": "FORCE_COLOR=1 concurrently \"pnpm dev-server\" \"node scripts/local-dev.mjs\"",
+    "dev-turbo": "FORCE_COLOR=1 concurrently \"pnpm dev-server-turbo\" \"node scripts/local-dev.mjs\"",
```
codeincontext, line 11: For my benefit, what tool needs FORCE_COLOR? Is it because concurrently is getting in the way? Could you add it to `.env` like `DEBUG_COLORS`?

I'm not convinced that we actually need the concurrent local dev server any more. I haven't noticed any issues having it disabled in the last week

apps/nextjs/next.config.js
```diff
@@ -161,37 +159,41 @@ const getConfig = async (phase) => {
 
 module.exports = getConfig;
 
-module.exports = withSentryConfig(module.exports, {
-  // For all available options, see:
-  // https://github.com/getsentry/sentry-webpack-plugin#options
+if (!process.env.TURBOPACK) {
```
codeincontext, line 162: I can see that there's a benefit for not having the monitoring instrumentation in here. But what if we need to test sentry locally?

stefl: In that case, running the server in non-turbo dev mode would let you test it

stefl: The benefit is just to try to remove the sentry / turbo incompatibility

codeincontext: I think that we should only have the turbo mode as default, to be honest

PR #262, opened 2024-10-24
chore: hide generated fixtures and recordings in PRs

stefl:
## Description

- In PRs we see all of the generated fixtures and recordings we use in tests
- This adds a .gitattributes file which sets their status as generated
- Doing so automatically puts them into a hidden state in the PR
- You can still review them in the PR, but by default it shows them in a closed state with a notice that they are generated
- We could add other files to this as we go


codeincontext (approved):
I didn't know about this! This is great

PR #263, opened 2024-10-24
feat: add an incrementing iteration number to the chat

stefl:
## Description

- During testing we sometimes see the lesson plan reverting to a previous state and it's hard to debug why that is happening
- This adds an incrementing "iteration" attribute to the chat
- We can then display the iteration when we load and display data
- Also adds logging about which keys the lesson plan has before and after edits are applied
- Fixes an issue where the stream closes unexpectedly
- Sends the chat completion event only once persistence has completed


packages/aila/src/core/Aila.test.ts
```diff
@@ -320,7 +334,7 @@ describe("Aila", () => {
       expect(ailaInstance.lesson.plan.title).toBe("Mocked Lesson Plan");
       expect(ailaInstance.lesson.plan.subject).toBe("Mocked Subject");
       expect(ailaInstance.lesson.plan.keyStage).toBe("key-stage-3");
-    });
+    }, 8000);
```
codeincontext, line 337: Do we still need these extended timeouts if we're using the mock categoriser?

packages/aila/src/core/Aila.test.ts
```diff
@@ -365,9 +379,10 @@ describe("Aila", () => {
       // Use MockLLMService to generate a response
       await ailaInstance.generateSync({ input: "Test input" });
 
+      console.log("Generated");
```
codeincontext, line 382: Is this just for debugging?

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -94,6 +99,7 @@ export class AilaStreamHandler {
       await this.fetchChunkFromStream();
     } catch (error) {
       await this._chat.generationFailed(error);
+      throw error;
```
codeincontext, line 102: Nice find

packages/aila/src/core/lesson/AilaLesson.ts
```diff
@@ -47,6 +51,10 @@ export class AilaLesson implements AilaLessonService {
     this._plan = plan;
   }
 
+  public setPlan(plan: LooseLessonPlan) {
+    this._plan = plan;
+  }
```
codeincontext, line 56: I think that we could do with some better patterns for getters and setters. For just the plan, we need the class definition, setting the initial value in the initialiser, 2 setter methods and a getter

Something like [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties), but I think that would only work for non-destructured properties

codeincontext (approved):
I haven't given it a poke at runtime, but this looks like a good change 👍

PR #265, opened 2024-10-24
feat: hint Aila as to which part of the lesson to update next

stefl:
## Description

We've had several issues to do with the prompt that this address:

- Automatically picking a lesson for basedOn
- Not completing all sections in a section group
- Doing a whole interaction when picking a basedOn rather than starting the lesson once the choice has been made
- Incorrectly using basedOn logic when there is no content
- Not having any guidelines for some sections
- Language being too lengthy
- Responses not relating to the actually-generated sections

This addresses several of these issues and fixes a bug whereby the with-RAG workflow was kicking in even though there were no relevant lessons.

Includes HB's recent prompt changes for the missing sections and updates the body with her PR.

### DIFF
--- a/apps/nextjs/src/lib/lessonPlan/sectionsInOrder.ts
+++ b/apps/nextjs/src/lib/lessonPlan/sectionsInOrder.ts
@@ -1,0 +1,23 @@
+import { LessonPlanKeys } from "@oakai/aila/src/protocol/schema";
+
+export const allSectionsInOrder: LessonPlanKeys[] = [
+  "learningOutcome",
+  "learningCycles",
+  "priorKnowledge",
+  "keyLearningPoints",
+  "misconceptions",
+  "keywords",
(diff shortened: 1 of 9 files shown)


codeincontext (approved):
The code here looks good 👍 

The prompt changes are a bit too extensive to unpick in detail, and I'm not sure which changes are you vs HB

We should definitely get @mikeritson-oak to test these prompt changes ahead of merging - though if they've already been tested in the PR this is pulled from then we're good to go!

stefl:
> The code here looks good 👍
> 
> The prompt changes are a bit too extensive to unpick in detail, and I'm not sure which changes are you vs HB
> 
> We should definitely get @mikeritson-oak to test these prompt changes ahead of merging - though if they've already been tested in the PR this is pulled from then we're good to go!

Yep - these are the prompt changes that we've QAed on the main PR

PR #266, opened 2024-10-24
fix: logger to stdout for info and warn

mantagen:
## Description

- 'debug' defaults to stderr, meaning all logs look like errors
- this PR sets it to stdout, which should fix that

image

### DIFF
--- a/packages/logger/index.ts
+++ b/packages/logger/index.ts
@@ -7,6 +7,8 @@ if (typeof window !== "undefined") {
 }
 
 const debugBase = debug("ai");
+// By default debug logs to stderr, we want to use stdout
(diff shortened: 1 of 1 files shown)


packages/logger/index.ts
```diff
@@ -68,6 +68,9 @@ type ChildKey =
 export function aiLogger(childKey: ChildKey) {
   const debugLogger = debugBase.extend(childKey);
 
+  // By default debug logs to stderr, we want to use stdout
+  debugLogger.log = console.log.bind(console);
```
codeincontext, line 72: This works, but you may as well set it on `debugBase` and the child loggers should inherit it

codeincontext (approved):

PR #267, opened 2024-10-24
chore: run the local Postgres DB on a non-default Postgres port

stefl:
## Description

- Sets the Postgres port to 8432 instead of the default 5432 in local development
- This means you can have Postgres running on your machine without a clash with the Oak Docker instance

If/when we merge this we will need to update the DB connection string in Doppler and rerun `pnpm docker-reset` in the db package.


codeincontext (approved):
Oh yeah, I think this has tripped me up before!

PR #268, opened 2024-10-24
fix: remove flash when loading wheel appears

stefl:
## Description

- Removes the white flash when the loading wheel appears by adding a visible attribute


codeincontext (approved):
Nice improvement

PR #269, opened 2024-10-24
chore: csp support for turbo and add environment type

stefl:
## Description

- Alters the CSP header generation to support Turbo by using UUID instead of the crypto library (it has trouble with it)
- Adds a stricter definition of the environment name to save us getting mixed up between "prd" and "production"
- Fixes a bug that I think is in production right now
- Adds support for skipping the frame rules so that mobile tests run in dev

### DIFF
--- a/apps/nextjs/src/middlewares/csp.ts
+++ b/apps/nextjs/src/middlewares/csp.ts
@@ -2,24 +2,19 @@ import { NextRequest, NextResponse } from "next/server";
 import { v4 as uuidv4 } from "uuid";
 
 function generateNonce(): string {
-  if (typeof crypto !== "undefined" && crypto.getRandomValues) {
-    const array = new Uint8Array(16);
-    crypto.getRandomValues(array);
-    return btoa(String.fromCharCode.apply(null, array as unknown as number[]));
-  } else {
-    // Use uuid library to generate a random value
-    return uuidv4();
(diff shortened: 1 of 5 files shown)


apps/nextjs/src/middlewares/csp.ts
```diff
@@ -2,31 +2,27 @@ import { NextRequest, NextResponse } from "next/server";
...
 function generateNonce(): string {
-  if (typeof crypto !== "undefined" && crypto.getRandomValues) {
-    const array = new Uint8Array(16);
-    crypto.getRandomValues(array);
-    return btoa(String.fromCharCode.apply(null, array as unknown as number[]));
-  } else {
-    // Use uuid library to generate a random value
-    return uuidv4();
-  }
+  return uuidv4();
```
codeincontext, line 5: The uuid package says "Cryptographically-strong random values", but it would still sit slightly better with me if there was a way to keep using crypto in production. This feels like a bit of a downgrade, but I don't think it would matter in practicality

codeincontext (approved):

apps/nextjs/src/middlewares/csp.ts
stefl: I've since discovered "node:crypto" so perhaps we merge this and then investigate that?

PR #271, opened 2024-10-25
fix: module resolution with "node" instead of "bundler"

stefl:
## Description

This moves us back to using "node" instead of "bundler" for moduleResolution.

We moved to bundler to support mocking certain imports for use in Storybook, but "node" is the recommended approach for Next.js projects and was causing some difficulties in tests.

This replaces the module aliases approach with a Typescript approach.


codeincontext (approved):
Much better!

PR #272, opened 2024-10-25
fix: remaining linting fixes

stefl:
## Description

This is the remaining linting work that has since been merged through other PRs.

- By being more specific about imports and including types, this aims to reduce how much code has to be processed in builds, potentially reducing test run times and dev server start times, and potentially improve production performance because of smaller bundle sizes

Code conventions:
- Reintroduces our custom eslint rules with everything set to "warn". Somehow this had got disabled at some point.
- Adds `@typescript-eslint/consistent-type-imports` so that we import with `import type` where appropriate (and VS Code will automatically do this for us)

Aila package:
- Disables Americanism detection unless you externally initialise it
- Removes barrel imports and adjust Aila imports to import the appropriate package
- Fixes some tricky edge cases with the Aila package as a result (failing to await promises when in an error state in streaming, etc.)
- Allows us to pass in externally-initialised analytics adapters into the package at intialisation, so we don't always have to import all dependencies for every adapter

The touched files count looks high but it's primarily altering import statements throughout


codeincontext (commented):
This is good work

The import changes make sense to me. I don't think they will be fully effective unless we encourage us to continue using the pattern. I think we should remove any `export * from ...` code so our autosuggest doesn't get tempted to do the wrong thing

The aila changes are really good. I wonder if we will start to need some JSdoc comments on the options so that it's clear what the options are that you can provide when configuring it

Importing inngest at test runtime - not one I'm convinced by at the moment. First off, I don't understand the underlying reason why it would improve performance. Secondly, I don't know if it applies to other packages, other test files, or when we would be able to remove the fix. It would help to see that in a different PR so we can look at a before/ after and get a better understanding of what it's doing and if it's still needed

stefl:
> This is good work
> 
> The import changes make sense to me. I don't think they will be fully effective unless we encourage us to continue using the pattern. I think we should remove any `export * from ...` code so our autosuggest doesn't get tempted to do the wrong thing
> 
> The aila changes are really good. I wonder if we will start to need some JSdoc comments on the options so that it's clear what the options are that you can provide when configuring it
> 
> Importing inngest at test runtime - not one I'm convinced by at the moment. First off, I don't understand the underlying reason why it would improve performance. Secondly, I don't know if it applies to other packages, other test files, or when we would be able to remove the fix. It would help to see that in a different PR so we can look at a before/ after and get a better understanding of what it's doing and if it's still needed

Thanks - I've reverted the inngest import change. A good point about adding jsdoc. Before we go for that, I think we need to look again at the feature factory approach - irrespective of which adapter we use, we are automatically importing all dependencies for all adapters, which feels like something to fix in a separate piece of work. That could change the pattern for how we do this.

codeincontext:
> > This is good work
> > The import changes make sense to me. I don't think they will be fully effective unless we encourage us to continue using the pattern. I think we should remove any `export * from ...` code so our autosuggest doesn't get tempted to do the wrong thing
> > The aila changes are really good. I wonder if we will start to need some JSdoc comments on the options so that it's clear what the options are that you can provide when configuring it
> > Importing inngest at test runtime - not one I'm convinced by at the moment. First off, I don't understand the underlying reason why it would improve performance. Secondly, I don't know if it applies to other packages, other test files, or when we would be able to remove the fix. It would help to see that in a different PR so we can look at a before/ after and get a better understanding of what it's doing and if it's still needed
> 
> Thanks - I've reverted the inngest import change. A good point about adding jsdoc. Before we go for that, I think we need to look again at the feature factory approach - irrespective of which adapter we use, we are automatically importing all dependencies for all adapters, which feels like something to fix in a separate piece of work. That could change the pattern for how we do this.

That makes sense 👍

.vscode/settings.json
```diff
@@ -193,5 +201,6 @@
   "yaml.schemas": {
     "file:///c%3A/Users/Simon/.vscode/extensions/atlassian.atlascode-2.8.6/resources/schemas/pipelines-schema.json": "bitbucket-pipelines.yml",
     "https://www.artillery.io/schema.json": []
-  }
+  },
+  "typescript.preferences.preferTypeOnlyAutoImports": true
```
stefl, line 205: This will automatically add import type { something } where appropriate

codeincontext (approved):

PR #274, opened 2024-10-28
chore: make turbopack the default for local dev

codeincontext:
## Description

Building with turbopack has significantly improved our developer experience, but we found that sentry was incompatible. After a couple of weeks it seems like that's our only hurdle

- The latest version of sentry detects turbo and emits the following warning:
  -  > [@sentry/nextjs] WARNING: You are using the Sentry SDK with `next dev --turbo`. The Sentry SDK doesn't yet fully support Turbopack. The SDK will not be loaded in the browser, and serverside instrumentation will be inaccurate or incomplete. Production builds without `--turbo` will still fully work. If you are just trying out Sentry or attempting to configure the SDK, we recommend temporarily removing the `--turbo` flag while you are developing locally. Follow this issue for progress on Sentry + Turbopack: https://github.com/getsentry/sentry-javascript/issues/8105. (You can suppress this warning by setting SENTRY_SUPPRESS_TURBOPACK_WARNING=1 as environment variable)
- I've added a `dev:sentry` command to the nextjs package.json so that we can test sentry without turbo if we need


apps/nextjs/next.config.js
```diff
@@ -155,41 +155,39 @@ const getConfig = async (phase) => {
...
+  disableLogger: true,
+
+  // Enables automatic instrumentation of Vercel Cron Monitors. (Does not yet work with App Router route handlers.)
+  // See the following for more information:
+  // https://docs.sentry.io/product/crons/
+  // https://vercel.com/docs/cron-jobs
+  automaticVercelMonitors: true,
+});
 
 module.exports.onRequestError = Sentry.captureRequestError;
```
lforst, line 193: Hi, randomly stumbled on this PR. You put this line in the wrong location. It needs to go in instrumentation.ts.

codeincontext: My apologies for generating a notification on your PR!

Thank you for this, will fix 🎉

lforst: All good 😄

stefl (approved):

PR #275, opened 2024-10-28
chore: use DEBUG logging var on the frontend

codeincontext:
The `DEBUG` env var has helped me to keep local server logs clear and help me focus on specific areas I'm working on. At the moment that variable isn't configured on the frontend

This PR forwards the value in `DEBUG` to the frontend as well so we can filter logs there

### DIFF
--- a/packages/logger/index.ts
+++ b/packages/logger/index.ts
@@ -1,13 +1,12 @@
 import debug from "debug";
+import invariant from "tiny-invariant";
 
 import browserLogger from "./browser";
 import type { StructuredLogger } from "./structuredLogger";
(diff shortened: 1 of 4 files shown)


packages/logger/index.ts
```diff
@@ -3,7 +3,10 @@ import debug from "debug";
 import structuredLogger, { StructuredLogger } from "./structuredLogger";
 
 if (typeof window !== "undefined") {
-  debug.enable("ai:*");
+  if (!process.env.NEXT_PUBLIC_DEBUG) {
```
stefl, line 6: I've been using tiny invariant for this

stefl (approved):

PR #282, opened 2024-10-28
fix: use prisma generate --no-engine and reinstate prompts script

stefl:
## Description

- In Vercel builds where we are using Prisma Accelerate we should be using the --no-engine flag for `prisma generate`
- I think that potentially in the move to the new repo / Vercel project, or somewhere prior to that we seem to have lost this configuration flag
- This adds a new turbo command `db-generate:no-engine` which is in the dependsOn array for `turbo build`
- We also seem to have lost the prompt generation script so we are possibly reliant on the application to create new prompt variants at runtime instead of build time. This also reinstates that script in the build pipeline
- In the prompt variant generation script, this also now upserts the "app" rows if they are missing in local dev


packages/core/src/models/promptVariants.ts
```diff
@@ -46,9 +46,16 @@ export class PromptVariants {
 
     const { name, appId: appSlug, inputSchema, outputSchema } = this.definition;
 
-    const app = await this.prisma.app.findFirstOrThrow({
-      where: { slug: appSlug },
-    });
+    let app: App | undefined = undefined;
+
+    try {
+      app = await this.prisma.app.findFirstOrThrow({
```
stefl, line 52: Add a log if we are trying to reference an app that doesn't exist

packages/core/src/scripts/setupPrompts.ts
```diff
@@ -4,8 +4,32 @@ import { PromptVariants } from "../models/promptVariants";
 import { lessonPlannerPrompts, quizGeneratorPrompts } from "../prompts";
 import { ailaGenerate } from "../prompts/lesson-assistant/variants";
 
+export const apps = [
```
stefl, line 7: Move this in from the redundant seeds script

turbo.json
```diff
@@ -21,9 +25,21 @@
       "interactive": false,
       "cache": false
     },
+    "build:dev": {
```
stefl, line 28: This allows you to build a production instance of the application locally. If we only use prisma --no-engine then it won’t be able to connect to your local DB because it won’t have the postgres client.

package.json
```diff
@@ -21,8 +21,8 @@
     "doppler:run:stg": "doppler run -c stg --silent",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "lint": "turbo lint",
-    "prompts": "turbo prompts --filter=core",
-    "prompts:dev": "turbo prompts:dev --filter=core",
+    "prompts": "turbo prompts",
```
stefl, line 24: "core" doesn't exist any more. Removing

packages/core/package.json
```diff
@@ -23,9 +23,9 @@
     "process:lessons": "pnpm doppler:run:dev -- ts-node --compiler-options {\\\"module\\\":\\\"CommonJS\\\"} ./src/scripts/processLessons.ts",
     "process:lessons:dev": "pnpm doppler:run:dev -- pnpm process:lessons",
     "prompts": "ts-node --compiler-options {\\\"module\\\":\\\"CommonJS\\\"} ./src/scripts/setupPrompts.ts",
-    "prompts:dev": "pnpm doppler:run:dev -- ts-node --compiler-options {\\\"module\\\":\\\"CommonJS\\\"} ./src/scripts/setupPrompts.ts",
+    "prompts:dev": "dotenv -e ../../.env -- ts-node --compiler-options {\\\"module\\\":\\\"CommonJS\\\"} ./src/scripts/setupPrompts.ts",
```
stefl, line 26: This allows you to use your own postgres database locally, otherwise it is overridden by doppler

codeincontext: Good call. I hadn't considered that

packages/core/src/models/promptVariants.ts
codeincontext, line 52: I have an opinion about this particular pattern, but it's the norm for the project so it doesn't reflect on this PR. I'll share on slack as it's part of a wider conversation

codeincontext (approved):

PR #286, opened 2024-10-29
chore: remove feature flag and reduce id to 5characters

tomwisecodes:
## Description

- removes feature flag on donwload all button
- shortens id in the download name


apps/nextjs/src/app/api/aila-download-all/route.ts
```diff
@@ -152,7 +152,7 @@ async function getHandler(req: Request): Promise<Response> {
 
         const { data } = res;
 
-        const filename = `${lessonTitle} - ${lessonExport.id} - ${getReadableExportType(
+        const filename = `${lessonTitle} - ${lessonExport.id.slice(0, 5)} - ${getReadableExportType(
```
codeincontext, line 155: Have you sanity checked what the ids look like? I know some of the IDs around exports are UUIDs which can have the same prefix

mikeritson-oak (approved):
All looks good with the changes in this PR. 👍 
We had some issues but it was due to checks being done on the old branch rather than this one. All now cleared up.

codeincontext (approved):

PR #295, opened 2024-10-29
fix: linting for app components

stefl:
## Description

- Linting fixes for the AppComponents directory
- Small typescript fixes to the same
- Renames ShareButttonGroup.tsx (extra t)


apps/nextjs/src/components/AppComponents/common/ShareButtonGroup.tsx
```diff
@@ -0,0 +1,66 @@
+import { Box } from "@radix-ui/themes";
```
stefl, line 1: Typo in the filename

codeincontext: Is this a moved file?

stefl: Yes

codeincontext: It's just showing as an addition in the diff

stefl: I've removed and readded using git mv. Should be correct now

codeincontext (approved):

PR #299, opened 2024-10-29
fix: ensure that flag and modify button metadata uses strings (WIP)

stefl:
## Description

- While refactoring and adding better type constraints, I discovered that numbers are not allowable in the modifySection.sectionValue type
- This stores numbers as strings if we have attempted to set a section value to a number
- Waiting to merge other PRs to add this


stefl:
Closing - this is old. We might want to revisit

PR #306, opened 2024-10-29
fix: add missing dependencies to Analytics Provider

stefl:
## Description

- Adds missing dependencies to the Analytics Provider


codeincontext (approved):
Approving on the assumption that `posthogAiBeta.client` is stable

PR #311, opened 2024-10-29
fix: add missing dependencies to drop down form wrapper

stefl:
## Description

- Add missing dependencies to drop down form wrapper


codeincontext (approved):

codeincontext (commented):
Approved. It would be useful to confirm that the listeners aren't remounting on each render with the new change

PR #313, opened 2024-10-29
fix: await handling stream errors + try/catch

stefl:
## Description

- Handles an edge case where handling the stream error might itself throw
- Also ensures we await the handling of stream errors

### DIFF
--- a/packages/aila/src/core/chat/AilaStreamHandler.ts
+++ b/packages/aila/src/core/chat/AilaStreamHandler.ts
@@ -44,7 +44,12 @@ export class AilaStreamHandler {
         await this.readFromStream();
       }
     } catch (e) {
-      this.handleStreamError(e);
+      try {
+        await this.handleStreamError(e);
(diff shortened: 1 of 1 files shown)


packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -44,7 +44,12 @@ export class AilaStreamHandler {
         await this.readFromStream();
       }
     } catch (e) {
-      this.handleStreamError(e);
+      try {
+        await this.handleStreamError(e);
+      } catch (error) {
+        this._chat.aila.errorReporter?.reportError(error);
```
codeincontext, line 50: Beware that there is a ` this._chat.aila.errorReporter?.reportError(` within `handleStreamError`. It means it could double report an error if it's a non-helicone error object

stefl:
Closing this one because it feels like an edge case and hard to replicate

PR #314, opened 2024-10-29
fix: aila initialisation without await (possibly cancelling)

stefl:
## Description

- The Aila initialisation step does not need to be async/await


codeincontext (approved):

stefl:
I think this is actually a symptom of a bug - this initialise method is async and should be called here, so I will verify and close if so

PR #318, opened 2024-10-29
fix: add types for SVG imports, remove unused

stefl:
## Description

- Adds a type for the SVG imports
- Removes unused inline SVG icons


apps/nextjs/src/components/Icon/svgs.tsx
```diff
@@ -73,77 +73,86 @@ import upload from "@/assets/svg/upload.svg";
...
-  padlock: padlock,
-  "ai-lesson": aiLesson,
+type SVGImport = {
+  src: string;
+  height: number;
+  width: number;
+  blurDataURL: string;
+  blurWidth: number;
+  blurHeight: number;
+};
```
codeincontext, line 83: This approach does work, but there should be a way to set the type for svg imports. I know there is in webpack, not sure how to do it with turbopack

codeincontext (approved):

codeincontext:
Merging this one as it's passing tests and is stale

PR #320, opened 2024-10-30
fix: bootstrap posthog feature flags

codeincontext:
We use posthog to set feature flags

We can use feature flags for:
- Giving specific users access to a features
  - internal, beta users, etc
- Toggling features independently of code releases
  - merging work to features in development, instead of ling-lived feature branches
  - merging work that hasn't been QA'ed yet
  - disabling features which have failed QA or had a problem in production

### Problem

When we tried a feature flag for download-all, we found that they weren't working if you hadn't given cookie consent. I spoke to posthog and they said that feature flags should work with `opt_out_capturing` enabled. It sounds like a quirk of the way we're interacting with the consent banner

### Solution

The most robust way to load flags from posthog is to [bootstrap](https://posthog.com/docs/feature-flags/bootstrapping) them from the server. Then they will be immediately available on the frontend rather than relying on a frontend call

We don't really want to delay page load with the posthog call. We could cache it in KV for X hours, but then we would need to clear the cache when toggling a flag. Another option is [local flag evaluation](https://posthog.com/docs/feature-flags/local-evaluation)

With local flag evaluation, the posthog client periodically fetches the definitions for the flags, and will evaluate them for each user on the server. Because we're using edge functions, we need to cache those definitions so I've created a fetch client to handle that

If we also want to bootstrap flags that look at properties like the user's email address then we will need a follow up to make that available when we bootstrap. I have ideas for that, but one step at a time...


packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const KV_KEY = "posthog-feature-flag-local-evaluation";
+const ONE_DAY = 24 * 60 * 60 * 1000;
+const ONE_MINUTE = 60 * 1000;
+
+const log = aiLogger("analytics:feature-flags");
+
+const setKv = async (response: Response) => {
+  const value = await response.text();
+  await kv.set(KV_KEY, value, { ex: ONE_MINUTE });
```
stefl, line 13: Does this need `kv.set<some type>`?

packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const getKv = async () => {
+  const value = await kv.get(KV_KEY);
```
stefl, line 17: And here I think if we use `kv.get<some type>` we can validate it matches the type? I'm not sure where I saw this, but perhaps look up whether this is possible?

packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (url.includes("/decide")) {
+    log.warn("WARN: feature flag loaded through API");
+  }
+
+  return await fetch(url, options);
+};
+
+export const featureFlagsPollingInterval =
+  process.env.NODE_ENV === "production"
```
stefl, line 54: Another candidate for a isEnvironment helper?

stefl (approved):

packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
codeincontext: Possibly, though I wonder if we could get a similar result if we have stricter types on the env vars

packages/core/src/analytics/posthogAiBetaServerClient/index.ts
```diff
@@ -0,0 +1,26 @@
+import { PostHog } from "posthog-node";
+
+import {
+  cachedFetch,
+  featureFlagsPollingInterval,
+} from "./featureFlagEvaluation";
+
+const host = process.env.NEXT_PUBLIC_POSTHOG_HOST as string;
+const apiKey = process.env.NEXT_PUBLIC_POSTHOG_API_KEY || "*";
```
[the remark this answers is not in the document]
codeincontext: I usually do too, but I wonder if there's a greater chance of an empty string on an environment variable?

packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
codeincontext: The thing is, I don't know if it matters to us what the shape is if we're just caching it. Posthog's own code that calls fetch doesn't actually put a type on it either, so it could be a bit of extra effort for something that we don't need to validate and could drift from the type

https://github.com/PostHog/posthog-js-lite/blob/b853833c126ba713eb25696c96a3f0a6193ea2f0/posthog-node/src/feature-flags.ts#L422

packages/core/src/analytics/posthogAiBetaServerClient/featureFlagEvaluation.ts
codeincontext: I did have a look, but I haven't seen it in the docs. I guess for runtime validation it would need to be an argument, and then it could use something like zod. But that could be extra effort for what we actually need

PR #324, opened 2024-10-30
fix: design-changes-to-footer

tomwisecodes:
## Description
- matches https://www.notion.so/oaknationalacademy/Footer-Mobile-Footer-010ae9ef0d8f4a49bc04d0fa238bd348


apps/nextjs/src/app/layout.tsx
```diff
@@ -94,7 +94,7 @@ export default async function RootLayout({
             grayColor="olive"
             scaling="110%"
             color="#22222"
```
codeincontext, line 96: Should we delete this line, or explain why it's still needed if overridden below?

codeincontext:
CleanShot 2024-11-04 at 11 15 16@2x

It looks like every link is an external link now

codeincontext (approved):

PR #329, opened 2024-10-31
feat: use featureFlagGroup claim for feature flags

codeincontext:
In #320 we added local evaluation of feature flags. Local evaluation calculates the feature flags for the user, but only has user data from the session token. That works for feature flags which are either on or off, but we sometimes want feature flags for a specific group of users, like the AI team or external testers

This work:
- Adds featureFlagGroup to the labs Clerk metadata. We can manually populate this with `ai-team` for the team
- Passes the `featureFlagGroup` into the feature flag evaluation, so that we can calculate the user's flags without needing their email address

Testing:
- See that the [download-all-button](https://eu.posthog.com/project/8837/feature_flags/35349) feature flag is enabled for the group `ai-team`
- Log in with a user with `ai-team` set in clerk metadata (I've set it for all of our Oak emails)
- Load the homepage
  - Look for `Bootstrapping feature flags { 'download-all-button': true }` in the server log
- Sign in with another account
  - Look for `Bootstrapping feature flags { 'download-all-button': false }` in the server log

### DIFF
--- a/apps/nextjs/src/lib/feature-flags/bootstrap.ts
+++ b/apps/nextjs/src/lib/feature-flags/bootstrap.ts
@@ -32,15 +32,21 @@ function getDistinctIdFromCookie(headers: ReadonlyHeaders) {
 }
 
 export async function getBootstrappedFeatures(headers: ReadonlyHeaders) {
-  const { userId } = auth();
+  const { userId, sessionClaims } = auth();
 
   const distinctId = userId ?? getDistinctIdFromCookie(headers) ?? "0";
   log.info("Evaluating feature flags for", distinctId);
+
+  const personProperties = sessionClaims?.labs?.featureFlagGroup
+    ? { featureFlagGroup: sessionClaims.labs.featureFlagGroup }
+    : undefined;
+
   const features = await posthogAiBetaServerClient.getAllFlags(distinctId, {
-    // Only bootstrap flags which don't depend on user properties
-    // These are typically flags representing new features
+    // Only bootstrap flags which don't depend on PII
     onlyEvaluateLocally: true,
+    personProperties,
   });
+
   log.info("Bootstrapping feature flags", features);
   return features;
 }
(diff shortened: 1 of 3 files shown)


apps/nextjs/src/lib/feature-flags/bootstrap.ts
```diff
@@ -32,15 +32,21 @@ function getDistinctIdFromCookie(headers: ReadonlyHeaders) {
...
+
   const features = await posthogAiBetaServerClient.getAllFlags(distinctId, {
-    // Only bootstrap flags which don't depend on user properties
-    // These are typically flags representing new features
+    // Only bootstrap flags which don't depend on PII
     onlyEvaluateLocally: true,
+    personProperties,
   });
+
   log.info("Bootstrapping feature flags", features);
```
stefl, line 50: Worth logging some of the new properties here?

apps/nextjs/src/lib/feature-flags/bootstrap.ts
```diff
@@ -32,15 +32,21 @@ function getDistinctIdFromCookie(headers: ReadonlyHeaders) {
 }
 
 export async function getBootstrappedFeatures(headers: ReadonlyHeaders) {
-  const { userId } = auth();
+  const { userId, sessionClaims } = auth();
 
   const distinctId = userId ?? getDistinctIdFromCookie(headers) ?? "0";
   log.info("Evaluating feature flags for", distinctId);
+
+  const personProperties = sessionClaims?.labs?.featureFlagGroup
```
stefl, line 40: Is labs already decided? I believe we're using "aila" as the name for everything now

stefl (approved):

apps/nextjs/src/lib/feature-flags/bootstrap.ts
codeincontext: Yeah, we already have a number of keys under labs. It comes from when the teacher squad started their proof of concept for Clerk. They have `owa` and we have `labs`. There's no reason why we can't change it, but we'd need some code to read both for a time

apps/nextjs/src/lib/feature-flags/bootstrap.ts
codeincontext: Nice idea. Will do

PR #330, opened 2024-10-31
feat: sync featureFlagGroup from clerk to posthog

codeincontext:
#329 adds the ability to set feature flags based on a user's featureFlagGroup

This PR syncs changes in clark metadata to Posthog, so that posthog has an up-to-date version of featureFlagGroup

It's probably inevitable that we need to receive webhooks from Clerk, so I think this could prove useful in the long run as well


apps/nextjs/src/app/api/webhooks/clerk/route.ts
```diff
@@ -0,0 +1,95 @@
...
+  return primaryEmail.email_address;
+}
+
+async function syncUserToPosthog(user: UserJSON) {
+  const featureFlagGroup = user.public_metadata.labs?.featureFlagGroup ?? "";
+  posthogAiBetaServerClient.identify({
+    distinctId: getPrimaryEmail(user),
+    properties: { featureFlagGroup },
+  });
+  await posthogAiBetaServerClient.shutdown();
```
stefl, line 38: I think I learnt that we just need to do .flush() for this, otherwise it shuts it down for anything else that imports it? "Shutdown should only be called once per request" I think was the guidance.

apps/nextjs/src/app/api/webhooks/clerk/route.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    // Create a new Svix instance with your secret.
+    invariant(WEBHOOK_SECRET, "Missing CLERK_WEBHOOK_SECRET");
+    const wh = new Webhook(WEBHOOK_SECRET);
+
+    let evt: WebhookEvent;
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const payload = await req.json();
```
stefl, line 63: Presumably we don't have a Zod schema available to us for this?

apps/nextjs/src/app/api/webhooks/clerk/route.ts
```diff
@@ -0,0 +1,95 @@
...
+        return new Response("Updated", { status: 200 });
+      }
+      default:
+        log.error("Unknown event type:", eventType);
+        return new Response(`Unknown event type: ${eventType}`, {
+          status: 400,
+        });
+    }
+  } catch (error) {
+    Sentry.captureException(error);
```
stefl, line 92: Do we want a breadcrumb or metadata here about the shape of the incoming data?

stefl (approved):

apps/nextjs/src/app/api/webhooks/clerk/route.ts
stefl: Although it looks like you're only using it here so maybe not important

codeincontext: Yeah, the actual goal here is to flush the events, so that makes sense. It would be much more intuitive

codeincontext: I'll change it to a flush

apps/nextjs/src/app/api/webhooks/clerk/route.ts
codeincontext: I haven't found one provided by Clerk. You'll see a couple of lines below that they provide `as WebhookEvent` which is a really useful union. we'd just have to trust that they keep their end of the deal by not deviating from that type. It seems unlikely to me

We could validate just the fields we need with zod? Though I think if any were missing it would already throw an error

codeincontext: Where do you think would be a good place to put in the schema validation without losing that more detailed type information? Maybe in `syncUserToPosthog`?

apps/nextjs/src/app/api/webhooks/clerk/route.ts
codeincontext: I've got an idea. Why don't we set the webhook ID as a breadcrumb and then we can look up the exact shape and content

apps/nextjs/src/app/api/webhooks/clerk/route.ts
codeincontext: I'm going to stick with it as-is, and if we find it needs more safety we can add it later. At the moment we're not using this for anything mission critical so it should be OK

PR #332, opened 2024-11-01
fix: dropdown sections redux (WIP)

stefl:
## Description

- Working branch resolving scroll management and scrolling-while-streaming issues
- Experiment with adjusting the context provider to just share what it needs to each component, and seeing what the result is for rendering performance
- Feedback from Adam is we should experiment with [use-context-selector](https://www.npmjs.com/package/use-context-selector) instead of breaking up into multiple providers, which should have the same effect


stefl:
Closing this for now. We're most likely to go with selectively subscribing to what the main chat provider provides

PR #333, opened 2024-11-02
fix: remove global prisma (experiment)

stefl:
## Description

- An attempt to see if we can address the Prisma client connection issues we are having by adjusting how we're instantiating the Prisma client

### DIFF
--- a/packages/api/src/router/chats.ts
+++ b/packages/api/src/router/chats.ts
@@ -1,3 +1,4 @@
+import { prisma } from "@oakai/db/client";
 import * as Sentry from "@sentry/nextjs";
 import { TRPCError } from "@trpc/server";
 import { isTruthy } from "remeda";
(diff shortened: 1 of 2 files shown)


codeincontext (commented):
It looks like this will have a similar behaviour in production (which could fix the problem), but that it will create a new client on hot reloads in development. Is that right?

packages/api/src/router/chats.ts
```diff
@@ -1,3 +1,4 @@
+import { prisma } from "@oakai/db/client";
```
stefl, line 1: @codeincontext if we don't want to change the global import behaviour, we need to add this anyway. This file isn't importing prisma at the moment because of the declare global in db/client.

I'm not sure how this isn't throwing an error.

codeincontext: Yeah, very odd. Maybe we don't actually call these endpoints? I've approved #342

PR #337, opened 2024-11-04
fix: null render in sidebar provider

codeincontext:
I found that these storybook interactions weren't working as the first render was empty. After some digging, I found that that sidebar provider renders the whole page as null on its first render while it's loading from localstorage

This means that all of our pages have been rendering as a blank page, then updating to have content

It also means that in non-dev nextjs builds, the initial server render was wild! Huge icons and other problems from the lack of styling. I've based this PR on #341 to fix that

# Testing

- The sidebar opens and closes as expected
- The general page UI hasn't started flashing or shifting when reloading the page

### DIFF
--- a/apps/nextjs/src/lib/hooks/use-sidebar.tsx
+++ b/apps/nextjs/src/lib/hooks/use-sidebar.tsx
@@ -27,7 +27,7 @@ interface SidebarProviderProps {
 }
 
 export function SidebarProvider({ children }: SidebarProviderProps) {
-  const [isSidebarOpen, setSidebarOpen] = React.useState(true);
+  const [isSidebarOpen, setSidebarOpen] = React.useState(false);
   const [isLoading, setLoading] = React.useState(true);
 
   React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/nextjs/src/lib/hooks/use-sidebar.tsx
```diff
@@ -27,7 +27,7 @@ interface SidebarProviderProps {
 }
 
 export function SidebarProvider({ children }: SidebarProviderProps) {
-  const [isSidebarOpen, setSidebarOpen] = React.useState(true);
+  const [isSidebarOpen, setSidebarOpen] = React.useState(false);
```
mantagen, line 30: Default was open? 🤔

mantagen (approved):

apps/nextjs/src/lib/hooks/use-sidebar.tsx
codeincontext: The default was "the page doesn't render until we've looked at localstorage", which I suppose was falsey

I don't really know why the sidebar state would even need to be in localstorage though?

mantagen: 🤷

mikeritson-oak (approved):
This all looks good during testing. The sidebar is loading as expected and nothing unusual seen during page refreshes/loads. Checked across browser. 👍

PR #340, opened 2024-11-04
test: delete test users with null activeAt

codeincontext:
I noticed in Clerk that we're recycling a small number of test users at the moment. Maybe 15 or so. Looking at the code for the test user pool, it's because some test users have been created but have no lastActiveAt. They then get sorted at the end of the clerk response after the most recently active users

This PR takes a different approach to cleanup:
- Filter staging users by whether the have a test email address
- Sort test users by lastActiveAt OR createdAt
- Delete one of the oldest 8 users (to reduce conflicts between simultaneous runs)

### DIFF
--- a/packages/api/src/router/testSupport/prepareUser.ts
+++ b/packages/api/src/router/testSupport/prepareUser.ts
@@ -1,4 +1,5 @@
 import { clerkClient } from "@clerk/nextjs/server";
+import { isClerkAPIResponseError } from "@clerk/shared";
 import { aiLogger } from "@oakai/logger";
 import { waitUntil } from "@vercel/functions";
 import os from "os";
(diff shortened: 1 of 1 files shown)


packages/api/src/router/testSupport/prepareUser.ts
```diff
@@ -75,24 +75,50 @@ const generateEmailAddress = (personaName: keyof typeof personas) => {
...
+
+  if (userToDelete) {
+    try {
+      await clerkClient.users.deleteUser(userToDelete.id);
+      log.info(
+        "Deleted old test user",
+        userToDelete.primaryEmailAddress?.emailAddress,
+      );
+    } catch (e) {
+      if (e instanceof Error && e.message.includes("Not Found")) {
```
codeincontext, line 113: This is a clerk error instance, but I couldn't find it exported anywhere

mantagen (approved):
Lgtm

PR #341, opened 2024-11-04
fix: set up server side styled-components

codeincontext:
In #337 I noticed that the initial render for our app is always blank. I had a suspicion that this would be hiding some surprises

If you load that page without JS enabled you'll see that the initial render is missing the styled components styling. I've followed the [official guide](https://nextjs.org/docs/app/building-your-application/styling/css-in-js#styled-components) to integration styled components, and checked that it's the same approach the OWA uses


mantagen (approved):
Noice

PR #344, opened 2024-11-06
test: set up chromatic config and get storybook building

codeincontext:
## Description

- Remove unused providers for now
  - This fixes a lookup-on-undefined error related to a "text-primary" key. The error was only present when building storybook rather than opening in dev mode
  - We we add more stories we can evaluate the correct providers to use glovally vs the providers to set on a per-story basis. For example, we will need to mock some providers with setup state
- Add chromatic configuration
- Remove ChatUserAccessCheck. It had a story but we don't actually use it any more


apps/nextjs/package.json
```diff
@@ -23,7 +23,8 @@
     "with-env": "dotenv -e ../../.env --",
     "aila": "tsx scripts/aila-cli.ts",
     "storybook": "dotenv -e ../../.env -- storybook dev -p 6006 --no-open",
-    "build-storybook": "dotenv -e ../../.env -- storybook build"
+    "build-storybook": "dotenv -e ../../.env -- storybook build",
+    "chromatic": "pnpm with-env npx chromatic"
```
stefl, line 27: Is npx safe here? Should it be pnpm dlx?

stefl (approved):
Great stuff! Excited to get this in

apps/nextjs/package.json
codeincontext: Good call. Thank you

PR #345, opened 2024-11-06
feat: move delete all button and restyle side menu

tomwisecodes:
## Description

Moves delete all button and organises menu to match this ticket
https://www.notion.so/oaknationalacademy/Gleap-clear-history-button-464dd50ad44c4bdb97f3857bbf9aa1c8

### DIFF
--- a/apps/nextjs/src/app/page.tsx
+++ b/apps/nextjs/src/app/page.tsx
@@ -4,6 +4,6 @@ import HomePage from "./home-page";
 
 export default async function Page() {
   const result = await fetchAiHomepage();
-
+  // test comment
   return <HomePage pageData={result} />;
 }


apps/nextjs/src/components/AppComponents/Chat/clear-history.tsx
```diff
@@ -1,81 +1,51 @@
...
+  }
   return (
-    <AlertDialog open={open} onOpenChange={setOpen}>
-      <AlertDialogTrigger asChild>
-        <Button
-          variant="ghost"
-          disabled={!isEnabled || isPending}
-          className="fixed bottom-0 right-0 z-40 flex w-full justify-center bg-white p-16 pl-0 sm:w-[298px]"
+    <OakBoxWithCustomWidth
+      width={447}
```
codeincontext, line 23: Is it possible that this could overflow on mobile?

apps/nextjs/src/components/AppComponents/Chat/clear-history.tsx
```diff
@@ -1,81 +1,51 @@
...
+      $position="fixed"
```
codeincontext, line 24: I can't see the designs, but if this is within a container like a sidebar, should it be position:relative?

apps/nextjs/src/components/AppComponents/Chat/sidebar-item.tsx
```diff
@@ -1,53 +1,51 @@
...
-        },
-      }}
-      transition={{
-        duration: 0.25,
-        ease: "easeIn",
-      }}
+    <OakFlex
+      $justifyContent="space-between"
+      $alignItems="center"
+      $background={isHovered ? `grey30` : `white`}
```
codeincontext, line 39: Surely Oak components provide a way to do this with CSS? Coudl we use styled-components if not?

apps/nextjs/src/components/DialogControl/ContentOptions/ClearSingleChatFromChatHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const ClearSingleChatFromChatHistory = ({
+  closeDialog,
+}: {
+  closeDialog: () => void;
+}) => {
+  const { mutate, isLoading } =
+    trpc.chat.appSessions.deleteChatById.useMutation({
+      onSuccess() {
+        console.log("Chat history cleared");
```
codeincontext, line 25: This should be giving an eslint warning. Have you tried using an instance of `aiLogger`?

apps/nextjs/src/components/DialogControl/ContentOptions/ClearSingleChatFromChatHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+        closeDialog();
+        localStorage.removeItem("chatIdToDelete");
+      },
+      onError() {
+        console.error("Failed to clear chat history");
+      },
+    });
+
+  function deleteChatById() {
+    const id = localStorage.getItem("chatIdToDelete");
```
codeincontext, line 35: This really shows me that the way we're doing dialogs isn't appropriate at all. Maybe you could set it on the dialog context?

We don't really need localstorage for this as it doesn't need to persist. You could equally set a value on `window` on on an object exported from another file

If we think we'll tidy this up later then we could go ahead with this knowing that it's not the right way to be doing it

apps/nextjs/src/components/AppComponents/Chat/sidebar-items.tsx
```diff
@@ -12,21 +14,65 @@ interface SidebarItemsProps {
...
+      const monthYear = format(chat.updatedAt, "MMMM yyyy");
+      if (!acc[monthYear]) {
+        acc[monthYear] = [];
+      }
+      acc[monthYear]?.push(chat);
+    }
+    return acc;
+  }, {});
+
+  const renderChats = (title: string, chats: SideBarChatItem[]) => (
```
stefl, line 49: Could this be a separate component rather than declaring it inline?

apps/nextjs/src/components/AppComponents/Layout/index.tsx
```diff
@@ -15,15 +15,17 @@ const Layout = ({
...
-    <div className="flex min-h-screen flex-col ">
-      <Header />
-      <main
-        className={`flex h-full flex-1 flex-col bg-muted/50 bg-white ${isDemoUser && "pt-28 sm:pt-20"}`}
-      >
-        <DialogProvider>{children}</DialogProvider>
-      </main>
-      {includeFooter && <Footer />}
-    </div>
+    <DialogProvider>
```
stefl, line 18: Any danger that we end up with nested DialogProviders? I thought this was in the layout.tsx file

apps/nextjs/src/components/BinIcon.tsx
```diff
@@ -0,0 +1,18 @@
...
+        fill="#222222"
```
stefl, line 12: If you use fill currentColor then it will use the current text colour and make this more reusable

apps/nextjs/src/components/DialogControl/ContentOptions/ClearChatHistory.tsx
```diff
@@ -0,0 +1,67 @@
...
+import LoadingWheel from "@/components/LoadingWheel";
+import { trpc } from "@/utils/trpc";
+
+const ClearChatHistory = ({ closeDialog }: { closeDialog: () => void }) => {
+  const router = useRouter();
+  const clearChats = trpc.chat.appSessions.deleteAllChats.useMutation({
+    onSuccess() {
+      toast.success("Chat history cleared");
+      closeDialog();
+      router.push("/");
```
stefl, line 23: Should this push to "/aila"?

apps/nextjs/src/components/DialogControl/ContentOptions/ClearSingleChatFromChatHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+  closeDialog,
+}: {
+  closeDialog: () => void;
+}) => {
+  const { mutate, isLoading } =
+    trpc.chat.appSessions.deleteChatById.useMutation({
+      onSuccess() {
+        console.log("Chat history cleared");
+        closeDialog();
+        localStorage.removeItem("chatIdToDelete");
```
stefl, line 27: We've seen localStorage failures on Oak when it's not available on some teacher computers. Maybe try / catch this

apps/nextjs/src/components/DialogControl/ContentOptions/ClearSingleChatFromChatHistory.tsx
stefl: Localstorage is sometimes not available on some school computers so this might throw and make it so you can't clear the history

packages/api/src/router/appSessions.ts
```diff
@@ -199,6 +199,7 @@ export const appSessionsRouter = router({
     const sessions = await ctx.prisma.$queryRaw`
       SELECT
         id,
+        "updated_at" as "updatedAt",
```
stefl, line 202: Strange this isn't mapped on the prisma schema like this!

stefl (approved):
A couple of issues with local storage we might want to look at

apps/nextjs/src/components/AppComponents/Chat/sidebar-item.tsx
tomwisecodes: This is a bit of a hack sure its to match the current functionality which had a slightly weird hover

apps/nextjs/src/components/DialogControl/ContentOptions/ClearSingleChatFromChatHistory.tsx
tomwisecodes: Yeah we have ran into an issue here. 
Traditionally, with this set up we can pass props to a dialog directly however due to the knarly css here the dialog nested at this point does not work at all. I can have another look now im zoomed out a bit and see if I can refactor the css to allow for the dialog to sit at this level. Alt a some context could work too

mikeritson-oak (changes requested):
Issues found in the preview branch and added to the Notion ticket under 'Testing notes'.

PR #348, opened 2024-11-07
fix: only categorise initial user input once

stefl:
## Description

Back-end only, no user-facing change other than speeding up each message by a few seconds in the initial response time.

- We currently have a bug where we categorise the lesson plan each time we send a message
- This is because we were failing to call the aila.initialise() method consistently
- This add a private boolean flag to Aila to store that it has been initialised
- generate and generateAsync will throw if the instance has not been initialised
- Updates tests to call the initialise method
- Update the Aila test to ensure we only categorise empty lesson plans once
- We have this separate initialise method because it is async and cannot be called in the constructor (unless I am mistaken!)

## How to test
* Run the dev server
* Start a new lesson
* See a log for aila:categorisation with the initial message as input
* Send a another message
* No further aila:categorisation logs appear

### DIFF
--- a/apps/nextjs/src/app/api/chat/config.ts
+++ b/apps/nextjs/src/app/api/chat/config.ts
@@ -17,7 +17,7 @@ export const defaultConfig: Config = {
   prisma: globalPrisma,
   createAila: async (options) => {
     const webActionsPlugin = createWebActionsPlugin(globalPrisma);
-    return new Aila({
+    const createdAila = new Aila({
       ...options,
       plugins: [...(options.plugins || []), webActionsPlugin],
       prisma: options.prisma ?? globalPrisma,
@@ -26,5 +26,7 @@ export const defaultConfig: Config = {
         userId: undefined,
       },
     });
+    await createdAila.initialise();
+    return createdAila;
   },
 };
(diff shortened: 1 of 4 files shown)


codeincontext:
> * We have this separate initialise method because it is async and cannot be called in the constructor (unless I am mistaken!)

That's an interesting distinction between a class and using a "regular" closure I hadn't considered

codeincontext (approved):
Great PR. I tested on the preview branch and it worked as expected

I suppose we would need to add an agent if the user wanted to change the key stage or subject later

stefl:
> Great PR. I tested on the preview branch and it worked as expected
> 
> I suppose we would need to add an agent if the user wanted to change the key stage or subject later

I think Aila might try to do its best if you try to change the entire title / subject / key stage but really the best approach there is to start a new lesson

codeincontext:
> > Great PR. I tested on the preview branch and it worked as expected
> > I suppose we would need to add an agent if the user wanted to change the key stage or subject later
> 
> I think Aila might try to do its best if you try to change the entire title / subject / key stage but really the best approach there is to start a new lesson

I think as we change the UI around picking key stage that problem might go away anyway

PR #350, opened 2024-11-07
fix: handle aborts and add logging to the stream handler

stefl:
## Description

Back-end only

- While working on the streaming handling to send fixtures to the front end during tests, it was useful to see more logs relating to streaming
- This adds logging for each step (in the absence of tracing)
- The streaming logic was maintaining its own isStreaming boolean which would get out of sync with the actual state of the controller
- This removes that logic and uses the controller as the source of truth
- The impact is that if you close the connection Aila should gracefully handle the controller going away


codeincontext (approved):
Nice

PR #353, opened 2024-11-07
feat: add FeatureFlagProvider

codeincontext:
## Description

In #320 we added server-side generation of feature flags. That populated the initial flags bootstrapped on posthog.init, and we could then query posthog to get the feature flag status

It turns out that this has two problems:
- Posthog is initialised in a useEffect, so the server render and initial client render don't have information about flags
- useAnalyticsService only initialises posthog when cookie consent is granted. If the user denies consent or never interacts with the cookie banner, posthog won't work as expected

This PR:

- Creates FeatureFlagProvider
- Passes bootstrapped server feature flags to FeatureFlagProvider
- Makes useClientSideFeatureFlag consume the provider, and fall back to a bootstrapped value if posthog hasn't loaded flags

Testing:
-  We don't have any flags at the moment, but you can play with this on #354


JBR90 (approved):
LGTM, tested the 'cool new feature' and worked as expected

codeincontext:
> LGTM, tested the 'cool new feature' and worked as expected

Great. Thanks Joe

PR #365, opened 2024-11-14
test: add storybook stories for public pages

codeincontext:
Adds storybook stories for each of our public pages

In the long run we might want more targeted stories for the atoms and molecules which make up pages, but for now this is a good way to get test coverage

For each page I've created an inner `MyPageContent` component without the `<Layout>` wrapper. We want to test the layout independently. This has caused large diffs due to the indentation change, but there actually isn't that much being changed


apps/nextjs/src/app/faqs/index.tsx
```diff
@@ -36,821 +36,816 @@ const FAQPage = () => {
     }
   };
   return (
-    <Layout>
```
codeincontext, line 39: The only change here is removing this `<Layout>`

apps/nextjs/src/cms/types/policyDocument.ts
```diff
@@ -1,7 +1,7 @@
 export interface PolicyDocument {
   title: string;
   slug: string;
-  fake_updatedAt: string;
+  fake_updatedAt: string | null;
```
codeincontext, line 4: The fixture data was failing the type

JBR90 (approved):
Looks good!

mikeritson-oak (approved):
This looks good. Checked in Chromatic and approved and, as this is a new process, also checked that all of the affected pages still render in the preview branch. 👍

PR #367, opened 2024-11-14
chore: copy updates for help page - AI-492

JBR90:
## Description

- Adds the copy changes from this [ticket](https://www.notion.so/oaknationalacademy/Updates-to-help-page-5aad6523583549148dd7bdfbcc8edfa0)

## How to test

https://oak-ai-lesson-assistant-9bde9x93r.vercel-preview.thenational.academy/aila/help

## Checklist


codeincontext (approved):

apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -204,8 +207,9 @@ const Help = () => {
               When you have completed all the sections in the lesson, you will
               be able to download the lesson resources, including the full
               lesson plan (which includes all sections), editable slides and
-              starter and exit quizzes. You can also share your lesson as a
-              link, which can be viewed online.
+              starter and exit quizzes, a worksheet with your practice tasks and
```
mikeritson-oak, line 210: I feel like this new paragraph is a bit clunky and has at least one too many 'and' in there, possibly this one:
editable slides and starter and exit quizzes, as we then use another one at the end of the sentence. May just need confirmation of how we want the actual section to read. Seems like it could be an intro sentence, list and then sharing outro sentence, but it also kind of replicates the list that is already up above. @hannah-bethclark

apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -187,7 +190,7 @@ const Help = () => {
               each section, for example:
             </p>
             <ul className="mb-6 list-disc pl-12">
-              <li>adapt the content for a specific reading age.</li>
+              <li>increase or decrease the literacy level of your keywords</li>
```
mikeritson-oak, line 193: This addition doesn't have a full stop at the end of it whilst all other items in this list do. I'm sure the general rule is that items in lists should not have full stops, so it might be that the other items are wrong, not this one.

mikeritson-oak (changes requested):

mikeritson-oak (changes requested):
In the final section `AI and accuracy` I don't think Retrieval augmented generation should be capitalised, if we want to adhere to the sentence case policy.
(I couldn't find how to comment on this inline with the code.)

JBR90:
> Two general notes from testing:
> 
> 1. I noticed that all of the headings we now have are not reflected in the menu structure on the left, is this intentional? I'd assume they should all be present.
> 2. Clicking on the links in this list of section doesn't take the viewport to the selected section. There seems to be some movement but only to the top of the content.
> 
> Double checked on production and both are these are happening there, so not necessarily due to these changes but in this area.

Have address the comments, fixed the nav and added missing sections and refactored everything to Oak Components. Will await @hannah-bethclark feedback for any copy changes

apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -30,196 +44,247 @@ export const HelpContent = () => {
...
-                  onClick={() => scrollToRefWithOffset(downloadsRef)}
-                >
+                </OakHeading>
+                <OakP $mb={"space-between-xs"}>
+                  Lessons are created using Oak&apos;s curriculum principles and
+                  include the following sections:
+                </OakP>
+                <OakUL className="mb-6 list-disc pl-12">
+                  <OakLI>Learning outcome</OakLI>
+                  <OakLI>Learning cycle outcomes</OakLI>
```
stefl, line 166: This seems different to the language in the app

apps/nextjs/src/app/aila/help/index.tsx
```diff
@@ -30,196 +44,247 @@ export const HelpContent = () => {
...
-                </button>
-              </li>
-              <li>
-                <button
-                  className="mb-14 text-blue underline"
-                  onClick={() => scrollToRefWithOffset(aiRef)}
-                >
+                </OakHeading>
+                <OakP $mb={"space-between-xs"}>
+                  Once you’ve completed all the sections in the lesson, you’ll
```
stefl, line 261: Should this be &apos; instead of a '?

stefl (approved):

apps/nextjs/src/app/aila/help/index.tsx
codeincontext: I think it's a `’` rather than `'`, which is correct (and would be an &apos; symbol in raw html)

codeincontext (changes requested):
Bringing the layout into the component has added the layout to the storybook story, which we don't normally do. If you look at other pages you'll see a pattern to rename this component to `HelpContent` and wrap it with a component which includes the layout

apps/nextjs/src/app/aila/help/index.tsx
JBR90: Yeah good spot, have adjusted to match current lesson sections

mikeritson-oak (approved):
Changes have all been made, we have one question around copy outstanding that will be raised in a new ticket. 👍

PR #372, opened 2024-11-19
test: stories for dialogs

codeincontext:
Adds stories for each of our dialogs. Each dialog has a component, but we interact with them through the DialogContents component. I figured it made sense to test through that interface

### Mocking providers

Some of the dialogs need some mocked data from posthog, the demo status, or the demo API call. For the time being I've mocked each of these at the highest point possible, to help with coverage. eg: Mock the tRPC call from the DemoProvider, rather than mocking the DemoProvider itself. These patterns might change in the future as we learn more about storybook and have more tests in place

### MSW (Mock Service Worker)

This is the recommended way to mock network requests in storybook. It seems to work really nicely out of the box

### Snapshots

You can see the story changes [in chromatic](https://www.chromatic.com/review?appId=672908473f629875f0be294f&number=372&type=linked&view=changes)

### DIFF
--- a/apps/nextjs/.storybook/public/mockServiceWorker.js
+++ b/apps/nextjs/.storybook/public/mockServiceWorker.js
@@ -1,0 +1,295 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker.
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
+ */
+
+const PACKAGE_VERSION = '2.6.5'
+const INTEGRITY_CHECKSUM = 'ca7800994cc8bfb5eb961e037c877074'
+const IS_MOCKED_RESPONSE = Symbol('isMockedResponse')
+const activeClientIds = new Set()
+
+self.addEventListener('install', function () {
+  self.skipWaiting()
+})
+
+self.addEventListener('activate', function (event) {
+  event.waitUntil(self.clients.claim())
+})
+
+self.addEventListener('message', async function (event) {
+  const clientId = event.source.id
+
+  if (!clientId || !self.clients) {
+    return
+  }
+
+  const client = await self.clients.get(clientId)
+
+  if (!client) {
+    return
+  }
+
+  const allClients = await self.clients.matchAll({
+    type: 'window',
+  })
+
+  switch (event.data) {
+    case 'KEEPALIVE_REQUEST': {
+      sendToClient(client, {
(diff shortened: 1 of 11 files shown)


apps/nextjs/src/components/DialogControl/DialogContents.tsx
```diff
@@ -35,7 +35,7 @@ const dialogTitlesAndIcons: Record<
     iconName: "warning",
   },
   "demo-interstitial": {
-    title: "Lesson limit reached",
+    title: "Demo lesson limits",
```
codeincontext, line 38: This is a copy change, as the previous title was inccorect a lot of the time

apps/nextjs/src/components/DialogControl/DialogContents.tsx
```diff
@@ -67,7 +67,13 @@ const DialogContents = ({
...
         title={dialogTitlesAndIcons[dialogWindow].title}
         iconName={dialogTitlesAndIcons[dialogWindow].iconName}
-        hideIcon={dialogWindow === "feedback" || dialogWindow === "share-chat"}
+        iconOverride={
+          dialogWindow === "feedback" || dialogWindow === "share-chat"
+            ? {
+                $display: "none",
+              }
+            : {}
+        }
```
codeincontext, line 76: `hideIcon` doesn't seem to exist. This seems to do the job

apps/nextjs/.storybook/public/mockServiceWorker.js
```diff
@@ -0,0 +1,295 @@
+/* eslint-disable */
```
codeincontext, line 1: This is autogenerated

apps/nextjs/src/components/DialogControl/DialogContents.tsx
codeincontext: @mikeritson-oak it might be worth checking this

JBR90 (approved):
LGTM

apps/nextjs/src/components/DialogControl/DialogContents.tsx
mantagen: You might need to run `npm install` (to get the later version of oak components)

codeincontext: oh! I'll have a look

codeincontext: Correct! Brownie points for you. Thanks

PR #373, opened 2024-11-20
refactor: add more granular logging to cron job

JBR90:
## Description

Google drive delete cron job is throwing errors from Prisma, 

-  Update each record 1 at a time to identify [error](https://oak-national-academy.sentry.io/issues/11846058/?alert_rule_id=2299&alert_type=issue&environment=production&notification_uuid=accb24d2-7777-4df8-a3f3-f031453a52a0&project=4507089597300816&referrer=slack) with specific id's
- Allow job to continue to delete files when an id is not found in the database and report missing ids at the end with error

How to test: (comment out auth)
http://localhost:2525/api/cron-jobs/expired-exports

https://www.notion.so/oaknationalacademy/Google-Drive-file-deletion-cron-job-14426cc4e1b180b1bf88dd8d2c566ad1

## Checklist

### DIFF
--- a/apps/nextjs/src/app/api/cron-jobs/expired-exports/route.ts
+++ b/apps/nextjs/src/app/api/cron-jobs/expired-exports/route.ts
@@ -15,47 +15,53 @@ requiredEnvVars.forEach((envVar) => {
   }
 });
 
-async function updateExpiredAt(fileIds: string[]) {
+async function updateExpiredAtAndDelete(fileIds: string[]) {
   if (fileIds.length === 0) {
     log.info("No file IDs to update.");
     return;
   }
 
-  try {
-    const result = await prisma.lessonExport.updateMany({
-      where: {
-        gdriveFileId: {
-          in: fileIds,
-        },
-      },
-      data: {
-        expiredAt: new Date(),
-      },
-    });
-    log.info(`Updated expiredAt for ${fileIds.length} files.`);
+  const failedIds: string[] = [];
 
-    if (result.count === fileIds.length) {
-      log.info("All files updated successfully.");
-    } else {
-      throw new Error(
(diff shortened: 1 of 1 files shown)


apps/nextjs/src/app/api/cron-jobs/expired-exports/route.ts
```diff
@@ -15,47 +15,53 @@ requiredEnvVars.forEach((envVar) => {
...
-    }
-  } catch (error) {
-    log.error("Error updating expiredAt field in the database:", error);
-    throw error;
-  }
-}
+  for (const id of fileIds) {
+    try {
+      const record = await prisma.lessonExport.findFirst({
+        where: { gdriveFileId: id },
```
JBR90, line 29: Not unique value in database so find first to get lesson id

mantagen: Would we not want to get all rather than just the first one?

mantagen: > I could findAll and throw error if it finds more than one
Could we not findAll and updateMany?
Other option would be to add a unique constraint on that field

JBR90: The trouble with update many is that it only returns a count of the updated ids and not any info on failed ids

mantagen: Sure, fair enough. Let's put a unique constraint on it then, as we're not expecting multiple records per file! I'll put that in a separate PR

JBR90:
> Do you have an idea of how much longer this approach will take to run? Thinking the function could time out -- but happy to trial it

Lets give it a go, i will at least give us some more info from the logs on ID's causing problem as if loops round.

PR #377, opened 2024-11-21
test: stories for right hand side of lesson planner

codeincontext:
Stories for:

- LessonPlanDisplay: The right hand side view that pulls it all together
- ExportButtons
- MobileExportButtons: These are extracted from LessonPlanDisplay
- DropDownSection: Including variants for different states, content, and for the modify dropdown

These components use the ChatContext, so I've specified a mock context in each file. If we find we're doing this a lot and that there are some consistent use cases we can find a reusable pattern later

### Challenges:
- There's a snapshot to show a collapsed section. That needed a `play` function to click the component. In both the "docs" page and chromatic I had to add a 50ms timeout before clicking
- The mobile export buttons only render on mobile, so I had to specify a mobile viewport

You can view the changes [in chromatic](https://www.chromatic.com/review?appId=672908473f629875f0be294f&number=377&type=linked&view=changes)

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.stories.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.stories.tsx
@@ -8,6 +8,8 @@ import DropDownSection from "./";
 
 const MAX_INT32 = 2 ** 31 - 1;
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 const ChatDecorator: Story["decorators"] = (Story, { parameters }) => (
   <ChatContext.Provider
     value={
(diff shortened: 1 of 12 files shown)


apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+  </ChatContext.Provider>
+);
+
+const meta: Meta<typeof DropDownSection> = {
+  title: "Components/LessonPlan/DropDownSection",
+  component: DropDownSection,
+  tags: ["autodocs"],
+  args: {
+    objectKey: "learningOutcome",
+    // objectKey,
```
mantagen, line 44: Can remove commented keys in this file?

mantagen (commented):
Looking good, though haven’t run it locally as I’m out and about. Left a comment about comments!

apps/nextjs/src/components/AppComponents/Chat/drop-down-section/index.stories.tsx
codeincontext: Good spot. Thanks

mikeritson-oak (approved):
Looks good in preview and Chromatic. 👍

PR #378, opened 2024-11-22
refactor: cron job gd exports error after update

JBR90:
## Description

This is for the cron job that runs overnight and deletes google drive files that have expired (older than 2 weeks)

- Moves error outside of loop, reports error with all file ids that have failed in one error.

## Checklist


apps/nextjs/src/app/api/cron-jobs/expired-exports/route.ts
```diff
@@ -54,14 +54,13 @@ async function updateExpiredAtAndDelete(fileIds: string[]) {
...
-      log.error(errorMessage);
-      throw new Error(errorMessage);
-    }
+  }
+  if (failedIds.length > 0) {
+    const errorMessage = `Failed to process the following file IDs: ${failedIds.join(
+      ", ",
+    )}`;
+    log.error(errorMessage);
+    throw new Error(errorMessage);
```
codeincontext, line 63: Could this potentially be 1000 IDs? I don't know if there's a limit on an error message length, but I'm happy to go with it and we'll see if it needs another tweak or not

codeincontext (approved):

packages/api/src/router/testSupport/prepareUser.ts
```diff
@@ -95,7 +95,7 @@ const generateEmailAddress = (personaName: keyof typeof personas) => {
     // All users use the "test@thenational.academy" mailbox with a `+` alias
     "test",
     // Replace non-alphanumeric characters with -
-    branch.replace(/\W+/g, "-"),
+    branch.replace(/\W+/g, "-").slice(-15),
```
JBR90, line 98: @codeincontext - bit of a crude solution here. Will this cause uniqueness issues?

codeincontext: Let's go for it and see what happens. I think it'll be fine

codeincontext (approved):

PR #381, opened 2024-11-26
fix: assert readonly props

stefl:
## Description

- Addresses multiple instances of Sonar warnings about not setting React component props as readonly
- Adds an eslint rule to asssert that we mark props as readonly
- Prefers type over interface so that we can us the Readonly helper rather than prefix all type definitions with "readonly" repeatedly
- Where types have changed, ensures the types are exported so that they can be used in tests

[Sonar rule](https://sonarcloud.io/organizations/oaknational/rules?open=typescript%3AS6759&rule_key=typescript%3AS6759)

[43 flagged issues](https://sonarcloud.io/project/issues?impactSoftwareQualities=MAINTAINABILITY&rules=typescript:S6759&issueStatuses=OPEN,CONFIRMED&id=oaknational_ai-experiments)


apps/nextjs/src/components/AppComponents/Chat/markdown.tsx
```diff
@@ -18,15 +18,17 @@ const MemoizedReactMarkdown: FC<Options> = memo(
     prevProps.className === nextProps.className,
 );
 
+export type ReactMarkdownWithStylesProps = Readonly<{
+  markdown: string;
+  lessonPlanSectionDescription?: string;
+  className?: string;
+}>;
```
codeincontext, line 25: It's nice to extract this to a named type outside the function. Make sense

apps/nextjs/src/components/AppComponents/QuizDesigner/SuggestedQuestionCard.tsx
```diff
@@ -8,21 +8,23 @@ import type { PotentialQuestionsType } from "hooks/useSuggestedQuestions";
...
 import { Icon } from "@/components/Icon";
 
+export type SuggestedLessonCardProps = Readonly<{
+  answer: PotentialQuestionsType[0];
+  dispatch: Dispatch<QuizAppAction>;
+  questionsWrapperRef: React.RefObject<HTMLDivElement>;
+  potentialNewQuestions: PotentialQuestionsType;
+  setPotentialNewQuestions: React.Dispatch<PotentialQuestionsType>;
+  questionRefs: React.MutableRefObject<(HTMLDivElement | null)[]>;
+}>;
```
codeincontext, line 18: 😍

codeincontext (approved):
Nicely done

apps/nextjs/src/components/AppComponents/Chat/markdown.tsx
stefl: If we like this pattern we might want to do a pass on all components to follow the same approach.

codeincontext: I'm personally a fan of it! As part of my "you should be able to scan what a function does without too much scrolling" guideline

PR #383, opened 2024-11-26
feat: soft delete

tomwisecodes:
## Description

- Center text in dialogs for clearing chat 
- Stops user from hard deleting chat and rather adds a flag to the database
- Stops get chat methods from finding soft delted chats

***contains migration*** 

how to test: 
- delteing a chat works as expected
-  the text in the dialog is centered on mobile


mantagen (approved):
Code looks good

mikeritson-oak (approved):
Soft delete checked in the frontend and the DB with help from @mantagen. Looks good. 👍

PR #384, opened 2024-11-26
fix: minor linting

stefl:
## Description

- This addresses some of the items marked as Minor in our Sonar Cloud report
- The idea is to reduce down some of the very small issues so that we can more easily triage the intermediate and major items that have been flagged

Things included:
* Typos
* Options chains
* Nullish coalescing
* Readonly attributes on classes
* Combining type imports
* Empty fragments
* Unnecessary return


apps/nextjs/.storybook/public/mockServiceWorker.js
```diff
@@ -8,133 +8,133 @@
  * - Please do NOT serve this file on production.
  */
 
-const PACKAGE_VERSION = '2.6.5'
-const INTEGRITY_CHECKSUM = 'ca7800994cc8bfb5eb961e037c877074'
-const IS_MOCKED_RESPONSE = Symbol('isMockedResponse')
-const activeClientIds = new Set()
+const PACKAGE_VERSION = "2.6.5";
```
stefl, line 11: These quote changes seem to be automated when saving the file

codeincontext: This file is autogenerated so I assume that we shouldn't be modifying or linting it?

codeincontext (approved):
Approved, though I'm not sure we care too much about the MSW file

apps/nextjs/.storybook/public/mockServiceWorker.js
stefl: Thanks - I've removed that an added it to ignore config

PR #392, opened 2024-11-27
fix: add linting command to db package

stefl:
## Description

- Adds pnpm lint and type-check commands to the db package which were missing
- Fixes some small linting isses in the db package
- Marks more complex issues that require a refactor as ignore for now
- Unfortunately this means our sonar error count will go up most probably :-(
- Adjusts the CI process for linting to use no-engine and run it as a separate step


codeincontext (commented):
Looks good to me. I like the idea of ignoring older lint errors rather then getting sucked into fixing them all immediately

PR #394, opened 2024-11-27
fix: sonar maintain linting #1

stefl:
## Description

- Addresses some sonar issues marked as "maintain"
- Promisifies some non-async methods
- Adds readonly properties to various classes that are not modifying them

### DIFF
--- a/packages/aila/src/features/generation/AilaGeneration.ts
+++ b/packages/aila/src/features/generation/AilaGeneration.ts
@@ -16,18 +16,18 @@ import type { AilaGenerationStatus } from "./types";
 const log = aiLogger("generation");
 
 export class AilaGeneration {
-  private _aila: AilaServices;
+  private readonly _aila: AilaServices;
   private _id: string;
-  private _chat: AilaChat;
+  private readonly _chat: AilaChat;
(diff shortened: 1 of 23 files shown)


stefl:
@tomwisecodes could you check the tabindex change is okay? Sonar recommends only using -1 and 0. I'm not sure why this is appearing on both these PRs

packages/aila/src/features/generation/AilaGeneration.ts
```diff
@@ -139,16 +139,20 @@ export class AilaGeneration {
   }
 
   private async calculateTokenUsage(): Promise<void> {
```
codeincontext, line 141: If we remove the `async` declaration here then I imagine we don't need to wrap in a promise?

stefl: I didn't want to change the behaviour itself. I think because we originally had another method for calculating tokens and it might be that we need to keep this async

codeincontext: Got it. Looking at the interface I didn't see any other implementations that would be async, but I didn't think of that

codeincontext (approved):

PR #395, opened 2024-11-27
fix: sonar maintain linting #2

stefl:
## Description

- Fixes some Sonar issues marked as "maintain"


stefl:
@tomwisecodes could you check the tabindex change is okay? Sonar recommends only using -1 and 0

JBR90 (approved):

PR #399, opened 2024-11-27
fix: help page cloudflare email

codeincontext:
The help page renders a help email in the server render. Cloudflare obfuscates this and provides some JS to render it with JS. Our CSP is blocking that script

This PR emulates what cloudflare is doing


mantagen (approved):
LGTM

PR #409, opened 2024-11-28
fix: do not use array index for key / use void for onSubmit

stefl:
## Description

- Fixes a small issue with using an array for the key of a repeated set of items
- Fixes a missing "void"s in the same files

### DIFF
--- a/apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
+++ b/apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
@@ -7,7 +7,11 @@ import { OakAccordion, OakPrimaryButton } from "@oaknational/oak-components";
 
 import { trpc } from "@/utils/trpc";
 
-function ModerationListItem({ moderation }: { readonly moderation: Moderation }) {
+function ModerationListItem({
+  moderation,
+}: {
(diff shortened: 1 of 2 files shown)


apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
```diff
@@ -41,14 +45,16 @@ function ModerationListItem({ moderation }: { readonly moderation: Moderation })
...
-                className="inline-block rounded-md bg-zinc-300 px-8 py-4 text-xs font-semibold text-zinc-800"
-              >
-                {String(category)}
-              </span>
-            ))}
+            {Array.from(new Set(moderation.categories))
+              .map((c) => String(c))
+              .map((category) => (
+                <span
+                  key={category}
```
Comment (JBR90) [inline apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:52]: Is this always unique?

JBR90 (approved):

apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
Comment (stefl) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:52]: Yes, we have a set of categories that the moderation can flag

PR #411, opened 2024-11-28
fix: more flexible and efficient solution for images in docs

mantagen:
## Description

- Changes how we replace markdown images with actual images in Google Docs
  1. Handle text replacement (as we did before)
  2. Find markdown images (and their indexes in the document)
  3. Create and submit a list of requests to remove the markdown image reference and insert images at the appropriate index. 

- Slides with images is behind a feature flag so this won't have any effect unless you're in the feature flag group, in which case you should see the quiz exports have 6 questions and all images replaced if you generate a maths lesson.
- To enable this feature for your user you can set `featureFlagGroup=maths-quiz` in your user's public Clerk metadata


packages/aila/src/utils/experimentalPatches/mathsQuiz.fixture.ts
```diff
@@ -44,4 +44,15 @@ export const mathsQuizFixture: Quiz = [
       "neither similar nor congruent.",
     ],
   },
+  {
+    question:
+      "Here is an image ![image](http://oaknationalacademy-res.cloudinary.com/image/upload/v1706110974/fukcqeavzcevgjhmm1n4.png) and here is the same image  ![image](http://oaknationalacademy-res.cloudinary.com/image/upload/v1706110974/fukcqeavzcevgjhmm1n4.png)",
```
mantagen, line 49: This ensure there are 6 questions in the ficutre, and a question that has `{image} {text} {image}`, which was previously failing to replace

packages/exports/src/gSuite/docs/populateDoc.ts
```diff
@@ -48,13 +49,26 @@ export async function populateDoc<
       });
     }
 
+    const markdownImages = await findMarkdownImages(googleDocs, documentId);
```
mantagen, line 49: This is the high level change

codeincontext (approved):

PR #413, opened 2024-11-28
fix: do not define components inline

stefl:
## Description

- Fixes a sonar complaint about defining components inline


codeincontext (approved):
Nice

PR #418, opened 2024-12-02
fix: moderation persist scores [AI-696]

mantagen:
## Description

- store the scores we get back from the moderation agent
- **Note: contains migration**

### DIFF
--- a/packages/aila/src/features/moderation/moderators/OpenAiModerator.ts
+++ b/packages/aila/src/features/moderation/moderators/OpenAiModerator.ts
@@ -123,10 +123,7 @@ export class OpenAiModerator extends AilaModerator {
     );
 
     const log = aiLogger("aila:moderation:response");
-    log.info(
-      "Moderation response: ",
-      JSON.stringify(moderationResponse, null, 2),
-    );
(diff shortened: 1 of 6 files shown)


packages/aila/src/features/moderation/moderators/OpenAiModerator.ts
```diff
@@ -123,10 +123,7 @@ export class OpenAiModerator extends AilaModerator {
     );
 
     const log = aiLogger("aila:moderation:response");
-    log.info(
-      "Moderation response: ",
-      JSON.stringify(moderationResponse, null, 2),
-    );
+    log.info(JSON.stringify(moderationResponse));
```
mantagen, line 126: @mfdowland we've been logging the full response ... so i';; see if i can get old scores exported!

tomwisecodes (approved):

packages/aila/src/features/moderation/moderators/OpenAiModerator.ts
mantagen: Actually they were excluded in prod so looks like we won;t be getting these

PR #419, opened 2024-12-02
fix: use TeX for math notation [AI-538]

mantagen:
## Description

- add lines to prompt to request that math is returned using appropriate syntax
- adds https://github.com/fast-reflexes/better-react-mathjax which OWA uses to render it

## Note

This is a quick spike, not intended to be merged. To solve it properly we also need to look at rendering the notation correctly in exports.

It looks [API math notation isn't available](https://issuetracker.google.com/issues/36761273?pli=1) on Google docs etc

[This plugin might be worth looking into](https://tex.stackexchange.com/questions/173317/is-there-a-latex-wrapper-for-use-in-google-docs)

### DIFF
--- a/apps/nextjs/package.json
+++ b/apps/nextjs/package.json
@@ -71,6 +71,7 @@
     "archiver": "^7.0.1",
     "axios": "^1.6.8",
     "base64url": "^3.0.1",
+    "better-react-mathjax": "2.0.4-beta1",
     "class-variance-authority": "^0.7.0",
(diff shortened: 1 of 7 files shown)


mantagen:
Closing, as there's no plan for now to continue this. The task is tracked [here](https://www.notion.so/oaknationalacademy/fix-use-TeX-for-math-notation-AI-538-15026cc4e1b1817389cbec521df5c53b)

PR #420, opened 2024-12-02
test: add story for chat lhs header

codeincontext:
- Add stories
- I've broken out the environment check so that we can test the two versions


mantagen (approved):
nice

mikeritson-oak (approved):
Looks good. 👍

PR #437, opened 2024-12-09
test: add stories for chat header

codeincontext:
## Description

- Add stories for the "chat" header - the header shown when using Aila
- Fix story height for the main header which is also `position:fixed`


mantagen (approved):
LGTM

PR #438, opened 2024-12-10
chore: add dependency cruiser

stefl:
## Description

- Adds [Dependency Cruiser](https://www.npmjs.com/package/dependency-cruiser)
- This allows us to statically analyse the dependencies in the project and find issues like:
  - Circular dependencies
  - Dev dependencies being used when they should be production
  - Orphan code that is no longer used
  
This just adds the package and we can run it locally with `pnpm deps`

It also makes a graph which is fun: `pnpm deps:graph` but a bit hard to understand at the moment!

dependency-graph


JBR90 (approved):
That graph is mad!

PR #440, opened 2024-12-11
fix: posthog consistent event properties

mantagen:
## Description

- ensures consistent key stages are sent with posthog events
- i think this parsing should really happen higher up (as part of the lesson plan creation itself), but this feels a less risky way of testing it out before implementing more deepy 

## How to test

- tricky to test as mostly Aila gets it right
- but locally you could tests it by overriding an Aila response to have key stage "year 11", but the posthog event should have key stage "ks4"


apps/nextjs/src/lib/analytics/helpers.ts
```diff
@@ -71,3 +71,96 @@ export function getModerationTypes(
...
+  let str = "";
+
+  for (const char of strippedMaybeKeyStage) {
+    str += char;
+    const startMatch = startKeyStageMap[str];
+
+    if (startMatch) {
+      return startMatch;
+    }
+  }
```
codeincontext, line 163: I think you could also do this by looping `Object.entries` and comparing with `startsWith`. Could be clearer for people familiar with functional programming. Could be less clear for people who aren't I suppose!

codeincontext (approved):
This makes sense to me and is approvable 👍 

I would prefer to see a self-contained part of the analytics handling like this in its own file like `keyStages.ts` rather than making `helpers.ts` any longer, but it's a nit pick

PR #442, opened 2024-12-11
build: release candidate

codeincontext:
The previous release failed with this error:

```
[11:01:58 AM] [semantic-release] › ✘  An error occurred while running semantic-release: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/oak-ai-lesson-assistant/oak-ai-lesson-assistant/release.config.js from /home/runner/work/oak-ai-lesson-assistant/oak-ai-lesson-assistant/node_modules/cosmiconfig/dist/loaders.js not supported.
release.config.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename release.config.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /home/runner/work/oak-ai-lesson-assistant/oak-ai-lesson-assistant/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

This is a consequence of us missing a js file which needs to be renamed to .cjs


mikeritson-oak (approved):
Rename to .cjs looks OK to me. 👍

PR #443, opened 2024-12-11
test: storybook decorators

codeincontext:
## Description

Storybook and chromatic will only be useful for us if we can build them into a good habit. This means having easy to use helpers for things like context providers. So far, I've been adding custom decorators into each story file as we need them. Now we're ready to make these reusable

A decorator

- Add the following story decorators to .storybook/decorators:
  - SidebarDecorator
  - AnalyticsDecorator
  - ChatDecorator
  - LessonPlanTrackingDecorator
  - DemoDecorator
  - DialogContentDecorator
- Update stories using context to use these decorators
- Specify types for the story parameters supported by each decorator
- DemoContext used to have a union which varied the keys available. This was hard tow ork with in TS, so I've moved the optional keys to a `demo` object in the context

I wasn't able to import from `@/components` in the .storybook folder, so I've used relative imports for now. I've dabbled with the storybook typescript config and I found that the `tsconfig.storybook.json` file in the project isn't being used - so that work should come in a new PR

## How to test

Testing this is actually a bit blocked while we talk to Chromatic about our usage quota

### DIFF
--- a/apps/nextjs/.storybook/decorators/LessonPlanTrackingDecorator.tsx
+++ b/apps/nextjs/.storybook/decorators/LessonPlanTrackingDecorator.tsx
@@ -1,0 +1,20 @@
+import React from "react";
+
+import type { Decorator } from "@storybook/react";
+
+import { lessonPlanTrackingContext } from "../../src/lib/analytics/lessonPlanTrackingContext";
+
+export const LessonPlanTrackingDecorator: Decorator = (Story) => (
+  <lessonPlanTrackingContext.Provider
+    value={{
+      onClickContinue: () => {},
+      onClickRetry: () => {},
+      onClickStartFromExample: () => {},
+      onClickStartFromFreeText: () => {},
+      onStreamFinished: () => {},
+      onSubmitText: () => {},
+    }}
+  >
+    <Story />
+  </lessonPlanTrackingContext.Provider>
+);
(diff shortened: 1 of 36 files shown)


apps/nextjs/.storybook/decorators/LessonPlanTrackingDecorator.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+import type { Decorator } from "@storybook/react";
+
+import { lessonPlanTrackingContext } from "../../src/lib/analytics/lessonPlanTrackingContext";
+
+export const LessonPlanTrackingDecorator: Decorator = (Story) => (
+  <lessonPlanTrackingContext.Provider
```
stefl, line 8: Seems a little strange to have this lowercase

codeincontext: The react docs use a capital for this, so I'll update

mantagen (approved):
Bit annoying that discriminated unions and decorator params don't play nicely, but solution works.
LGTM

codeincontext:
> Bit annoying that discriminated unions and decorator params don't play nicely, but solution works. LGTM

Thinking about it, I do think there are ways to solve that. For example a helper to build a full context in the parameters for each story rather than building up the values in multiple places. I'll buzz you now to work it out

PR #460, opened 2025-01-03
test: add e2e tests for onboarding

codeincontext:
## Description

- Add playwright tests for:
  - User who hasn't been onboarded
  - User who has onboarded before we set demo status metadata
  - User trying to re-onboard
- Add `needs-demo-status` persona
- Add test user preparation to reset metadata if it's changed from the persona values
- Tidy up testSupport folder as there'e now more code to extract out to files

## Issue(s)

Fixes #AI-717

## How to test

- e2e tests work!
- Existing behaviour still works:
  - Onboarding works for new users
  - An onboarded user without `isDemoUser` set will see the "preparing your account" screen


apps/nextjs/src/middlewares/auth.middleware.ts
```diff
@@ -72,11 +72,9 @@ const isPreloadableRoute = createRouteMatcher([
 
 const isOnboardingRoute = createRouteMatcher([
```
mantagen, line 73: What does it mean to be an 'onboarding route' again?

mantagen: .. one that's part of the onboarding flow?

apps/nextjs/src/middlewares/auth.middleware.ts
```diff
@@ -120,6 +118,14 @@ function conditionallyProtectRoute(
       return NextResponse.redirect(new URL("/onboarding", req.url));
     }
   }
+  if (
+    userId &&
+    isOnboardingRoute(req) &&
+    !needsToCompleteOnboarding(sessionClaims)
+  ) {
+    log("Already onboarded: REDIRECT");
```
mantagen, line 126: Intentional to keep the log?

packages/api/src/router/testSupport/personas.ts
```diff
@@ -0,0 +1,97 @@
+const GENERATIONS_PER_24H = parseInt(
```
mantagen, line 1: Is `/testSupport` essentially test helpers?

apps/nextjs/src/middlewares/auth.middleware.ts
mantagen: If so, any idea why we had `/sign-in` and `/api/trpc/main` in the list of routes before?

mantagen (commented):
Code looks good, couple questions! And PR e2e tests timing out likely for unrelated reasons

packages/api/src/router/testSupport/personas.ts
codeincontext: Yep exactly. They're routes which are used by the e2e tests

apps/nextjs/src/middlewares/auth.middleware.ts
codeincontext: Yep. You'll see that there are other similar logs in this file. They're under `middleware:auth` which is hidden with the default DEBUG env var

apps/nextjs/src/middlewares/auth.middleware.ts
codeincontext: I think it originally meant "routes that are needed for onboarding". Now it's a little more like "routes limited to onboarding"

`/api/trpc/main`: I think I originally copied this from [here](https://github.com/oaknational/oak-ai-lesson-assistant/blob/2f255b3656c4c1c53bafa3b1bed1b795c70a84c6/apps/nextjs/src/middlewares/auth.middleware.ts#L35-L40) in order to enable trpc calls, but it never would have worked as it won't match trpc calls. Hence the more specific endpoint entries in the list

I think that the `/api/trpc`, `/api/trpc/main`, `/api/trpc/chat` declarations in the public route list have never worked, as the url pattern for tRPC is `.api/trpc/main/router.endpoint,router.endpoint2`. This means that trpc calls are authenticated by this middleware rather than the trpc protected procedure logic

If we wanted to use the tRPC prorectected procedures we would need `/api/trpc/main/*` and `/api/trpc/chat/*`

`sign-in`: On main this is only used in `shouldInterceptRouteForOnboarding`, but that's wrapped in `userId && needsToCompleteOnboarding(sessionClaims)`. You can only reach it if you're logged in. My working assumption is that it was needed in an earlier version of this logic which has since been changed/refactored. I haven't been able to recreate a situation where you'd need it but I've added @mikeritson-oak as a reviewer to give it a good poke

mantagen: Great, thanks for this!

mantagen (approved):

mikeritson-oak (approved):
Looks good. Tests are passing and account creation/onboarding are working on the preview environment. 👍

codeincontext:
> Looks good. Tests are passing and account creation/onboarding are working on the preview environment. 👍

Perfect. Thank you

PR #463, opened 2025-01-06
chore: refactor schema.ts to make it more DRY

stefl:
Refactor the schema.ts file to:

* Reduce code duplication
* Break out the multiple Zod describe calls into a single definition object per schema
* Remove unused schema exports
* Organise the file a little better
* Move helper methods into a new file

## Description

- Typescript-only changes, no changes to UX


packages/aila/src/protocol/schema.ts
```diff
@@ -1,79 +1,75 @@
...
-    .max(250)
-    .describe(
-      "No more than 2 sentences addressing the reason for the misconception and how it can be addressed in the lesson.",
-    ),
-  definition: z
-    .string()
-    .optional()
-    .describe("Not to be used. Only included here for legacy support."), // There is some confusion here about what the LLM generates
-});
+// ********** MISCONCEPTIONS **********
```
stefl, line 33: Potentially we put each of these into a separate file?

packages/aila/src/protocol/schema.ts
```diff
@@ -84,29 +80,31 @@ export type MisconceptionsOptional = z.infer<
...
-export const QuizQuestionSchema = z.object({
-  question: z.string().describe("The question to be asked in the quiz."),
-  answers: z.array(z.string()).length(1).describe("The correct answer."),
-  distractors: z.array(z.string()).length(2).describe("A set of distractors."),
-});
+export const QUIZ_DESCRIPTIONS = {
+  question: "The question to be asked in the quiz.",
+  answers: "The correct answer. This should be an array of only one item.",
+  distractors: "A set of distractors. This must be an array with two items.",
+} as const as {
```
stefl, line 89: This typing is somewhat annoying but seems to be necessary unless anyone has a better idea.

packages/aila/src/protocol/schema.ts
```diff
@@ -117,392 +115,312 @@ export const QuizOptionalSchema = z.array(QuizQuestionOptionalSchema);
...
-    "A keyword that is used in the lesson. Written in the TEACHER_TO_PUPIL_SLIDES voice.",
-  );
+  .describe(KEYWORD_DESCRIPTIONS.schema);
 
-export const KeywordOptionalSchema = z.object({
-  keyword: z.string().optional(),
-  definition: z.string().optional(),
-  description: z.string().optional(), // There is some confusion here about what the LLM generates
+export const KeywordSchema = KeywordSchemaWithoutLength.extend({
+  keyword: KeywordSchemaWithoutLength.shape.keyword.max(30),
```
stefl, line 286: To make these more DRY we can use .shape on the parent schema

packages/aila/src/protocol/schema.ts
```diff
@@ -117,392 +115,312 @@ export const QuizOptionalSchema = z.array(QuizQuestionOptionalSchema);
...
+      "key-stage-2",
+      "key-stage-3",
+      "key-stage-4",
+      "key-stage-5",
+      "early-years-foundation-stage",
+      "specialist",
+      "further-education",
+      "higher-education",
+    ]),
+    z.string(),
```
stefl, line 369: I'm unhappy with this, but leaving it unchanged on this PR

packages/aila/src/protocol/schema.ts
```diff
@@ -117,392 +115,312 @@ export const QuizOptionalSchema = z.array(QuizQuestionOptionalSchema);
...
+  .array(z.string())
+  .describe(LESSON_PLAN_DESCRIPTIONS.keyLearningPoints);
+
+export const AdditionalMaterialsSchema = z
+  .string()
+  .optional()
+  .describe(LESSON_PLAN_DESCRIPTIONS.additionalMaterials);
+
+// Main schema
 export const CompletedLessonPlanSchema = z.object({
```
stefl, line 401: The overall lesson plan schema now has per-section schema definitions including descriptions

packages/aila/src/protocol/schema.ts
```diff
@@ -117,392 +115,312 @@ export const QuizOptionalSchema = z.array(QuizQuestionOptionalSchema);
...
+      "early-years-foundation-stage",
+      "specialist",
+      "further-education",
+      "higher-education",
+    ]),
+    z.string(),
+  ])
+  .describe(LESSON_PLAN_DESCRIPTIONS.keyStage);
+
+export const SubjectSchema = z
```
stefl, line 373: These are exported but currently unused. In a subsequent PR I'm planning on using these for edits to these sections, which are currently not present in the main prompt.

packages/aila/src/protocol/schema.ts
```diff
@@ -1,79 +1,75 @@
...
+    It should be the ID of the lesson plan that was used to inform
+    the content of this lesson plan.
+    It should not be the number that the user has selected when
+    choosing which lesson on which to base their new lesson.`,
+  title: "The human-readable title of the lesson.",
+  schema: dedent`A reference to a lesson plan that this lesson is
+    based on. This value should only be set if the user has explicitly
+    chosen to base their lesson on an existing lesson plan by selecting
+    one from a selection of options, otherwise this should be blank.`,
+} as const as { id: string; title: string; schema: string };
```
codeincontext, line 20: I'm not familiar with `as const as`. Maybe you're thinking of `satisfies`?

packages/aila/src/protocol/schema.ts
```diff
@@ -1,79 +1,75 @@
...
 import { zodToJsonSchema } from "zod-to-json-schema";
 
+import { minMaxText } from "./schemaHelpers";
+
+// ********** BASED_ON **********
+export const BASED_ON_DESCRIPTIONS = {
+  id: dedent`The database ID of an existing lesson plan that this
+    lesson plan is based upon.
+    It should be the ID of the lesson plan that was used to inform
+    the content of this lesson plan.
```
codeincontext, line 12: I think that dedent will maintain the line breaks, so there will be a break between "inform" and "the" in the output. Is that intentional?

codeincontext: This applies to other uses of dedent as well

packages/aila/src/protocol/schema.ts
codeincontext: Can you explain what the problem is with the type? As far as I can tell, this is already the type of the literal object in this case

codeincontext (commented):
I think there are good changes on the whole. I'm not sure about:

- Using dedent with wrapped strings. Thesw will also be wrapped in the output. Does that matter?
- The `as const as` types. Just nee da bit of clarity on those

packages/aila/src/protocol/schema.ts
stefl: To be able to use const we have to also add the { question: string; ...} etc. I tried it without or with Record<string, string> but that was giving me complaints.

packages/aila/src/protocol/schema.ts
stefl: I also tried satisfies and it gave me type complaints. Similarly - not familiar with const as!

codeincontext: Does `as { id: string; title: string; schema: string }` work as expected?

packages/aila/src/protocol/schema.ts
stefl: Ah, good catch, thanks!

packages/aila/src/protocol/schema.ts
stefl: This seems to be resolved now!

codeincontext: Lovely

PR #464, opened 2025-01-07
fix: Aria label for buttons, reduce complexity

stefl:
## Description

- The ButtonCore component was emitting `arial-label="[object Object]"`
- It was also failing a code complexity check
- This resolves the aria issue by only providing an aria-label if it is an icon, otherwise the button text would be deduced by the browser
- Refactors the ButtonCore component to break it into sub-components


codeincontext (approved):
Good thinking. This makes sense to me

PR #468, opened 2025-01-07
chore: clean up types and reduce complexity in exports package

stefl:
## Description

- Addresses some small code complexity issues with export methods
- Uses invariant to simplify code where we are expecting process.env variables
- Refactors the findMarkdownImages method to reduce code complexity by breaking out methods
- No user-facing changes


packages/exports/src/gSuite/slides/deleteSlides.ts
```diff
@@ -2,8 +2,9 @@ import type { slides_v1 } from "@googleapis/slides";
 
 import type { Result } from "../../types";
 
-// #TODO Remove this string fallback and properly type the speaker notes tags
-export type SpeakerNotesTag = "cycle1" | "cycle2" | "cycle3" | string;
+export type SpeakerNotesTag =
```
stefl, line 5: I'm not sure why this is called SpeakerNotesTag when it's actually a value indicating a slide that should be deleted

packages/exports/src/exportQuizDesignerSlides.ts
```diff
@@ -21,14 +22,10 @@ export const exportQuizDesignerSlides = async ({
...
-    const numberOfQuestions = quiz.questions.length;
-    for (let i = 1; i <= 20; i++) {
-      if (i > numberOfQuestions) {
-        speakerNotesTagsToDelete.push(`question${i}`);
-      }
-    }
+    const speakerNotesTagsToDelete: SpeakerNotesTag[] = Array.from(
+      { length: 20 - quiz.questions.length },
+      (_, i) => i + quiz.questions.length + 1,
+    ).map((n) => `question${n}` as SpeakerNotesTag);
```
mantagen, line 28: Do we need the assertion here? Will annotation not do?

packages/exports/src/exportSlidesFullLesson.ts
```diff
@@ -20,16 +22,9 @@ export const exportSlidesFullLesson = async ({
...
-    }
-    if (!lesson.cycle2) {
-      speakerNotesTagsToDelete.push("cycle2");
-    }
-    if (!lesson.cycle3) {
-      speakerNotesTagsToDelete.push("cycle3");
-    }
+    const speakerNotesTagsToDelete: SpeakerNotesTag[] = [1, 2, 3]
+      .filter((num) => !lesson[`cycle${num}` as keyof typeof lesson])
+      .map((num) => `cycle${num}` as SpeakerNotesTag);
```
mantagen, line 27: Same question as before around assertion, feels like we're losing a bit of typesafety here vs the previous implementation

packages/exports/src/gSuite/slides/deleteSlides.ts
mantagen: I'm not sure either. How are speaker notes used?

packages/exports/src/schema/quizDocsTemplate.schema.ts
```diff
@@ -1,32 +1,27 @@
...
-  question_4_answer_c: z.string().nullish(),
-  question_5: z.string().nullish(),
-  question_5_answer_a: z.string().nullish(),
-  question_5_answer_b: z.string().nullish(),
-  question_5_answer_c: z.string().nullish(),
-  question_6: z.string().nullish(),
-  question_6_answer_a: z.string().nullish(),
-  question_6_answer_b: z.string().nullish(),
-  question_6_answer_c: z.string().nullish(),
+  ...Object.assign(
```
mantagen, line 19: Nice

mantagen (commented):
Couple comments but looks good! Would want a thorough test from @mikeritson-oak before it goes in, as it touches some brittle parts of the codebase

packages/exports/src/exportQuizDesignerSlides.ts
stefl: For some reason TS complains unless you have the as. I think because it's a runtime string manipulation

packages/exports/src/exportSlidesFullLesson.ts
stefl: Same - because it's strong manipulation it complains without the assertion

packages/exports/src/gSuite/slides/deleteSlides.ts
codeincontext: I think I was involved in this. The slides template has text in the speaker notes with the name of each slide. This code selectively deletes slides based on the text "tag" in the speaker notes

codeincontext: Maybe we need a comment here. This type reflects the values available in the "speaker notes" of the template slides document

packages/exports/src/exportQuizDesignerSlides.ts
codeincontext: I think it could be that typescript can no longer infer the numbers coming from the Array used to build the strings. You should find that it works if you type Array.from or n to be `1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20`

packages/exports/src/gSuite/slides/deleteSlides.ts
mantagen: Ah yes that rings a bell now!

mikeritson-oak (approved):
I've done some general testing, as per the usual test plan, on the preview branch, across browsers and some basic mobile testing. No issues or errors seen during the testing at this point. 👍 
If there is anything specifically that might need more thorough testing then let me know, otherwise we can look at this again if and when merged to main.

packages/exports/src/gSuite/slides/deleteSlides.ts
stefl: Added a comment

packages/exports/src/exportQuizDesignerSlides.ts
stefl: Okay I've done some extra work to tighten the types here even further

PR #469, opened 2025-01-07
test: don't force desktop chromatic mode for all stories

codeincontext:
## Description

In storybook, Chromatic is configured through the parameters object:

```
      modes: {
          desktop: { viewport: "desktop" },
      },
```

We provide a default with just the desktop mode, and then stories can add their own modes to the parameters object. The problem is that Storybook merges values in parameters, which means that there's no way for us to remove the desktop mode for a component which is only relevant on mobile

- Remove default chromatic mode from preview.tsx
- Add explicit chromatic viewport config to each story file
- Add ChromaticValidationDecorator to ensure that chromatic config is set
- Fix config for mobile-wide chromatic mode


mikeritson-oak (approved):
Looks good. 👍

JBR90 (approved):

PR #471, opened 2025-01-08
fix: sonar and typescript issues in download all route

JBR90:
## Description

- Fixes sonar complexity issue in download all route
- Fixes type error with archiver in download all route
- Fixes unnecessary variable type error in cron job

## How to test
Download all should work as before

## Checklist


apps/nextjs/src/app/api/aila-download-all/route.ts
```diff
@@ -95,68 +144,25 @@ async function getHandler(req: Request): Promise<Response> {
...
+      },
+    ),
+  );
+
+  await Promise.all(downloadPromises);
 
-  if (filesProcessed === 0) {
+  if (filesProcessed > 0) {
+    archive.finalize().catch((error: Error) => {
+      Sentry.captureException(error, { level: "error" });
```
codeincontext, line 159: I think it's also worth logging here

codeincontext: Also, should execution continue if finalize fails?

codeincontext: Obviously, if you do want to throw when finalize fails, then we should get the log and sentry report anyway

mikeritson-oak (approved):
Looks good functionally, can be merged to main if all OK with the code. 👍

apps/nextjs/src/app/api/aila-download-all/route.ts
JBR90: Added in the types for Archive - are you happy for this go in @codeincontext  ?

codeincontext: I've sent you a message to work this out. I'm still a bit confused by the move to manual promise syntax here

apps/nextjs/src/app/api/aila-download-all/route.ts
```diff
@@ -90,73 +139,34 @@ async function getHandler(req: Request): Promise<Response> {
...
-  if (filesProcessed === 0) {
+  await Promise.all(downloadPromises);
+
+  if (filesProcessed > 0) {
+    archive.finalize().catch(async (error: Error) => {
+      Sentry.captureException(error, { level: "error" });
+      await kv.set(taskId, "failed");
+      return new Response("Archive finalize error", {
+        status: 500,
+      });
```
codeincontext, line 163: Without returning from `archive.finalize()`, I think this Response will get lost. If we can use the async/await syntax we would be less likely to hit issues like this

JBR90: Have changed to await, the false positive type/lint issues which creating unnecessary problems here. We just merged that to main and a simple await is fine now

PR #475, opened 2025-01-10
fix: async data prep functions should use Promise.resolve

stefl:
## Description

- These data prep functions for exports are marked as async
- Use Promise.resolve to retain the expected async behaviour without changing the signature of each function
- Not user-facing


codeincontext (approved):
> Use Promise.resolve to retain the expected async behaviour without changing the signature of each function

I think this is an issue in linting, as the behaviour in both cases is identical. As far as I know, returning `Promise.resolve` in an async function is a no-op, the goal in this PR is to fix the lint warnings saying "this is an async function and it doesn't look like it needs to be"

PR #476, opened 2025-01-10
chore: derive prisma type using ReturnType

stefl:
## Description

- Removes a linting issue around unused variables
- Derives the type of PrismaClientWithAccelerate using ReturnType


codeincontext (approved):
Seems like a nicer approach

PR #478, opened 2025-01-10
chore: add error logging, remove async function in ingest script

stefl:
## Description

- Fix some issues with ingest around async functions not being expected, add logging
- Not user-facing

### DIFF
--- a/packages/ingest/src/import-lessons/importLessonsFromCSV.ts
+++ b/packages/ingest/src/import-lessons/importLessonsFromCSV.ts
@@ -38,6 +38,10 @@ export function importLessonsFromCSV({
             subject_slug,
           } = row;
 
+          if (typeof lesson_id !== "string") {
+            throw new Error("lesson_id must be a string");
+          }
+
           const parsedLesson = RawLessonSchema.parse({
             oakLessonId: parseInt(lesson_id),
             lessonSlug: lesson_slug,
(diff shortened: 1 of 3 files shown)


packages/ingest/src/import-lessons/importLessonsFromCSV.ts
```diff
@@ -60,18 +64,18 @@ export function importLessonsFromCSV({
...
             data,
             fn: async (data) => {
               await prisma.ingestLesson.createMany({
                 data,
               });
             },
             chunkSize: 30000,
+          }).then(() => {
+            log.info(`Imported ${parsedLessons.length} lessons from CSV`);
+            resolve();
```
codeincontext, line 77: What's the error being solved with this change? That the `on` function doesn't accept an async function? The replacement looks like a step backwards to me so it would be useful to know

stefl: Yes, that's correct. There was a linting complaint about on not accepting an async function

codeincontext: Ok, yeah. You can work around it by wrapping it in an arrow function like:

```ts
() => void (async () => {
  ...
})()
```

But that's not too readable!

I'm happy with this to resolve the error

codeincontext (approved):

PR #480, opened 2025-01-10
chore: fix linting errors around await / logging

stefl:
## Description

- Fixes several small linting errors around async functions needing Promise.resolve
- Adds aiLogger in tests
- Fixes a few small issues around not awaiting promises

### DIFF
--- a/packages/aila/src/features/persistence/adaptors/file/index.ts
+++ b/packages/aila/src/features/persistence/adaptors/file/index.ts
@@ -1,5 +1,8 @@
 import { AilaPersistence } from "../..";
-import type { AilaChatService, AilaServices } from "../../../../core/AilaServices";
+import type {
+  AilaChatService,
+  AilaServices,
+} from "../../../../core/AilaServices";
 import type { AilaGeneration } from "../../../generation/AilaGeneration";
 
 export class AilaFilePersistence extends AilaPersistence {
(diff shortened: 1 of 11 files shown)


packages/eslint-config/src/index.mjs
```diff
@@ -137,6 +138,24 @@ const config = [
       },
     },
   },
+  {
+    files: ["**/packages/db/prisma/zod-schemas/*.ts"],
```
stefl, line 142: Ignore linting errors in generated zod schemas

packages/aila/src/features/persistence/adaptors/file/index.ts
```diff
@@ -1,26 +1,28 @@
...
   constructor({ aila, chat }: { aila: AilaServices; chat: AilaChatService }) {
     super({ aila, chat, name: "AilaFilePersistence" });
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async upsertChat(): Promise<void> {
     // Implementation goes here
-    throw new Error("Not implemented");
+    return Promise.reject(new Error("Not implemented"));
   }
```
codeincontext, line 16: Was this creating a lint error? I can't see any difference, is it just that the lint ruls isn't smart enough to see that throwing in an async function returns a promise?

codeincontext (commented):
This looks good so far, but I'm not sure about the changes to `Promise.reject` in an async function. I'd expect to either have an async function or to return `Promise.reject`. If you return the reject then I imagine it doesn't need the async keyword either

packages/aila/src/features/persistence/adaptors/file/index.ts
stefl: I think so - it was showing as a linting error in VSCode for me and appearing in the Problems tab without this

stefl:
@codeincontext because these are extending from an abstract class they need to be async and return a promise - is there a better pattern?

codeincontext:
> @codeincontext because these are extending from an abstract class they need to be async and return a promise - is there a better pattern?



> @codeincontext because these are extending from an abstract class they need to be async and return a promise - is there a better pattern?

Oh, I see! I think it's not really about the promise, but about there being no path to return a value as it always throws. That's definitely in the category of changing the code to resolve a lint error we can ignore - but that's fine

codeincontext (approved):

PR #481, opened 2025-01-13
chore: linting for unused errors and imports

stefl:
## Description

- Removes unused errors that are not logged, and don't seem to need to be logged
- Removes unused imports
- Removes some unused ts-ignore comments
- Not user-facing


apps/nextjs/src/components/AppComponents/Chat/sidebar-actions.tsx
```diff
@@ -107,7 +107,7 @@ export function SidebarActions({ chat }: SidebarActionsProps) {
                     router.refresh();
                     router.push("/");
                     toast.success("Chat deleted");
-                  } catch (err) {
+                  } catch {
```
JBR90, line 110: Should we report this ?

JBR90 (approved):

PR #483, opened 2025-01-13
fix: resolve "error typed" issue with linting

stefl:
## Description

- Resolves the "error typed" problem that we are facing
- This appears to be due to a mix of aliases not resolving in different scenarios and per-project linting not being set up correctly to match the typescript config
- Introduces per-package linting rules 
- Ensures that we use aliases for "cms" etc. so that it is clear these are aliased rather than assuming they are imported from ./src/... within the Nextjs project
- No user-facing changes


apps/nextjs/tsconfig.json
```diff
@@ -17,14 +17,18 @@
...
-    "baseUrl": "./src",
+    "baseUrl": ".",
     "paths": {
-      "@/components/*": ["components/*"],
-      "@/app/*": ["app/*"],
-      "@/lib/*": ["lib/*"],
-      "@/utils/*": ["utils/*"],
-      "@/assets/*": ["assets/*"],
-      "@/storybook/*": ["../.storybook/*"]
+      "@/components/*": ["src/components/*"],
```
stefl, line 22: Add aliases for all top level directories in src

packages/eslint-config/src/index.mjs
```diff
@@ -13,6 +13,78 @@ import globals from 'globals';
...
 import { rules } from "./rules.mjs";
 
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const projectRoot = path.resolve(__dirname, '../../..');
+
+const packages = [
```
stefl, line 23: This is a major part of the fix - per-project linting definitions so that eslint uses the appropriate rules, and the correct tsconfig is used for each package for linting purposes.

JBR90 (approved):

PR #488, opened 2025-01-15
feat: decouple aila engine stream reader

mantagen:
## Description

- in as simple way as possible, allow Aila to use a different 'engine'
- only routes traffic away _after_ lesson is initialised with (title,subject,key_stage,topic)
- it doesn't concern itself with any persistence or moderation currently. It would act as a drop in replacement for **the combination of the Prompt Builder and OpenAI**.
- due to improvements in streaming objects, a lot of the 'chunk' parsing will become redundant, however as I'm not wanting to delve into that part of the codebase (and risk jeopardising backward compatibility).

### DIFF
--- a/packages/aila/src/core/chat/AilaStreamHandler.ts
+++ b/packages/aila/src/core/chat/AilaStreamHandler.ts
@@ -94,8 +94,14 @@ export class AilaStreamHandler {
 
     const AILA_ENGINE_STREAM_URL = process.env.AILA_ENGINE_STREAM_URL;
 
-    const lastUserMessage = this._chat.messages.slice(-1)[0];
     if (AILA_ENGINE_STREAM_URL) {
+      /**
+       * If the AILA_ENGINE_STREAM_URL is defined, we will use it to stream the
+       * engine's response rather than using an AilaPromptBuilder with OpenAI's API.
+       *
+       * It allows us to experinent with different paradigms in a safe and unobtrusive way.
+       */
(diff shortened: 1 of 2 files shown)


packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -89,9 +90,20 @@ export class AilaStreamHandler {
...
-    this._streamReader =
-      await this._chat.createChatCompletionObjectStream(messages);
+    const currentLessonPlan = this._chat.aila.lessonPlan;
+
+    const AILA_ENGINE_STREAM_URL = process.env.AILA_ENGINE_STREAM_URL;
+
+    const lastUserMessage = this._chat.messages.slice(-1)[0];
+    if (AILA_ENGINE_STREAM_URL) {
+      this._streamReader = await createStreamReaderFromURL(
+        `${AILA_ENGINE_STREAM_URL}?prompt=${JSON.stringify(currentLessonPlan)}\n\n${lastUserMessage?.content}`,
```
mantagen, line 111: Currently, this is super naive. Just sends the current lesson plan concatenated with the last user message to the server. i.e. not the full message history.

Plan to amend that on a subsequent update.

mantagen:
Closing in favour of [a different approach in future](https://www.notion.so/oaknationalacademy/Aila-Engine-18a26cc4e1b180bf8b3cff7b4166dcf3)

PR #496, opened 2025-01-20
feat: store user actions

tomwisecodes:
## Description

- Moves user actions to be in the store
- Moves streaming status to be in the store

Fixes AI-759, AI-760


apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -293,55 +293,15 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
 
   // Handle queued user actions and messages
 
-  const [queuedUserAction, setQueuedUserAction] = useState<string | null>(null);
```
tomwisecodes, line 296: Was a bit unsure about what this does, as far as I could see was just internal to this compoenent

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -56,14 +57,14 @@ export type ChatContextProps = {
     message: Message | CreateMessage,
     chatRequestOptions?: ChatRequestOptions | undefined,
   ) => Promise<string | null | undefined>;
-  reload: () => void;
-  stop: () => void;
+  // reload: () => void;
+  // stop: () => void;
```
tomwisecodes, line 61: can delete this pre merge

apps/nextjs/src/components/AppComponents/Chat/chat-quick-buttons.tsx
```diff
@@ -107,20 +102,22 @@ const QuickActionButtons = () => {
...
-              stop();
-            }}
-            testId="chat-stop"
-          >
-            <span className="opacity-50">
-              <IconStop className="mr-3" />
-            </span>
-            <span className="font-light text-[#575757]">Stop</span>
-          </ChatButton>
+          <>
```
tomwisecodes, line 105: will remove

apps/nextjs/src/components/AppComponents/Chat/chat-quick-buttons.stories.tsx
```diff
@@ -97,7 +97,6 @@ export const StreamingWithQueuedUserAction: Story = {
   args: {},
   parameters: {
     chatContext: {
-      queuedUserAction: "regenerate",
       ailaStreamingStatus: "StreamingLessonPlan",
```
tomwisecodes, line 100: Due to a change in types had to remove this, @codeincontext might be in teh best place to comment on how we might reintroduce

apps/nextjs/src/stores/chatStore/actionFunctions/handleExecuteQueuedAction.ts
```diff
@@ -0,0 +1,31 @@
+export function handleExecuteQueuedAction(set, get) {
+  return async () => {
+    const { queuedUserAction, isExecutingAction } = get();
+
+    if (!queuedUserAction || isExecutingAction) return;
+
+    set({ isExecutingAction: true });
+    const actionToExecute = queuedUserAction;
+    set({ queuedUserAction: null });
```
codeincontext, line 9: Both of these sets can be combined into one call

apps/nextjs/src/stores/chatStore/actionFunctions/handleExecuteQueuedAction.ts
```diff
@@ -0,0 +1,31 @@
+export function handleExecuteQueuedAction(set, get) {
+  return async () => {
+    const { queuedUserAction, isExecutingAction } = get();
```
codeincontext, line 3: I think we should destructure the actions like append and reload up here as well

apps/nextjs/src/stores/chatStore/hooks.ts
```diff
@@ -1,17 +1,43 @@
...
+    setStop(stop);
+  }, [stop, setStop]);
+
+  useEffect(() => {
+    setAppend(append);
+  }, [append, setAppend]);
+
+  useEffect(() => {
+    setReload(reload);
+  }, [reload, setReload]);
```
codeincontext, line 42: There are a couple of related points here which might make things simpler when you add them together:
- Are you sure that these functions are stable? When I tried this I found that they were updating more often than they should be, potentially on every render. I think we only need to set the initial value. In my branch I did this with `useEffect(..., [])`
- I think these actions can all be set in one call to the store, like `state.setAiSdkActions`?
- Should we consider putting the actions under a single key in the store, like actions.stop, actions.regenerate, etc?

apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -12,56 +15,53 @@ type ChatStore = {
...
+  executeQueuedAction: () => Promise<void>;
 };
 
 export const useChatStore = create<ChatStore>((set, get) => ({
   // From AI SDK
   isLoading: false,
   stableMessages: [],
   streamingMessage: null,
+  queuedUserAction: null,
+  isExecutingAction: false,
```
codeincontext, line 48:
```suggestion

  // Queueing
  queuedUserAction: null,
  isExecutingAction: false,
```

apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -12,56 +15,53 @@ type ChatStore = {
...
-        ...(nextStableMessages && {
-          stableMessages: nextStableMessages,
-        }),
-      });
-    }
+  stop: () => {},
+  append: async () => "",
+  reload: () => {},
+  setAppend: (append) => set({ append }),
+  setReload: (reload) => set({ reload }),
```
codeincontext, line 55: I think we should group these functions used for config elsewhere, as they aren't really actions that we should think about when it comes to use the store

apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -12,56 +15,53 @@ type ChatStore = {
...
+  queueUserAction: (action: string) => {
+    set({ queuedUserAction: action });
   },
+  setStop: (stop) => set({ stop }),
   setIsLoading: (isLoading) => set({ isLoading }),
+
+  // Action functions
+  executeQueuedAction: handleExecuteQueuedAction(set, get),
+  setMessages: handleSetMessages(set, get),
```
codeincontext, line 64: I like this approach. I can imagine that some of the other actions like `append` and `stop` will eventually become this type of function which does multiple things

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
codeincontext: This is the behaviour that lets you start typing and submit a message while the last one is still moderating. It makes the page feel quicker if you can interact once the LLM response has finished, but we can only stream one thing at a time, and the moderation takes ~5 seconds

It looks like it should still work, but have you been able to give that a test?

apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -12,10 +15,28 @@ type ChatStore = {
   isLoading: boolean;
   stableMessages: ParsedMessage[];
   streamingMessage: ParsedMessage | null;
+  queuedUserAction: string | null;
+  isExecutingAction: boolean;
```
codeincontext, line 19: I think in the context of the store, this might make more sense as `isExecutingQueuedAction`?

apps/nextjs/src/stores/chatStore/actionFunctions/handleExecuteQueuedAction.ts
```diff
@@ -0,0 +1,31 @@
...
+      } else if (actionToExecute === "regenerate") {
+        reload();
+      } else {
+        await append({
+          content: actionToExecute,
+          role: "user",
+        });
+      }
+    } catch (error) {
+      console.error("Error handling queued action:", error);
```
codeincontext, line 26: Let's use a logger here

apps/nextjs/src/stores/chatStore/actionFunctions/handleSetMessages.ts
```diff
@@ -0,0 +1,68 @@
...
+      } else if (
+        currentMessageData.content.includes('"type":"prompt"') ||
+        currentMessageData.content.includes('\\"type\\":\\"prompt\\"')
+      ) {
+        streamingStatus = "StreamingChatResponse";
+      } else if (currentMessageData.content.includes("CHAT_START")) {
+        streamingStatus = "StreamingLessonPlan";
+      } else {
+        streamingStatus = "Loading";
+      }
```
codeincontext, line 57: I think it would make sense to extract this to a function that takes `currentMessageData` and returns the streaming status. It would be a bit cleaner than having the let variable for streamingStatus, and would make this function less noisy

apps/nextjs/src/stores/chatStore/actionFunctions/handleExecuteQueuedAction.ts
```diff
@@ -0,0 +1,31 @@
+export function handleExecuteQueuedAction(set, get) {
+  return async () => {
+    const { queuedUserAction, isExecutingAction, actions } = get();
+    const { append, reload } = actions;
+
+    if (!queuedUserAction || isExecutingAction) return;
```
codeincontext, line 6: I might have mentioned this before, but I think we should rename this to `isExecutingQueuedAction`

apps/nextjs/src/stores/chatStore/actionFunctions/handleQueueUserAction.ts
```diff
@@ -0,0 +1,8 @@
+import type { ChatStore } from "..";
+
+export const handleQueueUserAction =
+  (set: (partial: Partial<ChatStore>) => void, get: () => ChatStore) =>
+  async (action: string) => {
+    set({ queuedUserAction: action });
+    await get().executeQueuedAction();
```
codeincontext, line 7: Just checking that you want to execute the queued action at the same time as queueing it. I would expect this to queue it first, and then later execute when the streaming finishes

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -293,55 +293,15 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
       executeQueuedAction();
     }
   }, [hasFinished, executeQueuedAction]);
 
-  const stop = useCallback(() => {
-    if (queuedUserAction) {
-      setQueuedUserAction(null);
-    } else {
-      stopStreaming();
-    }
```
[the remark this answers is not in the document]
codeincontext: @tomwisecodes Still missing?

PR #497, opened 2025-01-20
feat: admin button to reverse safety violations

codeincontext:
## Description

When a user receives a toxic moderation they also get a safety violation. After 3 safety violations they are banned. Some users have had a genuine use case for trying lessons in areas which we don't allow, which has caused them to be banned. We want to be able to unban them and remove those violations while still keeping the offending lessons locked

- Fix bug where violations weren't recorded if the inngest enqueue for the slack notification failed (helps identify the issue in local dev)
- Add UI to the admin chat page to "reverse" a violation
- Return related violations from `admin.getModerations` endpoint
- Add removeSafetyViolation to admin router

## Issue(s)

Fixes AI-754

## How to test

1. Create 3 toxic lessons with `mod:tox`, making a note of the URLs of each lesson
2. User should be banned
3. Visit https://oak-ai-lesson-assistant-git-feat-reverse-violations.vercel-preview.thenational.academy/admin/aila/{CHAT-ID}
4. It should show that there was a violation
5. Reversing the first violation should unban the user

## Screenshots

CleanShot 2025-01-20 at 18 19 52@2x


apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
```diff
@@ -2,25 +2,67 @@ import { useState } from "react";
...
+        <OakP $font="body-2" $mr="space-between-m">
+          <OakP $font="heading-7">
+            This moderation triggered a safety violation.
+          </OakP>{" "}
+          Reversing may unban a banned user
+        </OakP>
+        <OakSmallSecondaryButton
+          iconName="cross"
+          onClick={() =>
+            void removeSafetyViolation.mutateAsync({ recordId: recordId })
```
Comment (stefl) [inline apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:37]: For lines like this which could fail because of a network or serve error I've been adding sentry and toast

Comment (stefl) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:37]: But perhaps that's the wrong pattern? We should give user feedback in these two cases

packages/api/src/router/admin.ts
```diff
@@ -74,4 +91,15 @@ export const adminRouter = router({
...
+  removeSafetyViolation: adminProcedure
+    .input(z.object({ recordId: z.string() }))
+    .mutation(async ({ input, ctx }) => {
+      const { recordId } = input;
+      const { prisma } = ctx;
+
+      // remove safety violation (and potentially unban user)
+      log.info("Removing safety violation", { recordId });
+      const safetyViolations = new SafetyViolations(prisma);
+      await safetyViolations.removeViolationsByRecordId(recordId);
```
stefl, line 103: Should this return the deleted violation? Do we want logging of the intended safety violation having been deleted?

apps/nextjs/src/app/api/chat/webActionsPlugin.ts
```diff
@@ -56,17 +57,24 @@ export const createWebActionsPlugin: PluginCreator = (
...
+          id: userId,
+        },
+        data: {
+          chatId: aila.chatId || "Unknown",
+          categories: moderation.categories as string[],
+          justification: moderation.justification ?? "Unknown",
+        },
+      });
+    } catch (e) {
+      log.error("Error sending slack notification", e);
```
stefl, line 74: This is async so "error scheduling slack..."?

stefl (approved):
Just a couple of comments here about logging and user feedback on error

apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
Comment (codeincontext) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:37]: Good point. I put this together quite quickly so I didn't put too much thought into it. I'll look into that 👍

apps/nextjs/src/app/api/chat/webActionsPlugin.ts
codeincontext: Good call, thanks

packages/api/src/router/admin.ts
codeincontext: That makes sense, I can do that 👍

apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
Comment (codeincontext) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:37]: Another option would be to render an error message in-line, but I don't think we have precedent for that

Comment (mantagen) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:37]: I think for admin facing features doing a toast is probably fine and simplest for now

apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
```diff
@@ -2,25 +2,67 @@ import { useState } from "react";
...
+          }
+          isLoading={removeSafetyViolation.isLoading}
+          disabled={!!removeSafetyViolation.isSuccess}
```
Comment (mantagen) [inline apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:40]: disable if `isLoading` also?

mantagen (approved):
LGTM

apps/nextjs/src/app/admin/aila/[chatId]/view.tsx
Comment (codeincontext) [reply apps/nextjs/src/app/admin/aila/[chatId]/view.tsx:40]: That makes sense 👍

mikeritson-oak (commented):
I've checked this in the preview branch and functionality is as expected, Further checks can be made in the release candidate.

PR #499, opened 2025-01-21
chore: maths quiz tighter types

mantagen:
## Description

- use remeda `pick` to get lesson plan subset
- improve types on `apiCallingUtils` (though curerntly this isn't called anywhere so haven't checked that the new types align with what the caller expects)


gclomax (approved):
LGTM

PR #502, opened 2025-01-21
refactor: moderation from chatProvider to moderation store

JBR90:
## Description

- Simplifies logic from lesson chat moderation to store.
- Triggers moderation from chat store on 'Idle' streaming status
- Moderations store in provider with shared state actions
- Replaces reference to moderation in  lessonChat  
- Moves trpc calls to store - creates a new trpcClient
- Create store decorator and update stories

## How to test

Moderations and toxic moderations should work the same as before

Fixes AI-757

## Checklist


apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      set({
+        ...(nextStableMessages && {
+          stableMessages: nextStableMessages,
+        }),
+        streamingMessage: null,
+      });
+    } else {
+      // The latest message is streaming, previous messages are stable
+      const currentMessageData = messages[messages.length - 1];
+      // invariant(currentMessageData, "Should have at least one message");
```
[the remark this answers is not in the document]
codeincontext: I saw that there was always a message while `isLoading` is true. Even on a fresh chat, the first message is the user message with the initial lesson title

If we want to clear the messages then I imagine that `isLoading` wouldn't be true, and also that none of this logic is relevant so I guess that we can stick to `clearMessages`?

apps/nextjs/src/stores/moderationStore/actionFunctions/handleInitialiseModerations.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const toxicInitial = mods.find(isToxic) ?? null;
+  const lastMod = mods[mods.length - 1] ?? null;
+  const toxicMod = lastMod && isToxic(lastMod) ? lastMod : toxicInitial;
+
+  set({
+    moderations: mods,
+    toxicInitialModeration: toxicInitial,
+    lastModeration: lastMod,
```
codeincontext, line 32: Values like `lastModeration` seem very easy for any consuming code to calculate. Do we need it denormalised in the store? It's usually safer to keep computed values computed so that they don't accidentally drift

Is there a good reason to be setting it upfront?

codeincontext: Oh, I see that the current version in ChatProvider uses useState rather than useMemo. I guess this is a sensible first version then!

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
```diff
@@ -0,0 +1,10 @@
+import type { PersistedModerationBase } from "@oakai/core/src/utils/ailaModeration/moderationSchema";
+import { useChatStore } from "src/stores/chatStore";
+
+export const handleToxicModeration = (
```
codeincontext, line 4:
```suggestion
export const handleFlagToxicModeration = (
```

Maybe?

codeincontext: Or maybe the opposite. Maybe remove the `flag` from other places as it's not very clear what flagging is

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
```diff
@@ -0,0 +1,10 @@
...
+  mod: PersistedModerationBase | null,
+  set: (state: { toxicModeration: PersistedModerationBase | null }) => void,
+) => {
+  set({ toxicModeration: mod });
+  useChatStore.getState().clearMessages();
```
codeincontext, line 9: We should be aware that this is a deviation from the current behaviour. At the moment it tells `useChat` to clear the messages, but this will presumably just remove our current copy

It would be an issue if `useChat` updated `messages` and the chat store synced again, but in the case of a toxic moderation that seems unlikely

apps/nextjs/src/stores/chatStore/index.ts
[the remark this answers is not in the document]
JBR90: I think it's fine as it is, if i create a fake toxic moderation it works as expected

apps/nextjs/src/stores/moderationStore/actionFunctions/handleInitialiseModerations.tsx
JBR90: Yeah i was reproducing the logic here, it's also used in a few places though out the chat

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
JBR90: Yeah there are no more messages if there is a toxic lesson, only the option to start a new lesson

apps/nextjs/.storybook/decorators/ModerationStoreDecorator.tsx
```diff
@@ -0,0 +1,29 @@
...
+  type ModerationStore,
+} from "../../src/stores/moderationStore";
+
+declare module "@storybook/csf" {
+  interface Parameters {
+    moderationStoreState?: Partial<ModerationStore>;
+  }
+}
+
+export const ModerationStoreDecorator: Decorator = (Story, { parameters }) => {
```
JBR90, line 16: There is a bug on the doc's page with this pattern. Currently the docs pages take the state of the last story for the component and ignores the parameters. So if the last story has moderation props that docs will show this in each variation of the component on the docs page. Tested this with the store in a provider pattern and there are no issues. There may be another approach though?

apps/nextjs/src/components/AppComponents/Chat/chat-right-hand-side-lesson.tsx
```diff
@@ -59,6 +60,14 @@ const ChatRightHandSideLesson = ({
 
   const endOfDocRef = useRef<HTMLDivElement>(null);
 
+  const isModerationsLoading = useModerationStore(
+    (state) => state.isModerationsLoading,
+  );
+
+  if (isModerationsLoading) {
```
JBR90, line 67: This has been moved here from the provider.

apps/nextjs/.storybook/decorators/ModerationStoreDecorator.tsx
JBR90: Tried iframes but `inlineStories: false` is not working, there no iframes rendered from adding to preview or individual stories

apps/nextjs/src/components/StoreComponents/ModerationSync.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const {
+    data: moderations,
+    isLoading: isModerationsLoading,
+    refetch: refetchModerations,
+  } = trpc.chat.appSessions.getModerations.useQuery(
+    { id },
+    {
+      refetchOnMount: true,
+      refetchOnWindowFocus: true,
+      staleTime: 0,
```
codeincontext, line 16: We will need to have a look at this before wrapping up this work

apps/nextjs/src/components/StoreComponents/ModerationSync.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    if (moderations !== undefined) {
+      setModeration(moderations || []);
+    }
+  }, [moderations, setModeration]);
+
+  useEffect(() => {
+    void refetchModerations();
+  }, [refetchModerations]);
```
codeincontext, line 39: I think this is fine to get things working, but we will be trying to move away from having logic and data loading powered by hooks and effects. I think all of this behaviour should live in an action function, though that's something we can look at in a follow up

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
codeincontext: We'll just need to be careful that we don't accidentally add behaviour that populates `messages` from the trpc getChat call. That gets triggered a lot and it could inadvertently repopulate the chat store

apps/nextjs/src/stores/moderationStore/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    handleUpdateModerationState(mod, set, get);
+  },
+  // reset
+  reset: (mod) => {
+    set({
+      moderations: mod.moderations ?? [],
+      toxicInitialModeration: mod.toxicModeration ?? null,
+      toxicModeration: mod.toxicModeration ?? null,
+      lastModeration: mod.lastModeration ?? null,
+    });
```
codeincontext, line 49: There's a nice pattern for this in #496. We can align all the stores once that's merged

apps/nextjs/src/stores/moderationStore/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  },
+}));
+
+export type ModerationStoreState = ReturnType<
+  typeof useModerationStore.getState
+>;
```
codeincontext, line 55: What's the difference between this and `ModerationStore`?

JBR90: Thats's not being used will remove

codeincontext (approved):
This looks good to me. There are some parts like data loading which we will want to bring into action functions rather then syncing from a hookl but that can some later. It's most useful to have a first version of this merged

Re storybook: Options are to move the store to a context (which we want to avoid for now), or to just disable docs for stories with problems. I think that's fine for now if the reason is written in a comment

mikeritson-oak (changes requested):
I've taken a look at the preview branch. As far as I could see, sensitive moderations seemed to be triggering as expected, certainly nothing out of the ordinary happened.
Toxic moderations seems to be having some issues. I noted that after a toxic moderation and lesson block, when I was trying to make another lesson, I was receiving further warnings each time. This happened twice. I then went to record the issue and on this attempt I received a sensitiave moderation instead, then carried on trying and eventually, after about 3-4 messages, my account was locked. 
It seems something is amiss with toxic moderation. I'm happy to have a call about this or check it out together?

codeincontext (approved):

apps/nextjs/src/components/AppComponents/Chat/toxic-moderation-view.tsx
```diff
@@ -45,6 +49,7 @@ const ToxicModerationView = ({
             variant="icon-only"
             title="New lesson"
             onClick={() => {
+              clearModerations();
```
codeincontext, line 52: Is this needed if we're now using the context? Will the store get recreated anyway?

JBR90: Yeah can get rif of this

apps/nextjs/src/stores/moderationStore/actionFunctions/handleFetchModeration.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const handleFetchModerations = async (
+  set: (state: Pick<ModerationStore, "isModerationsLoading">) => void,
+  get: () => ModerationStore,
+  trpcUtils: TrpcUtils,
+) => {
+  const { updateModerationState, id } = get();
+
+  if (!id) {
+    return;
+  }
```
codeincontext, line 16: Is this an expected situation, or would it represent an error? It could be worth an `invariant`

JBR90: I guess there should be an id at that point so yes maybe a invariant would be better

apps/nextjs/src/stores/moderationStore/actionFunctions/handleFetchModeration.tsx
```diff
@@ -0,0 +1,33 @@
...
+    isModerationsLoading: true,
+  });
+  try {
+    const fetchedModerations =
+      await trpcUtils.chat.appSessions.getModerations.fetch({
+        id,
+      });
+    updateModerationState([...fetchedModerations]);
+  } catch (error) {
+    Sentry.captureException(error);
```
codeincontext, line 27:
```suggestion
    log.error("Error fetching moderation", error)
    Sentry.captureException(error);
```

apps/nextjs/src/stores/moderationStore/actionFunctions/handleFetchModeration.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }
+  set({
+    isModerationsLoading: true,
+  });
+  try {
+    const fetchedModerations =
+      await trpcUtils.chat.appSessions.getModerations.fetch({
+        id,
+      });
+    updateModerationState([...fetchedModerations]);
```
codeincontext, line 25: I don't think that we should be changing the identity of this object on purpose. Did you have a problem with just passing them as-is?

apps/nextjs/src/stores/moderationStore/index.ts
```diff
@@ -0,0 +1,88 @@
...
+import type { ChatStore } from "../chatStore";
+import { logStoreUpdates } from "../zustandHelpers";
+import { handleFetchModerations } from "./actionFunctions/handleFetchModeration";
+import { handleToxicModeration } from "./actionFunctions/handleToxicModeration";
+import { handleUpdateModerationState } from "./actionFunctions/handleUpdateModerationState";
+
+export type ModerationStore = {
+  chatActions?: Pick<ChatStore, "setMessages">;
+
+  id: string | null;
```
codeincontext, line 16: Can this be null now that it's passed into the constructor?

apps/nextjs/src/stores/moderationStore/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  reset: (params: Partial<ModerationStore>) => void;
+  clearModerations: () => void;
+};
+
+export const createModerationStore = ({
+  id = null,
+  initialValues = {},
+  trpcUtils,
+}: {
+  id?: string | null;
```
codeincontext, line 37: Should we just always expect this? Or do you think that it would help with storybook and testing to make it optional?

apps/nextjs/src/stores/moderationStore/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // reset
+    reset: (mod) => {
+      set({
+        moderations: mod.moderations ?? [],
+        toxicInitialModeration: mod.toxicModeration ?? null,
+        toxicModeration: mod.toxicModeration ?? null,
+        lastModeration: mod.lastModeration ?? null,
+      });
+    },
```
codeincontext, line 69: Do we still need reset?

apps/nextjs/src/utils/trpc.tsx
```diff
@@ -22,7 +22,7 @@ type CombinedRouter = AppRouter & ChatAppRouter;
 
 const trpc = createTRPCReact<CombinedRouter>();
 
-const discriminatingRouterLink: TRPCLink<CombinedRouter> = (runtime) => {
+export const discriminatingRouterLink: TRPCLink<CombinedRouter> = (runtime) => {
```
codeincontext, line 25: We don't need this export any more

packages/api/src/router/appSessions.ts
```diff
@@ -100,9 +100,9 @@ export const appSessionsRouter = router({
       const { id } = input;
 
       const moderations = await getSessionModerations(id);
-      return moderations.filter((moderation) =>
-        userIsOwner(moderation, ctx.auth),
-      );
+      return moderations
+        .filter((moderation) => userIsOwner(moderation, ctx.auth))
+        .sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
```
codeincontext, line 105: It makes a lot of sense to me to sort it in the API. Nicely done

apps/nextjs/src/stores/moderationStore/actionFunctions/handleFetchModeration.tsx
JBR90: Was left from previous version, will change

mikeritson-oak (approved):
Checked in the preview branch. sensitive and toxic moderations are working fine. I've also checked account suspension and locking, plus blocking of shared content. Let's move it to a release candidate for some more testing there. 👍

PR #511, opened 2025-01-23
chore: refactor calls from provider to state

tomwisecodes:
## Description

- reduces returns from the provider and changes them to the store


apps/nextjs/.storybook/decorators/ChatStoreDecorator.tsx
```diff
@@ -10,8 +10,9 @@ import {
 declare module "@storybook/csf" {
   interface Parameters {
     chatStoreState?: {
-      ailaStreamingStatus: AilaStreamingStatus;
+      ailaStreamingStatus?: AilaStreamingStatus;
       queuedUserAction?: string | null;
+      isStreaming?: boolean;
```
codeincontext, line 15: Have a look at the screenshot I've sent in slack for typing this a bit nicer with `Partial<ChatStore>`

apps/nextjs/src/components/AppComponents/Chat/chat-layout.tsx
```diff
@@ -13,7 +13,8 @@ export interface ChatLayoutProps {
 }
 
 export const ChatLayout = ({ className }: Readonly<ChatLayoutProps>) => {
-  const { isLoading, lessonPlan, messages } = useLessonChat();
+  const { lessonPlan, messages } = useLessonChat();
+  const { isLoading } = useChatStore();
```
codeincontext, line 17:
```suggestion
  const isLoading = useChatStore((state) => state.isLoading);
```

apps/nextjs/src/components/AppComponents/Chat/chat-left-hand-side.tsx
```diff
@@ -26,7 +26,7 @@ const ChatLeftHandSide = ({
   demo,
   isDemoUser,
 }: Readonly<ChatLeftHandSideProps>) => {
-  const { chatAreaRef } = useLessonChat();
+  const { chatAreaRef } = useChatStore();
```
codeincontext, line 29:
```suggestion
  const chatAreaRef = useChatStore(state => state.chatAreaRef);
```

apps/nextjs/src/components/AppComponents/Chat/chat-list/index.tsx
```diff
@@ -111,7 +111,8 @@ export const ChatMessagesDisplay = ({
   ailaStreamingStatus,
   demo,
 }: ChatMessagesDisplayProps) => {
-  const { lessonPlan, isStreaming } = useLessonChat();
+  const { lessonPlan } = useLessonChat();
+  const { isStreaming } = useChatStore();
```
codeincontext, line 115:
```suggestion
  const isStreaming = useChatStore(state => state.isStreaming);
```

apps/nextjs/src/stores/chatStore/hooks/useChatStoreMirror.ts
```diff
@@ -30,4 +32,10 @@ export const useChatStoreMirror = (
     setAiSdkActions({ stop, append, reload });
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
+
+  const storeSetInput = useChatStore((state) => state.setInput);
+
+  useEffect(() => {
+    storeSetInput(input);
+  }, [input, storeSetInput]);
```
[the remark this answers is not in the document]
tomwisecodes: Can discuss next week, potentially it doesn't need to hit the store at all

apps/nextjs/src/app/aila/page-contents.tsx
```diff
@@ -8,10 +8,14 @@ import Layout from "@/components/AppComponents/Layout";
...
 import { useReactScan } from "@/hooks/useReactScan";
 import LessonPlanTrackingProvider from "@/lib/analytics/lessonPlanTrackingContext";
+import { useChatStore } from "@/stores/chatStore";
+import useStoreReset from "@/stores/chatStore/hooks/useStoreReset";
 
 const ChatPageContents = ({ id }: { readonly id: string }) => {
   useReactScan({ component: LessonPlanDisplay, interval: 10000 });
-
+  const setId = useChatStore((state) => state.setId);
+  setId(id);
```
codeincontext, line 17: Will this set the ID every thing the component renders? That sounds like the kind of pattern we're trying to get away from

apps/nextjs/src/components/AppComponents/Chat/chat-layout.tsx
```diff
@@ -13,7 +13,11 @@ export interface ChatLayoutProps {
 }
 
 export const ChatLayout = ({ className }: Readonly<ChatLayoutProps>) => {
-  const { isLoading, lessonPlan, messages } = useLessonChat();
+  const { lessonPlan, messages } = useLessonChat();
+  const stableMessages = useChatStore((state) => state.stableMessages);
+  const streamingMessage = useChatStore((state) => state.streamingMessage);
+  const messagesFromStore = [...stableMessages, streamingMessage];
```
codeincontext, line 19: This is going to be a different array object each render, I think it needs to be in a useMemo

It's ok to group them into one list as a drop in replacement for now, but let's put a TODO in here to remind us to split the rendering of stableMessages and streamingMessage so we don't need to rerender everything when just the streamingMessage is changing

apps/nextjs/src/components/AppComponents/Chat/chat-right-hand-side-lesson.tsx
```diff
@@ -23,7 +24,10 @@ const ChatRightHandSideLesson = ({
   closeMobileLessonPullOut,
   demo,
 }: Readonly<ChatRightHandSideLessonProps>) => {
-  const { messages } = useLessonChat();
+  // const { messages } = useLessonChat();
+  const stableMessages = useChatStore((state) => state.stableMessages);
+  const streamingMessage = useChatStore((state) => state.streamingMessage);
+  const messages = [...stableMessages, streamingMessage];
```
codeincontext, line 30: As above, in the long run we don't want to do this, but for this PR we could useMemo and put in a comment

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -40,28 +40,23 @@ import {
 const log = aiLogger("chat");
 
 export type ChatContextProps = {
+  // ID can only be removed when teh useChat changes location
```
codeincontext, line 43: Can it be undefined / null?

apps/nextjs/src/stores/chatStore/index.ts
```diff
@@ -87,10 +99,24 @@ export const useChatStore = create<ChatStore>((set, get) => ({
...
+      lessonPlan: null,
+      chatAreaRef: { current: null },
+      input: "",
+      isStreaming: false,
+      isLoading: false,
+      actions: {
+        stop: () => {},
+        reload: () => {},
+        append: async () => "",
+      },
```
codeincontext, line 119: I would think about extracting this to an object, and initialising the store with the same object

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
tomwisecodes: Will be removed shortly so happy to just leave it

PR #519, opened 2025-01-30
chore: remove env from test.yml

mantagen:
## Description

- remove env from test.yml which was unecessary and didn't fix the issue


gclomax (approved):
lgtm

PR #531, opened 2025-02-06
fix: add basedOn to LessonPlanKeySchema

codeincontext:
In #523 I created LessonPlanKeySchema. I didn't include `basedOn`, which causes zod to fail to parse if you select a lesson from rag


JBR90 (approved):
Suprised this didnt get picked up

codeincontext:
> Surprised this didnt get picked up by the e2e tests

Interesting point. Maybe we need a test that uses rag and have a gap in our coverage @mikeritson-oak

PR #533, opened 2025-02-10
feat: aila threat detection

stefl:
## Aila Threat Detection
(WIP for discussion)

- Extends the Aila Threat Detection config so that we can handle different threat detection providers
- Adds an additional step to call out to an external service and throw an error if a threat is detected in the user-provided content
- Adds config in the chat endpoint to configure threat detection to use different providers

Doesn't yet make changes to analytics events and the Prisma schema:
- helicone_threat_detected will need to be renamed
- ENUM for SafetyViolationSource will need to be migrated to include either "THREAT_DETECTOR" or the named service we use.

How it works:

* Have multiple pluggable threat detectors that extend from an abstract class
* Map the input into whatever the threat detector wants (this is why it's "unknown" - we might want plain text, an array of Messages, etc.)
* Map the output of the detection to a general schema we can use for analytics / reporting
* Run the external check before we hit the LLM
* If a threat is detected, throw a threat detection error
* Still keep Helicone as a proxy for now
* Handle Helicone threat detection errors in the same way we do now


packages/aila/src/features/threatDetection/detectors/lakera/schema.ts
```diff
@@ -0,0 +1,55 @@
...
+  content: z.string(),
+});
+
+export const lakeraGuardRequestSchema = z.object({
+  messages: z.array(messageSchema),
+  project_id: z.string().optional(),
+  payload: z.boolean().optional(),
+  breakdown: z.boolean().optional(),
+  metadata: z.record(z.any()).optional(),
+  dev_info: z.boolean().optional(),
```
mantagen, line 14: If Lakera schema changes and we don't know about it then every request breaks, right? I wondrer if we change the `optional` to `nullish`?

mantagen: And for the schemas below?

packages/aila/src/features/threatDetection/detectors/lakera/test-lakera.ts
```diff
@@ -0,0 +1,35 @@
+import { config } from "dotenv";
+import { resolve, dirname } from "path";
+import { fileURLToPath } from "url";
+
+/* eslint-disable no-console */
```
mantagen, line 5: Do we need this eslint-disable?

packages/aila/src/features/threatDetection/detectors/lakera/schema.ts
mantagen: Unless we're confident of their schemas

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -71,6 +74,12 @@ async function setupChatHandler(req: NextRequest) {
         chatId,
       );
 
+      const threatDetectors = [
+        new BasicThreatDetector(),
+        new HeliconeThreatDetector(),
+        new LakeraThreatDetector(),
```
mantagen, line 80: For now shall we just use Lakera? I think Helicone one can go, and maybe not use the Basic one either, so we can trial Lakera in isolation

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
```diff
@@ -0,0 +1,226 @@
...
+import {
+  type ThreatCategory,
+  type ThreatDetectionResult,
+  type ThreatSeverity,
+  AilaThreatDetector,
+} from "../AilaThreatDetector";
+import type { Message, LakeraGuardResponse, BreakdownItem } from "./schema";
+import { lakeraGuardRequestSchema, lakeraGuardResponseSchema } from "./schema";
+
+const log = {
```
mantagen, line 13: AI Logger here with a new namespace?

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
```diff
@@ -0,0 +1,226 @@
...
+    switch (detectorType) {
+      case "jailbreak":
+      case "prompt_injection":
+        return "critical";
+      case "pii":
+      case "sensitive_info":
+        return "high";
+      case "harmful_content":
+        return "high";
+      default:
```
mantagen, line 55: We might want to signal to ourselves here that there's a type we're not covering

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
mantagen: (and elsewhere we log to console)

mantagen (approved):
Approving, code looks good and it's mergeable (pending @mikeritson-oak's QA) - but there are a few actionable comments I've left.

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: No we still need this if the Helicone proxy is on. If it's off then this has no effect

packages/aila/src/features/threatDetection/detectors/lakera/test-lakera.ts
stefl: This is a local file just to run a test of their API

packages/aila/src/features/threatDetection/detectors/lakera/schema.ts
stefl: They version their schema so if it changes it would become v3 in the URL

packages/aila/src/features/threatDetection/helicone/HeliconeThreatDetector.ts
```diff
@@ -0,0 +1,27 @@
+import type { ThreatDetectionResult } from "../detectors/AilaThreatDetector";
+import { AilaThreatDetector } from "../detectors/AilaThreatDetector";
+
+export class HeliconeThreatDetector extends AilaThreatDetector {
```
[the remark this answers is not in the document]
stefl: We do so that we can process Helicone proxy errors if the flag is turned on

packages/aila/src/features/threatDetection/detectors/helicone/HeliconeThreatDetector.ts
```diff
@@ -0,0 +1,27 @@
+import type { ThreatDetectionResult } from "../AilaThreatDetector";
+import { AilaThreatDetector } from "../AilaThreatDetector";
+
+export class HeliconeThreatDetector extends AilaThreatDetector {
```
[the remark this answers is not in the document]
stefl: We still need to keep it in if the flag is turned on

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
stefl: Ah yes. I was having trouble seeing output. Will flip this back to the ai logger

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
stefl: This is a stub really - we have since learned that we get L1-4 classifications from Lakera

apps/nextjs/src/app/api/chat/chatHandler.ts
mantagen: Yep, okay. Do we still need the 'Basic' detector running? Feels superfluous

packages/aila/src/features/threatDetection/detectors/lakera/LakeraThreatDetector.ts
mantagen: Gotcha, so you thinking we can remove them, or are they still useful?

PR #536, opened 2025-02-11
refactor: lesson plan rendering and scrolling

codeincontext:
## Description

- Overhaul rendering of lesson plan sections. Remove local calculation of state like isStreaming, etc
- Overhaul scrolling logic in lesson plan

## Issue(s)

Fixes #

## How to test

This one relates specifically to the right hand side panel:
- The sections show as loading, open, closed, as you'd expect
- It scrolls between sections at the right times
- Clicking on a section name from the dropdown should still scroll to that section


apps/nextjs/src/components/AppComponents/Chat/lesson-plan-section/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const FullWidthButton = styled.button`
+  width: 100%;
+`;
+
+const DropDownSectionWrapper = styled(OakBox)`
+  &:first-child {
+    border-top: 2px solid black;
+  }
+  scroll-margin-top: 180px;
```
codeincontext, line 96: This is key to the scrolling

apps/nextjs/src/components/AppComponents/Chat/chat-lessonPlanDisplay.tsx
```diff
@@ -104,6 +145,27 @@ export const LessonPlanDisplay = ({
...
+      sectionRefs.current[key] = ref;
+    },
+    [sectionRefs],
+  );
+
+  useSectionScrolling({
+    sectionRefs: sectionRefs.current,
+    documentContainerRef,
+    userHasCancelledAutoScroll,
+  });
```
codeincontext, line 167: I think this could move to its own logic with a store and a context at some point

JBR90 (approved):

mikeritson-oak (changes requested):
Tested across browser and three issues/questions:

1. **Rendering** - The rendering/output of the sections appears to have changed from what we have in production. We currently render the section/placeholder and then populate with content, then move to the next. In this version we are rendering the sections/placeholders all at the same time, then cycle through them populating with content. It doesn't actually look 'bad', just different from what we currently have. Is this expected?
2. **Safari** - None of the section links within the sections dropdown function in Safari, checked and passed in all other browsers.
3. **Lesson details** - In the sections dropdown, the first option for 'Lesson details' does not take the user anywhere (I assume this should be the top of the lesson/lesson title). This is currently the same in Production. If it can't be rectified with this work I can raise a separate ticket for it.

codeincontext:
> Tested across browser and three issues/questions:
> 
> 1. **Rendering** - The rendering/output of the sections appears to have changed from what we have in production. We currently render the section/placeholder and then populate with content, then move to the next. In this version we are rendering the sections/placeholders all at the same time, then cycle through them populating with content. It doesn't actually look 'bad', just different from what we currently have. Is this expected?
> 2. **Safari** - None of the section links within the sections dropdown function in Safari, checked and passed in all other browsers.
> 3. **Lesson details** - In the sections dropdown, the first option for 'Lesson details' does not take the user anywhere (I assume this should be the top of the lesson/lesson title). This is currently the same in Production. If it can't be rectified with this work I can raise a separate ticket for it.

Thanks for the detailed review

1. Confirmed in slack that this is the intended behaviour which is currently broken or disabled in production
2. Good spot. I've added a library to make it work in Safari
3. I'll have a look at this now. Thanks

JBR90:
Sections links working in Safari ✅
Lesson details takes you to lesson outcome, should this take the user to key stage /  title ? Checked on production and it seems to be broken there too.

There's a few issues with multiple imports with sonar cloud.

codeincontext:
> Sections links working in Safari ✅ Lesson details takes you to lesson outcome, should this take the user to key stage / title ? Checked on production and it seems to be broken there too.
> 
> There's a few issues with multiple imports with sonar cloud.

Thanks Joe. Have fixed these!

PR #541, opened 2025-02-13
refactor: move input field to chat store

codeincontext:
Just a small one to move the text input field to the chat store


apps/nextjs/src/components/AppComponents/Chat/chat-panel.tsx
```diff
@@ -24,7 +24,9 @@ function LockedPromptForm() {
 
 export function ChatPanel({ isDemoLocked }: Readonly<ChatPanelProps>) {
   const chat = useLessonChat();
-  const { id, messages, input, setInput } = chat;
+  const { id, messages } = chat;
```
JBR90, line 27: Why are we still using id and messages from chat here ?

codeincontext: That's addressed in a previous PR, #540

JBR90 (approved):

PR #542, opened 2025-02-13
refactor: removing lesson plan and tracking events from chatProvider

codeincontext:
## Description

- Moves onStreamFinished tracking into store from provider
- removed useTemporaryLessonPlanWithStreamingEdits(
- overrides lesson plan in store when toxic mod
- removes hashing in provider

## How to test
- streaming analytics should be working as there are in production
- tox moderation should work as before, it now clears lesson plan from state as well.

## Streaming data analytics notes
Lesson refinement - this is firing every time the user clicks continue even if there is no user text, does this count as a refinement?  
"The user alters or sways the Lesson Assistant's Lesson Plan output in some way" - Avo

Bug - a user could enter chat text and click `continue` rather than the text input button, the text would be cleared and not sent with tracking event

Bug - we are not sending tracking for toxic lesson 'lessonPlanTerminated`

Bug - we are not always sending the component with the tracking event, actions in the provider not always working

Might be worth fixing these by adding the analytics provider logic into a store, otherwise separate tickets.

## Checklist


apps/nextjs/src/lib/analytics/lessonPlanTrackingContext.tsx
```diff
@@ -50,10 +50,12 @@ const LessonPlanTrackingProvider: FC<LessonPlanTrackingProviderProps> = ({
   const onStreamFinished = useCallback(
     ({ prevLesson, nextLesson, messages }: OnStreamFinishedProps) => {
       const ailaMessageContent = getLastAssistantMessage(messages)?.content;
+
       if (!ailaMessageContent) {
         return;
       }
-      const isFirstMessage = messages.length === 2;
+      const isFirstMessage = messages.length === 3;
```
JBR90, line 57: Clicking through from /aila there is no assistant message so returns on line 54 
When user clicks continue there 3 messages.

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -192,8 +166,6 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
         });
 
       setHasFinished(true);
-      shouldTrackStreamFinished.current = true;
-
       streamingFinished();
       messageFinished();
```
JBR90, line 170: Can we trigger this in handleChangedAilaStreamingStatus ?

apps/nextjs/src/stores/lessonPlanStore/stateActionFunctions/handleTrackingEvents.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import type { LessonPlanStore } from "..";
+
+export const handleTrackingEvents = (
+  lessonPlanTracking: LessonPlanTrackingContextProps,
+  get: () => LessonPlanStore,
+) => {
+  const { chatActions } = get();
+
+  const messages = chatActions?.getMessages();
```
codeincontext, line 13: I think this is a hint that we should tweak our pattern here. Maybe we need something like `chatStore.getState().stableMessages` or `getChatStore().stableMessages` so that we can access the latest version of the store directly?

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
```diff
@@ -9,9 +9,9 @@ export const handleToxicModeration = (
...
 ) => {
   set({ toxicModeration: mod });
-  const { chatActions } = get();
+  const { chatActions, lessonPlanActions } = get();
   invariant(chatActions, "Passed into store in provider");
+  invariant(lessonPlanActions, "Passed into store in provider");
   chatActions.setMessages([], false);
-
-  //  @TODO setOverrideLessonPlan({});
+  lessonPlanActions.resetStore();
```
codeincontext, line 16: I think we should decide on a pattern for actions between stores. Should the moderation store be telling the lesson plan store what to do (lessonPlanActions.resetStore), or should it tell the store the situation for it to react to (lessonPlanActions.handleToxicModeration)

apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
codeincontext: We chatted about this and will implement a subscription between stores for ailaStreamingStatus in AI-830

apps/nextjs/src/stores/moderationStore/actionFunctions/handleToxicModeration.tsx
codeincontext: AI-830

codeincontext:
This was originally my PR and Joe took it over. you can't approve your own PR so I'll just approve here

**APPROVED**

PR #544, opened 2025-02-13
refactor: move chat message rendering to chat store

codeincontext:
## Description

- Make ChatMessage component use ParsedMessage, as that's what's used in the store
- Render messages based on staticMessages and streamingMessage to reduce unnecesssary renders
- Remove messages from chat provider
- Remove working-on-it conditional logic, which was scattered around the message rendering  

## Issue(s)

Fixes #

## How to test

1. Create a lesson
2. Look at the chat messages
  - Are they showing at the right time?
  - Is the "working on it" message showing?
  - Is the spacing between items correct?
  - Does a moderation show as you'd expect?
  - Does the demo limit message show as you'd expect?
  - Do the in chat download buttons show as expected?


apps/nextjs/src/components/AppComponents/Chat/chat-message/layout.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function MessageContainer({
+  children,
+  roleType,
+  className,
+}: MessageContainerProps) {
+  const testId = `chat-message-wrapper-${roleType}`;
+  return (
+    <div
```
JBR90, line 29: Can we make net new component with Oak Components ?

JBR90 (approved):

apps/nextjs/src/components/AppComponents/Chat/chat-message/layout.tsx
codeincontext: Oh I like that! Good proposal for a follow up PR? I think a lot of the classes in this component aren't actually needed any more so maybe we can just give it a re-do

codeincontext: Added as AI-829

PR #550, opened 2025-02-18
fix: getChat overfetching - move initialMessages to store

codeincontext:
At the moment the chat page invalidates and refetches the getChat tRPC hook fairly often. We only have two things left that actually make use of it:
- Setting initialMessages on the AI SDK useChat
- Passing lessonPlan in the body of AI SDK requests

This PR moves both of those behaviours over to our stores so that we can remove the getChat tRPC hook from the provider

## Issue(s)

Fixes #

## How to test

1. Go to {deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

### DIFF
--- a/apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
+++ b/apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
@@ -1,22 +1,19 @@
 import type React from "react";
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 import { toast } from "react-hot-toast";
 
 import { generateMessageId } from "@oakai/aila/src/helpers/chat/generateMessageId";
 import { parseMessageParts } from "@oakai/aila/src/protocol/jsonPatchProtocol";
-import type { LooseLessonPlan } from "@oakai/aila/src/protocol/schema";
 import { aiLogger } from "@oakai/logger";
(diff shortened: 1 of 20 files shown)


apps/nextjs/src/components/ContextProviders/ChatProvider.tsx
```diff
@@ -107,34 +60,28 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
...
     sendExtraMessageFields: true,
+    // TODO: these initial messages are sent and used by the chat endpoint
     initialMessages,
     generateId: () => generateMessageId({ role: "user" }),
     id,
     body: {
       id,
-      lessonPlan: chat?.lessonPlan,
+      // TODO: this lesson plan is sent and used by the chat endpoint
+      lessonPlan,
```
codeincontext, line 78: Potential race condition here if setting the initial append before lessonPlan has been fetched. But if that's the case then the lessonPlan should be empty anyway

codeincontext: It looks like that was already the case with no problems though

tomwisecodes (approved):
Not fully across all of this work, but from a comb through the code nothing stands out so LGTM.

PR #554, opened 2025-02-19
refactor: simplify inter-store actions 

codeincontext:
Sometimes one store needs to call an action on another store. At the moment we add a field like `moderationStoreActions` on the store and set its value once all of the stores are set up. It feels a little clunky to me and I don't think that we need those store functions remembered in each store

We also sometimes need to get the values of another store. We recently added a `getMessages()` function to the chat store state, which feels a bit clunky given you can already call `chatStore.getState().stableMessages`

This new approach passes a "getStore" object to the constructor of each store instead

### DIFF
--- a/apps/nextjs/src/app/aila/[id]/page.tsx
+++ b/apps/nextjs/src/app/aila/[id]/page.tsx
@@ -1,7 +1,5 @@
 "use client";
 
-import { useUser } from "@clerk/nextjs";
-
 import ChatPageContents from "../page-contents";
 
 interface ChatPageProps {
@@ -11,7 +9,6 @@ interface ChatPageProps {
 }
 
 export default function ChatPage({ params }: Readonly<ChatPageProps>) {
-  const user = useUser();
   const { id } = params;
   // For local development so that we can warm up the server
   if (id === "health") {
--- a/apps/nextjs/src/app/aila/page-contents.tsx
+++ b/apps/nextjs/src/app/aila/page-contents.tsx
@@ -2,10 +2,10 @@
 
 import React from "react";
 
+import { AiSdk } from "@/components/AppComponents/Chat/AiSdk";
 import { Chat } from "@/components/AppComponents/Chat/Chat/chat";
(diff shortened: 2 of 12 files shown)


apps/nextjs/.storybook/decorators/StoreDecorator.tsx
```diff
@@ -26,24 +29,29 @@ export const StoreDecorator: Decorator = (Story, { parameters }) => {
...
     const id = "123";
     const trpcUtils = {} as TrpcUtils;
-    const moderationStore = createModerationStore({
+    const stores: Partial<AilaStores> = {};
+
+    const getStore = <T extends keyof AilaStores>(storeName: T) => {
+      invariant(stores[storeName], `Store ${storeName} not initialised`);
+      return stores[storeName].getState() as ExtractState<AilaStores[T]>;
+    };
+
```
tomwisecodes, line 38: nice

apps/nextjs/src/stores/AilaStoresProvider.tsx
```diff
@@ -35,25 +39,27 @@ export const AilaStoresProvider: React.FC<AilaStoresProviderProps> = ({
   const lessonPlanTracking = useLessonPlanTracking();
 
   const [stores] = useState(() => {
-    const moderationStore = createModerationStore({
+    const stores: Partial<AilaStores> = {};
+
+    const getStore = <T extends keyof AilaStores>(storeName: T) => {
+      invariant(stores[storeName], `Store ${storeName} not initialised`);
+      return stores[storeName].getState() as ExtractState<AilaStores[T]>;
+    };
```
tomwisecodes, line 47: Does this getStore need to be repeated?

codeincontext: Oh interesting. I'll have a look

tomwisecodes (approved):
One small q, but otherwise looks good

apps/nextjs/src/stores/AilaStoresProvider.tsx
codeincontext: Pushing a reusable version

PR #555, opened 2025-02-19
feat: send events tohubspot

tomwisecodes:
## Description

- Adds hubspot to avobridge function. This sends all events to hubspot
- Removes trackEvent from the leagacy quiz designer code without breaking it.
- Removes trackEvent from aila and the analytics provider


apps/nextjs/src/components/ContextProviders/AnalyticsProvider.tsx
```diff
@@ -185,8 +185,8 @@ export const AnalyticsProvider: React.FC<AnalyticsProviderProps> = ({
   initAvo(
     { env: getAvoEnv(), ...avoOptions },
     {},
-    getAvoBridge({ posthog: posthogOak }),
-    getAvoBridge({ posthog: posthogAiBeta }),
+    getAvoBridge({ posthog: posthogOak, hubspot }),
+    getAvoBridge({ posthog: posthogAiBeta, hubspot }),
```
codeincontext, line 189: I think this will double report to hubspot?

codeincontext: I think you need something closer to

```suggestion
    getAvoBridge({ posthog: posthogOak }),
    getAvoBridge({ posthog: posthogAiBeta }),
    getAvoBridge({ hubspot }),
```

tomwisecodes: Good point, I had a go like that but Avo init is not expecting a bridge. We could change that in Avo but it felt right to keep the analytics matching exactly what goes into the aila hubspot

codeincontext: Ok yeah, this looks fine now

apps/nextjs/src/components/AppComponents/Chat/open-side-bar-button.tsx
```diff
@@ -4,23 +4,19 @@ import { OakFlex, OakSpan } from "@oaknational/oak-components";
...
-  const { trackEvent } = useAnalytics();
-
   return (
     <Button
       variant="ghost"
       data-testid="sidebar-button"
       className="flex items-center px-5"
       onClick={() => {
         setOpenSidebar(true);
-        trackEvent("chat:toggle_sidebar");
```
tomwisecodes, line 23: This is not an event tracked in avo so can delete

PR #556, opened 2025-02-19
refactor: remove ChatProvider from layout

codeincontext:
Now that we don't have a chat context, we can move the AI SDK to a leaf component rather than having children. This might reduce re-renders while streaming


apps/nextjs/src/app/aila/page-contents.tsx
```diff
@@ -17,9 +17,8 @@ const ChatPageContents = ({ id }: { readonly id: string }) => {
     <Layout>
       <LessonPlanTrackingProvider chatId={id}>
         <AilaStoresProvider id={id}>
-          <ChatProvider id={id}>
-            <Chat />
-          </ChatProvider>
+          <AiSdk id={id} />
+          <Chat />
```
tomwisecodes, line 21: huge moment

apps/nextjs/src/components/AppComponents/Chat/AiSdk.tsx
```diff
@@ -82,12 +57,6 @@ export function ChatProvider({ id, children }: Readonly<ChatProviderProps>) {
   const { invokeActionMessages } = useActionMessages();
 
   /******************* Streaming of all chat starts from messages here *******************/
```
tomwisecodes, line 59: Now this is all alot cleaner and digestiable shall we get rid of any of these comments

codeincontext: Yeah, good shout

tomwisecodes (approved):

apps/nextjs/src/app/aila/page-contents.tsx
codeincontext: I know right!

PR #559, opened 2025-02-20
chore: move remaining lint warnings to errors

codeincontext:
Stef and Joe did some great work with our eslint rules, but some rules were hard to fix while we had the moving target of work in other branches. This PR brings the remaining warning rules into the errors list

### DIFF
--- a/packages/ingest/src/captions/getCaptionsByFileName.ts
+++ b/packages/ingest/src/captions/getCaptionsByFileName.ts
@@ -4,7 +4,7 @@ import type { Cue } from "webvtt-parser";
 import * as WebVTTModule from "webvtt-parser";
 
 const WebVTTParser =
-  // @ts-expect-error import issue: TO doesn't think .default exists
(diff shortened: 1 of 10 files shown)


packages/ingest/src/captions/getCaptionsByFileName.ts
```diff
@@ -3,10 +3,10 @@ import { aiLogger } from "@oakai/logger";
 import type { Cue } from "webvtt-parser";
 import * as WebVTTModule from "webvtt-parser";
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-expect-error
-const WebVTTParser = WebVTTModule.default
-  .WebVTTParser as typeof WebVTTModule.WebVTTParser;
+const WebVTTParser =
+  // @ts-expect-error import issue: TO doesn't think .default exists
```
codeincontext, line 7:
```suggestion
  // @ts-expect-error import issue: TS doesn't think .default exists
```

PR #566, opened 2025-02-25
feat: start migrating to have an AilaDocument

stefl:
## Description

- Starts migrating away from the Aila package being specifically tied to the concept of a lesson plan
- Fixes some small linting issues
- Turns off the experimental quiz test suite selectively with an env var, because you need to have records in the local database and there doesn't seem to be a script for inserting them
- Should be no user-facing changes


apps/nextjs/src/app/api/chat/webActionsPlugin.ts
```diff
@@ -28,10 +28,12 @@ export const createWebActionsPlugin: PluginCreator = (
       // #TODO change this to handleThreatDetectionError and move
       // the logic elsewhere. Stop passing Prisma
       const threatError = await handleThreatDetectionError(
-        aila.userId ?? "anonymous", // This should never be "anonymous" because we would get an authentication error
-        aila.chatId ?? "unknown",
-        error,
-        prisma,
+        {
```
stefl, line 31: This was complaining about argument complexity

packages/aila/src/core/Aila.ts
```diff
@@ -48,15 +47,14 @@ export class Aila implements AilaServices {
   private readonly _chat: AilaChatService;
   private readonly _errorReporter?: AilaErrorReportingFeature;
   private _isShutdown: boolean = false;
-  private readonly _lesson: AilaLessonService;
+  private readonly _document: AilaDocumentService;
```
stefl, line 50: Rather than _lesson, we now have _document

packages/aila/src/core/Aila.ts
```diff
@@ -242,7 +239,7 @@ export class Aila implements AilaServices {
   }
 
   public get rag() {
-    return this._rag;
+    throw new Error("Attempting to use old, unused RAG service");
```
stefl, line 242: The RAG service seems to be unused! So, removing it so that it's clearer what is doing what. RAG seems to all happen within the prompt builder now

packages/aila/src/core/quiz/LessonSlugQuizMapping.test.ts
```diff
@@ -1,170 +1,178 @@
 import { ElasticLessonQuizLookup } from "./LessonSlugQuizMapping";
 
-describe("ElasticLessonQuizLookup", () => {
-  let dbLookup: ElasticLessonQuizLookup;
+if (process.env.TEST_QUIZZES === "true") {
```
stefl, line 3: Selectively turns off these quiz tests until we have a way of inserting test data into the database

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -160,7 +159,7 @@ export class AilaStreamHandler {
     }
     log.info("Starting to read from stream");
     try {
-      while (true) {
+      while (!abortController?.signal.aborted) {
```
codeincontext, line 162: Nice

packages/aila/src/core/quiz/LessonSlugQuizMapping.test.ts
codeincontext: I would take the `!== "false"` approach from below so that you have to opt into this behaviour. Otherwise we risk having the tests disabled on staging

```suggestion
if (process.env.TEST_QUIZZES !== "false") {
```

Another approach could be to add a new pnpm command like `pnpm test:no-quizzes` for local dev

codeincontext: I'm also trying to think of a way to avoid this large diff. Have you tried:

```
const maybeDescribe = process.env.TEST_QUIZZES === "false" ? describe.skip : describe
```

codeincontext (commented):
This looks great. I've got a suggestion of optionally skipping the experimental tests rather than wrapping them in an if statement and removing them

The alternative is to ask @gclomax about a minimal seed file we can use for the tests

packages/aila/src/core/quiz/LessonSlugQuizMapping.test.ts
stefl: Ah that's a good idea. Thanks!

packages/aila/src/core/chat/AilaStreamHandler.ts
stefl: Actually, I realised the controller is optional!

packages/aila/src/core/quiz/LessonSlugQuizMapping.test.ts
stefl: I've managed to find a way to make that work in a similar way to what you've suggested, thanks

codeincontext (approved):

packages/aila/src/core/chat/AilaStreamHandler.ts
codeincontext: Yeah, that was a good spot!

PR #569, opened 2025-02-28
feat: only trust the lesson plan from the server side

stefl:
## Description

- Stops sending the current state of the lesson plan from the front end
- The server now saves and loads the lesson plan state and uses what is in the database as the source of truth
- This closes a potential opportunity for someone to construct a bad lesson plan by manipulating HTTP requests.
- Means that when a threat detection job is run, and it ignores the system prompt, we can trust the inputs to the system prompt because they come from the database

No user-facing change expected, but we should make sure that we don't see any issues where what's on-screen for the user flashes, reverts or disappears. This might indicate a mistake in the logic flow.


package.json
```diff
@@ -24,7 +24,8 @@
     "doppler:pull:stg": "doppler secrets download --config stg --no-file --format env > .env",
     "doppler:run:stg": "doppler run -c stg --silent",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
-    "lint": "turbo lint",
+    "lint": "turbo lint --pass-with-no-tests",
+    "lint:file": "eslint",
```
stefl, line 28: Add a way to run eslint on a particular file

codeincontext (commented):
Looking good. I appreciate that this is such a self-contained PR while I'm looking at other areas of the streaming

codeincontext (approved):
Approving on the assumption that the e2e tests work as expected

PR #570, opened 2025-02-28
feat: do not trust front end message history

stefl:
## Description

- Addresses a security / threat detection issue where we are currently trusting that the message history that the front end sends has not been manipulated
- Instead, we now load the message history from the database and append the user's most recent message
- We then use that as the input to the LLM, and this is what we send to the threat detection services
- In combination with the preceding PR, we can guarantee that the only new input that is sent to the LLM is the latest user message

To test:
* Start a chat
* Add a message using the left hand side
* Try repeatedly adding messages, leaving and reloading the chat to make sure that the state of the left and right hand sides of the UI remain consistent


apps/nextjs/src/components/AppComponents/Chat/AiSdk.tsx
```diff
@@ -69,8 +69,6 @@ export function AiSdk({ id }: Readonly<AiSdkProps>) {
     id,
     body: {
       id,
-      // NOTE: this lesson plan is used by the chat endpoint
-      lessonPlan,
```
codeincontext, line 73: Can we delete line 49 above now? `const lessonPlan = useLessonPlanStore((state) => state.lessonPlan);`

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -167,59 +171,227 @@ async function generateChatStream(
...
+    }
+
+    if (
+      hasLessonPlan(parsedOutput) &&
+      isValidLessonPlan(parsedOutput.lessonPlan)
+    ) {
+      lessonPlan = parsedOutput.lessonPlan as LooseLessonPlan;
+    }
+  } catch (error) {
+    log.error(`Error parsing output for chat ${chatId}`, error);
```
codeincontext, line 225: I just want to confirm that we don't want to report to sentry here

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -167,59 +171,227 @@ async function generateChatStream(
...
+  }
+
+  for (let i = frontendMessages.length - 1; i >= 0; i--) {
+    const message = frontendMessages[i];
+    if (message && message.role === "user") {
+      return message;
+    }
+  }
+
+  return null;
```
codeincontext, line 276: I think this can be tidied with [findLast](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast)

```suggestion
  return (frontendMessages || []).findLast(m => m.message?.role === "user")
```

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -167,59 +171,227 @@ async function generateChatStream(
...
+}: {
+  config: Config;
+  options: AilaOptions;
+  chatId: string;
+  userId: string | undefined;
+  messages: Message[];
+  lessonPlan: LooseLessonPlan;
+  llmService: ReturnType<typeof getFixtureLLMService>;
+  moderationAiClient: ReturnType<typeof getFixtureModerationOpenAiClient>;
+  threatDetectors: AilaThreatDetector[];
```
codeincontext, line 317: I would argue that this type is too large to be inline, though that's a style preference so a nitpick rather than a blocker!

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -167,59 +171,227 @@ async function generateChatStream(
...
+        config,
+        options,
+        chatId,
+        userId,
+        messages,
+        lessonPlan: dbLessonPlan,
+        llmService,
+        moderationAiClient,
+        threatDetectors,
+      });
```
codeincontext, line 393: Much nicer!

codeincontext (commented):
This is looking really good. I'll approve after you've seen the comments, but each is minor

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: Thanks!

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: Agreed - made that a separate type. TBH this file is getting pretty big and should probably be broken up

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: Cheers!

apps/nextjs/src/components/AppComponents/Chat/AiSdk.tsx
stefl: Indeed we can!

apps/nextjs/src/app/api/chat/chatHandler.ts
stefl: Thanks - I've added Sentry reporting. For now, no additional error handling for the front end though

codeincontext (approved):

PR #571, opened 2025-02-28
feat: further migration steps towards AilaDocument

stefl:
## Description

- This goes further with the process of getting Aila to support the generation of different types of document
- Introduces an AilaDocumentContent type, which is a compound type of our existing lesson plan and a dummy type that is currently unused
- Makes some methods generic (snapshots, americanisms) so that they can handle different types of document

No user-facing changes. No data migrations.

### DIFF
--- a/_migratingLessonsToDocuments.txt
+++ b/_migratingLessonsToDocuments.txt
@@ -1,8 +1,0 @@
-To do list for migrating the Aila package from just handling lessons, to allow it it handle different types of document.
-
-## Snapshots
-
-We need to generate a JSON schema for each document type.
-
-## Schemas for documents
-The document schemas should be in the Aila package, and move out of protocol.


packages/aila/src/core/Aila.liveWithOpenAI.test.ts
```diff
@@ -13,7 +13,7 @@ const runManually = process.env.RUN_LLM_TESTS === "true";
 
     beforeEach(() => {
       ailaInstance = new Aila({
-        lessonPlan: {},
+        document: { content: {} },
```
stefl, line 16: My thinking here is that we will probably want to pass a schema, a version, some metadata. So the document content probably needs to have this extra level of abstraction.

packages/aila/src/core/document/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { LooseLessonPlan } from "../../protocol/schema";
+
+export type AilaDummyDocumentContent = {
```
stefl, line 3: This is an example of a different type of document that Aila could produce.

I think we will probably want a discriminating "type" attribute so we can use Zod to do a discriminated union. Not implemented yet because that would change the shape of the LooseLessonPlan.

packages/aila/src/features/americanisms/AilaAmericanisms.ts
```diff
@@ -8,17 +8,20 @@ import type {
...
 
 export type TranslationResult = Record<
   string,
   Array<{ [phrase: string]: { issue: string; details: string } }>
 >;
 
-export class AilaAmericanisms implements AilaAmericanismsFeature {
+export type AilaDocumentContent = Record<string, unknown>;
+
+export class AilaAmericanisms<T extends AilaDocumentContent>
```
stefl, line 19: This removes the association with the lesson plan and treats it as an unknown object.

packages/aila/src/features/types.ts
```diff
@@ -12,7 +13,7 @@ import type { AilaThreatDetector } from "./threatDetection";
 export interface AilaModerationFeature {
   moderate(options: {
     messages: Message[];
-    lessonPlan: LooseLessonPlan;
+    content: AilaDocumentContent;
```
stefl, line 16: The moderation feature now just accepts some arbitrary content.

_migratingLessonsToDocuments.txt
```diff
@@ -0,0 +1,8 @@
+To do list for migrating the Aila package from just handling lessons, to allow it it handle different types of document.
+
+## Snapshots
+
+We need to generate a JSON schema for each document type.
+
+## Schemas for documents
+The document schemas should be in the Aila package, and move out of protocol.
```
codeincontext, line 8: Should we move this to notion? Maybe it belongs in an epic or an Aila Engineering Knowledge age

packages/aila/src/core/document/AilaDocument.ts
```diff
@@ -11,73 +11,70 @@ import {
...
   }
 
-  public get hasSetInitialState(): boolean {
-    return this._hasSetInitialState;
-  }
-
-  public set hasSetInitialState(value: boolean) {
-    this._hasSetInitialState = value;
+  public get hasInitialisedContentFromMessages(): boolean {
+    return this._hasInitialisedContentFromMessages;
```
codeincontext, line 58: I've always wondered if these could be defined as setters and getters. But changing something like that would be out of scope for a rename like this so let's ignore this comment!

codeincontext (approved):
Great work

_migratingLessonsToDocuments.txt
stefl: Ah yes - I spotted this had slipped in. It was a temporary file I thought I'd ignored!

packages/aila/src/core/document/AilaDocument.ts
stefl: I agree on this - we seem to have setPlan and also set plan. Probably quite easy to remove the setPlan method!

PR #575, opened 2025-03-03
feat: continue migrating Aila package to support variable documents

stefl:
## Description

- This continues the process of moving to AilaDocument instead of AilaLessonPlan
- Still no migrations - just changing some variable and class names

### DIFF
--- a/packages/aila/src/core/Aila.ts
+++ b/packages/aila/src/core/Aila.ts
@@ -141,7 +141,7 @@ export class Aila implements AilaServices {
     await this.loadChatIfPersisting();
     const persistedLessonPlan = this._chat.persistedChat?.lessonPlan;
     if (persistedLessonPlan) {
-      this._document.setContent(persistedLessonPlan);
+      this._document.content = persistedLessonPlan;
     }
     await this._document.initialiseContentFromMessages(this._chat.messages);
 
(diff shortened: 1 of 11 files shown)


packages/aila/src/core/Aila.ts
```diff
@@ -141,21 +141,21 @@ export class Aila implements AilaServices {
...
   }
 
-  private initialiseOptions(options?: AilaOptions) {
+  private initialiseOptions(
+    options?: AilaOptions,
+  ): AilaOptionsWithDefaultFallbackValues {
     return {
       useRag: options?.useRag ?? true,
       temperature: options?.temperature ?? DEFAULT_TEMPERATURE,
-      // #TODO we should find a way to make this less specifically tied
```
stefl, line 155: Addresses this TODO

packages/aila/src/core/document/AilaDocument.ts
```diff
@@ -50,10 +50,6 @@ export class AilaDocument implements AilaDocumentService {
     this._content = content;
   }
 
-  public setContent(content: AilaDocumentContent) {
```
stefl, line 53: Removes this method in preference of `set` above

packages/aila/src/core/document/types.ts
```diff
@@ -6,6 +6,10 @@ export type AilaDummyDocumentContent = {
   keyStage: string;
   topic?: string;
   body: string;
+  basedOn?: {
```
stefl, line 9: For now, all document types would need a basedOn attribute

codeincontext (approved):

PR #576, opened 2025-03-04
feat: aila instantiation includes option to specify a schema and categorisation approach

stefl:
## Description

- Continues on the move from Aila being specific to a lesson plan to supporting any kind of document
- Allows us to pass in a schema that Aila should work with
- Allows us to pass in an optional categoriser which would set up the document with initial values based on the user's first message(s)
- Includes one categoriser plugin for Lesson Plans, which would not be used if not specified when instantiating the Aila instance. TBH I think we should be removing the categoriser anyway, but this keeps it in for now.


packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -218,26 +237,66 @@ export class AilaChat implements AilaChatService {
     return applicableMessages;
   }
 
-  // #TODO this is the other part of the initial lesson state setting logic
```
stefl, line 221: Addresses this TODO about making the initial categorisation step something pluggable / specific to the type of document we are trying to create

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -287,9 +289,13 @@ export async function handleChatPostRequest(
               ],
               threatDetectors: () => threatDetectors,
             },
-
             document: {
               content: dbLessonPlan ?? {},
+              schema: LessonPlanSchema,
```
stefl, line 294: This is the main starting point. We want to pass in a schema that we want to work with.

apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -287,9 +289,13 @@ export async function handleChatPostRequest(
...
+              categorisationPlugin: (aila: AilaServices) =>
```
stefl, line 295: And an optional categorisation plugin

packages/aila/src/core/Aila.liveWithOpenAI.test.ts
```diff
@@ -13,7 +15,21 @@ const runManually = process.env.RUN_LLM_TESTS === "true";
 
     beforeEach(() => {
       ailaInstance = new Aila({
-        document: { content: {} },
+        document: {
+          content: {},
+          schema: LessonPlanSchema,
```
stefl, line 20: This is where we say what kind of document we want the Aila instance to deal with.

packages/aila/src/core/Aila.test.ts
```diff
@@ -115,36 +118,48 @@ describe("Aila", () => {
...
       await ailaInstance.initialise();
 
-      expect(mockCategoriser.categorise).toHaveBeenCalledTimes(1);
+      ailaInstance.document.content = {
+        ...ailaInstance.document.content,
+        title: "Roman Britain",
+        subject: "history",
+        keyStage: "key-stage-2",
+        topic: "The Roman Empire",
+      };
```
codeincontext, line 135: It would be useful to have a comment explaining this, as it seems to be overwriting the result of the test

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -218,26 +227,80 @@ export class AilaChat implements AilaChatService {
...
+          }
         }
       }
     }
   }
 
+  /**
+   * Ensures values are safe to use in enqueuePatch using Zod validation
+   * with a maximum recursion depth to prevent stack overflow
+   */
```
codeincontext, line 254: This is starting to feel too specific to live in the AilaChat module. Do we need a module relating to initial state?

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -218,26 +227,80 @@ export class AilaChat implements AilaChatService {
...
+
+        for (const [key, propValue] of Object.entries(value)) {
+          const safeValue = this.ensureSafeValue(propValue, depth + 1);
+          if (safeValue !== undefined) {
+            result[key] = safeValue;
+            hasValidProperties = true;
+          }
+        }
+
+        return hasValidProperties ? result : undefined;
```
codeincontext, line 292: I think you can drop hasValidProperties by using `Object.keys(result).length ? result : undefined`

That also opens up more functional options to build the object like a map and filter - though that's a code style thing

packages/aila/src/core/chat/AilaChat.ts
```diff
@@ -260,7 +323,10 @@ export class AilaChat implements AilaChatService {
     if (!warning) {
       return;
     }
-    await this.enqueue({ type: "prompt", message: warning });
+    // Optional "?" Necessary to avoid a "terminated" error
```
codeincontext, line 326: This is new to me. Are you saying that `this` can be undefined?

packages/aila/src/core/document/AilaDocument.test.ts
```diff
@@ -0,0 +1,274 @@
+import { aiLogger } from "@oakai/logger";
+import type { z } from "zod";
+
+import type { Message } from "../chat";
+import { LessonPlanSchema } from "./schemas/lessonPlan";
+import type { AilaDocumentContent, CategorisationPlugin } from "./types";
+
+const log = aiLogger("aila");
+
+class TestDocument {
```
codeincontext, line 10: This file could do with a comment explaining what it does, as as far as I can tell it's defining a local class which isn't used in the codebase (yet?) and then testing it

packages/aila/src/core/document/AilaDocument.ts
```diff
@@ -1,150 +1,327 @@
...
+    log.info("registerCategorisationPlugin called", {
+      pluginId: plugin.id,
+    });
+    this._categorisationPlugins.push(plugin);
+  }
+
+  /**
+   * Get the appropriate plugin for the content
+   * Since we now only register one plugin, we just return the first one
+   */
```
codeincontext, line 83: If we're only using one plugin, and it only works if there is one plugin, should we just support one plugin?

A bit of an embarrassing question, but what actually is a plugin? I can't find where it's being applied in the calling code

packages/aila/src/core/chat/AilaChat.ts
stefl: Yes I was surprised too. The instance detaches when it's removed, and so this no longer exists, or something like that? This is just reinserting a comment that should have been there to match the others. No code change

codeincontext:
Closing for now as it's going to start conflicting with other ongoing changes such as the pseudo-agentic approach

PR #584, opened 2025-03-10
feat: record THREAT in the database rather than HELICONE

stefl:
## Description

- This attributes detected threats as THREAT the safety violation source column in the DB, rather than HELICONE
- Updates the name of the analytics event to be threat_detected instead of helicone_threat_detected
- No user-facing changes
- Migration is required

### DIFF
--- a/packages/aila/src/utils/moderation/moderationErrorHandling.ts
+++ b/packages/aila/src/utils/moderation/moderationErrorHandling.ts
@@ -24,7 +24,7 @@ export async function handleThreatDetectionError(
   SafetyViolations = defaultSafetyViolations,
 ): Promise<ErrorDocument | ActionDocument> {
   await safelyReportAnalyticsEvent({
-    eventName: "helicone_threat_detected",
+    eventName: "threat_detected",
     userId,
     payload: {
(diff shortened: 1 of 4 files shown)


packages/aila/src/utils/moderation/moderationErrorHandling.ts
```diff
@@ -24,7 +24,7 @@ export async function handleThreatDetectionError(
   SafetyViolations = defaultSafetyViolations,
 ): Promise<ErrorDocument | ActionDocument> {
   await safelyReportAnalyticsEvent({
-    eventName: "helicone_threat_detected",
```
codeincontext, line 27: Do we need to make any changes in AVO to change this name?

stefl: I don't think so because we are sending events directly via Posthog, and not using the Avo client. If we were using the Avo client we would get a typescript error at this point. But worth checking what we want to do here.

codeincontext: 👍 understood

PR #586, opened 2025-03-10
feat: allow lifted rate limits for certain users

codeincontext:
Sometimes a user has a valid usage for some spiky usage which takes them over the rate limit, for example when preparing a batch of lessons to go on leave. This is an early version of allowing individual users to have higher limits in exceptional cases

## Description

- Check user metadata for `hasHigherLimits` when in UserBasedRateLimiter

## Issue(s)

Fixes #

## How to test

- Exhaust a user's rate limit (ask someone with doppler access to drop `RATELIMIT_GENERATIONS_PER_24H` to a number like 5
- Set the **hasHigherLimits** to `true` in the user's privateMetadata in clerk
- The user shouldn't be rate limited

### DIFF
--- a/packages/core/src/utils/rateLimiting/userBasedRateLimiter.ts
+++ b/packages/core/src/utils/rateLimiting/userBasedRateLimiter.ts
@@ -54,8 +54,9 @@ export const userBasedRateLimiter = (rateLimit: Ratelimit): RateLimiter => {
         );
       }
 
-      if (await isLimitFreeOakUser(userId)) {
-        log.info("Bypassing rate-limit for oak user %s", userId);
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/rateLimiting/userBasedRateLimiter.ts
```diff
@@ -104,3 +113,10 @@ function userHasOakEmail(user: User) {
       !email.emailAddress.includes("demo"),
```
stefl, line 113: Doesn't this mean that if I have `someone+demo@example.com` I will bypass the rate limiting?

stefl (approved):

PR #587, opened 2025-03-10
chore: update Octokit dependencies (endpoint 9.0.6, plugin-paginate-rest 9.2.2, request 8.4.1)

stefl:
## Description

- Updates some dependencies that @actions/github relies on
- Replaces three Dependabot PRs updating these packages


codeincontext (approved):

codeincontext (changes requested):
Sorry for un-approving. I think these should allow version bumps rather than being fixed versions

codeincontext (approved):
Ok, I went on a bit of exploration here. I think that we want to bump the version used in the lockfile, and we don't really want to limit our sub-dependencies to to a certain patch like you would get with override. Here's what I tried:

- `pnpm update @octokit/endpoint@9.0.6` - no luck with this, it didn't update it at all
- Taking the exact steps of your PR, and then just removing the `overrides` block from the package.json. This did bump the version being used in the lockfile without actually overriding future versions

Is that better? I'm not 100% sure. But at least we know for the future that we can remove these keys with no actual consequence to the codebase

PR #590, opened 2025-03-10
chore: lint and prettier for package.json files

stefl:
## Description

- Adds eslint config for package.json files
- Ensures all packages have MIT as the license type, matching the LICENSE file
- Adds Prettier config for package.json files
- Fixes inconsistencies in all package.json files
- Adds lint:fix command
- Fixes broken pnpm lint-fix in the nextjs package
- Adds a turbo rule for lint-fix
- No user-facing changes

We now get visual feedback if package.json is badly structured:
CleanShot 2025-03-10 at 17 11 15@2x
CleanShot 2025-03-10 at 17 10 43@2x


apps/nextjs/package.json
```diff
@@ -2,33 +2,40 @@
   "name": "@oakai/nextjs",
   "version": "1.0.0",
   "private": true,
+  "license": "MIT",
   "scripts": {
+    "aila": "tsx scripts/aila-cli.ts",
     "build": "next build",
+    "build-storybook": "dotenv -e ../../.env -- storybook build",
```
codeincontext, line 9: I wonder if this highlights that `storybook:build` would be a better convention?

codeincontext (approved):

apps/nextjs/package.json
stefl: Yes I think we want more consistency

PR #599, opened 2025-03-11
chore: allow _ prefix for unused variables

stefl:
## Description

- This adds an eslint rule to allow _ as a prefix for unused variables in various situations
- This can be helpful while developing a feature


codeincontext (approved):
This works for me 👍

PR #603, opened 2025-03-11
feat: add user metadata-based rate limiting

codeincontext:
## Description

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist


packages/core/src/utils/rateLimiting/slidingWindowRateLimiter.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Ratelimit } from "@upstash/ratelimit";
+import { waitUntil } from "@vercel/functions";
+import { kv } from "@vercel/kv";
+
+import { RateLimitExceededError } from "./errors";
+import type { Duration, RateLimiter } from "./types";
+
+const log = aiLogger("rate-limiting");
+
+type SlidingWindowRateLimiterArgs = {
```
stefl, line 11: Style question - I normally do export interface. Should we decide on a preferred pattern?

packages/core/src/utils/rateLimiting/types.ts
```diff
@@ -0,0 +1,16 @@
+type Unit = "ms" | "s" | "m" | "h" | "d";
+export type Duration = `${number} ${Unit}` | `${number}${Unit}`;
```
stefl, line 2: Is there any danger in this being called simply Duration? Eg. for autocomplete / auto-import?

stefl: Eg. RateLimitDuration, RateLimitUnit

stefl (approved):
Nice! Tiny style questions - feel free to ignore or discuss elsewhere

packages/core/src/utils/rateLimiting/slidingWindowRateLimiter.ts
codeincontext: I'll update this. I don't know the pros and cons of each approach, but it sounds sensible to me

packages/core/src/utils/rateLimiting/types.ts
codeincontext: I don't know about danger, but I agree it could be annoying. I've renamed it to RateLimitDuration

codeincontext:
From QA:

> I've tried this for a normal account, increased and decreased accordingly and then tested on my Oak account as well, both with and without a restriction. Nothing unexpected seen.

I've also tested demo accounts

PR #617, opened 2025-03-13
revert: revert prompt schema alignment

codeincontext:
We've seen an increase in issues in the LLM response. One example is that it's referring to sections by their camelcase key names. This could relate to us adding those key names in a JSON schema change


mikeritson-oak (approved):
Checked in the preview branch with various test lessons, no issues seen.

PR #619, opened 2025-03-18
fix: add lesson plan tracking store

codeincontext:
## Description

Our lesson plan tracking has become unreliable and we have a number of holes in our analytics. The problem relates to two things:

### The lesson plan tracking provider is fiddly and is missing some features

For example:
- It doesn't know about queueing, so tracking queued messages is undefined
- The logic to determine the initial message is broken

It's a great candidate for a zustand store

### We've been silently returning in a case that should be exceptional and shout loudly

Here's how it works:

- User clicks a button or types a message => We save their UI intention/behaviour in a useState
- Streaming finishes => We send the "Lesson Plan Refined" event with both the result of the streamed data and the user intention in the UI

The problem is that through new features and refactors, a lot of the intent recording had been dropped. When streaming finished, a clause like `if (!userIntention) { return }` was preventing the event from firing at all

## List of changes

- Create LessonPlanTrackingStore
- Call intents like `clickedContinue` on each user interaction that starts streaming
- Handle queued intents if the user submits while streaming is in "moderating" state
- Store previous lesson plan in LessonPlanTrackingStore to compare changes before and after streaming
- Add tracking for the modify button (the additional materials also come under this)
- Add tracking for the initial message
- Fix tracking for the first refinement message
- Unit tests for reach type of message or property
- Report to sentry (or later throw) if there isn't an intent when streaming finishes

## Issue(s)

Fixes #AI-888

## How to test

- Go to a lesson URL and add `?avo_debug=1` to the url. This will show the avo debugger on screen
- Test each of the [aila events](https://www.avo.app/schemas/5PhajbVijwhXVKIJtGMT/events?categories=RyFnbhx4O) listed in Avo
   - Initialised
     - From free text
     - From the example link
   - Refined
     - From free text
     - From the continue button
     - From the modify dropdown 
     - From the additional materials dropdown
   - Terminated
     - When the account is locked or there's a toxic moderation
   - Completed
     - When it reaches 10/10 sections (and only once)


JBR90:
Looks great so far, have manually tested `Lesson Plan Initiated` , `Lesson Plan Refined ` `Lesson Plan Completed` in the browser.
Did check avo but it seems the inspector is sill not working.

So much easier to follow the logic in the store now compared to the provider!

codeincontext:
> Looks great so far, have manually tested `Lesson Plan Initiated` , `Lesson Plan Refined ` `Lesson Plan Completed` in the browser. Did check avo but it seems the inspector is sill not working.
> 
> So much easier to follow the logic in the store now compared to the provider!

Thanks for looking. Just to check, the inspector is different from the web dubugger. You can trigger the debugger by setting `?avo_debug=1` in the URL, or setting `webDebugger: true` in the avo config

apps/nextjs/src/stores/AilaStoresProvider.tsx
```diff
@@ -3,7 +3,7 @@ import { createContext, useContext, useEffect, useRef, useState } from "react";
 import invariant from "tiny-invariant";
```
codeincontext, line 3: This file is where the main changes start to happen - putting the store in the context

apps/nextjs/src/stores/lessonPlanTrackingStore/actionFunctions/handleTrackStreamingComplete.ts
```diff
@@ -0,0 +1,122 @@
...
+import type { GetStore } from "@/stores/AilaStoresProvider";
+
+import type {
+  LessonPlanTrackingGetter,
+  LessonPlanTrackingSetter,
+} from "../types";
+
+const log = aiLogger("analytics:lesson:store");
+
+export const handleTrackCompletion =
```
codeincontext, line 29: This function is the bulk of the tracking logic

avo.json
```diff
@@ -7,18 +7,18 @@
     "name": "Oak's Tracking Plan"
```
codeincontext, line 7: I'm not 100% sure on the process for this. At the moment it's an avo branch, I guess if it passes QA we can merge the avo branch and put the new avo main code in here

JBR90 (approved):

PR #621, opened 2025-03-23
fix: don't throw on slack notification failure

codeincontext:
This has tripped me up a few times in local development. If the inngest server can't be contacted, we currently halt execution and don't report the moderation to the client


mantagen (approved):
Nice

PR #632, opened 2025-03-25
chore: emit error objects to console from logger

codeincontext:
The Pino logger we've been using to output errors is using a format which drops the details of an error such as the error name and stack trace. Given that we haven't needed to use Pino at all, let's fall back to the basic console.error

This has allowed us to miss important errors when debugging recent features

Example based on the following:
```
  log.error("it blew up", new Error("Unexpected Detonation"));
```

Before
CleanShot 2025-03-25 at 14 03 20@2x

After
CleanShot 2025-03-25 at 14 00 13@2x


mantagen (approved):
Good idea I was thinking the same

PR #642, opened 2025-04-01
refactor: additional material package AI-944

JBR90:
## Description

- Refactor the additional material package types / config
- Reorganise folder structure
- Allow for different prompt context options

## Checklist


packages/additional-materials/src/aiProviders/getGeneration.ts
```diff
@@ -0,0 +1,28 @@
...
+    });
+
+    return validatedDocumentObject;
+  } catch (error) {
+    if (error instanceof ZodError) {
+      throw new Error(
+        `Context schema validation failed: ${JSON.stringify(error.issues, null, 2)}`,
+      );
+    }
+    throw new Error(`Failed to generate : ${String(error)}`);
```
mantagen, line 26: https://github.com/sindresorhus/serialize-error

Could be cool to use this? More of a repo wide point potentially but could trial it in this package

mantagen (approved):
Looking good! One comment but not to block this PR.

Also could be good to put some uni tests in for some of the logic?

packages/additional-materials/src/aiProviders/getGeneration.ts
JBR90: Cool , yeah will give this a go

JBR90:
> Looking good! One comment but not to block this PR.
> 
> Also could be good to put some uni tests in for some of the logic?

Yes will add some tests for sure, wanted to check the approach was ok first

JBR90:
@mikeritson-oak - This is all behind a feature flag, needs a quick check that alia functionality is working as expected when you get a min. thanks!

mikeritson-oak (approved):
I've done a few tests on this and nothing out of the ordinary was seen. If we merge it the further tests in the normal release process will also serve as a double check. 👍

PR #646, opened 2025-04-08
chore: hubspot id in posthog

tomwisecodes:
## Description
https://www.notion.so/oaknationalacademy/RFC-Attaching-Key-Analytics-Data-to-a-Hubspot-Contact-1c126cc4e1b1805dbd5cefa413420a00

- Adds user_id from hubspot to the user profile in posthob


apps/nextjs/src/lib/analytics/hubspot/HubspotClient.ts
```diff
@@ -18,6 +26,59 @@ const getQueue = (): HubspotEvent[] => {
...
+        hubspotContactId = window.hubspotUserData.contact_id;
+        console.log("Retrieved HubSpot contact_id:", hubspotContactId);
+        clearInterval(checkForHubspotData);
+      }
+    }, 500);
+
+    // Only check for 10 seconds max
+    setTimeout(() => clearInterval(checkForHubspotData), 10000);
+  }
+};
```
tomwisecodes, line 80: This block was generated by claude, keeping the comments in there as it explains the logic fairly well. 

Wasn't sure about this time out, however testing on a few browsers it did work better with rather than without.

PR #647, opened 2025-04-10
feat: banner implementaion

tomwisecodes:
## Description

Adds the a banner promoting a webinar

When the banner should show:
logged out :white_tick:
loggined in and not demo user :white_tick:
logged in and demo user :x:

### DIFF
--- a/apps/nextjs/src/components/Header.tsx
+++ b/apps/nextjs/src/components/Header.tsx
@@ -20,6 +20,7 @@ const HamburgerButton = styled.button`
 const Header = ({ menuOpen, setMenuOpen }: Readonly<HeaderProps>) => {
   const { isDemoUser } = useDemoUser();
   const { isSignedIn } = useAuth();
+  const showTempBanner = !isSignedIn || (isSignedIn && !isDemoUser);
   return (
     <OakBox
       $position="absolute"
@@ -27,6 +28,7 @@ const Header = ({ menuOpen, setMenuOpen }: Readonly<HeaderProps>) => {
       $left="all-spacing-0"
       $right="all-spacing-0"
       $zIndex="fixed-header"
+      $bb={showTempBanner ? undefined : "border-solid-m"}
     >
       <OakFlex
         as="header"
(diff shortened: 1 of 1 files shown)


JBR90 (approved):

mikeritson-oak (changes requested):
Checked for banner on logged out/logged in users and no banner for demo users. A couple of minor issues were seen:

1. The banner interferes with/overlays the controls that sit at the top of the page when in the main Aila user journey, so the chat window, downloads window, etc. This happens on both desktop and mobile viewports:

SCR-20250424-khct
SCR-20250424-khev
SCR-20250424-kico

2. Demo user does not show the banner but their top navigation/menu section is missing the bottom border:
SCR-20250424-kjcc-2
SCR-20250424-kjnk
SCR-20250424-kjjp

3. Not an issue, more a consideration/observation - the banner shows on all pages once the user is logged in, so /aila/, downloads, faqs, help, cookies, etc. Have we considered where we want to show it? Are we happy for it to be everywhere?

mikeritson-oak (changes requested):
This has now been removed from the /aila pages, so doesn't overlap onto the lesson controls any more. The banner is showing on the help and faq pages but it doesn't cause any issues on these static pages.

We do still seem to have an issue, the same as reported in #2 previously for demo users, with the missing border between the top navigation and the main page content. As we have now removed the banner from the /aila flow, this is now also an issue for standard accounts:

SCR-20250428-kstv-2
SCR-20250428-ksoc-2

mikeritson-oak (changes requested):
All looks OK for standard users.
Demo users are not shown the banner (whether intentional or not) but on the pages that the banner is not shown, as compared to a standard user, then the main navigation bottom border is still missing for demo users.

Example pages:

labs.
/help
/faqs
/legal/cookies
/legal/accessibility-statement

If we should be showing the banner for demo users, then it is not showing, if it is to be hidden completely, then the border is missing.

mikeritson-oak (approved):
All now looks good! 👍 

Logged out - Banner shows ✅ 
Logged in (standard user) - Banner shows ✅ 
Logged in (demo user) - No banner ❌

The banner shows on the labs homepage and static pages. The banner does not show on any of the main Aila lesson creation pages.

PR #651, opened 2025-04-17
chore: remove useless logs

tomwisecodes:
## Description

These logs are only meant to show up in development but when there is an error dump a giant and useless error into the terminal that stops you from actually being able to debug as the message includes every trpc definition taking up the entire console window.


mantagen (approved):
Good idea

PR #657, opened 2025-04-25
feat: additional resources - threat detection and moderation

JBR90:
## Description

- Add interaction schema and persists generations
- Adds Lakera threat detection step
- Adds moderation step

## Issue(s)

Fixes #

## How to test

1. Go to {deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist


apps/nextjs/src/app/additional-material-spike/AdditionalMaterials.tsx
```diff
@@ -130,7 +107,10 @@ const AdditionalMaterials: FC<AdditionalMaterialsProps> = ({ pageData }) => {
         },
       });
 
-      setGeneration(res);
+      log.info("res **************", res);
```
mantagen, line 110: Can remove

packages/db/prisma/schema.prisma
```diff
@@ -1154,3 +1154,29 @@ model RagLessonPlanPart {
   @@map("rag_lesson_plan_parts")
   @@schema("rag")
 }
+
+model Interaction {
```
mantagen, line 1158: I wonder if this needs to be `AdditionalResourceInteraction` for clarity?

packages/db/prisma/schema.prisma
```diff
@@ -1154,3 +1154,29 @@ model RagLessonPlanPart {
...
+  derivedFromId String?       @map("derived_from_id")
+  derivedFrom   Interaction?  @relation("DerivedFrom", fields: [derivedFromId], references: [id])
+  derivatives   Interaction[] @relation("DerivedFrom")
+
+  adaptsOutputId String?       @map("adapts_output_id")
+  adaptsOutput   Interaction?  @relation("AdaptedOutput", fields: [adaptsOutputId], references: [id])
+  adaptedBy      Interaction[] @relation("AdaptedOutput")
+
+  config               Json    @db.JsonB
+  inputText            String? @map("input_text")
```
mantagen, line 1171: Is this optional? Should it not be?

apps/nextjs/src/app/additional-material-spike/AdditionalMaterials.tsx
```diff
@@ -130,7 +107,10 @@ const AdditionalMaterials: FC<AdditionalMaterialsProps> = ({ pageData }) => {
...
+
+      setGeneration(res.resource);
+      setModeration(res.moderation.justification ?? null);
     } catch (error) {
       log.error("error", error);
```
mantagen, line 115: I guess we don't need the first arg: `"error"` ?

packages/additional-materials/src/threatDetection/lakeraThreatCheck.ts
```diff
@@ -0,0 +1,83 @@
...
+  apiKey = process.env.LAKERA_GUARD_API_KEY,
+}: {
+  messages: Message[];
+  projectId?: string;
+  apiKey?: string;
+}): Promise<z.infer<typeof lakeraGuardResponseSchema>> {
+  if (!apiKey) {
+    log.error("Lakera API key not found");
+    throw new Error("Lakera API key not found");
+  }
```
mantagen, line 53: Do we also want to throw if no projectId ?

packages/additional-materials/src/threatDetection/lakeraThreatCheck.ts
```diff
@@ -0,0 +1,83 @@
...
+  const requestBody = {
+    messages,
+    ...(projectId && { project_id: projectId }),
+    payload: true,
+    breakdown: true,
+  };
+
+  const parsedBody = lakeraGuardRequestSchema.parse(requestBody);
+
+  const response = await fetch("https://api.lakera.ai/v2/guard", {
```
mantagen, line 64: Is Lakera api endpoint an in-code variable or an env var, perhaps should be if used elsewhere too? Not a blocker

packages/api/src/router/additionalMaterials.ts
```diff
@@ -108,26 +76,22 @@ export const additionalMaterialsRouter = router({
...
-        }
-
-        return result;
       } catch (cause) {
         const TrpcError = new Error(
           "Failed to fetch additional material moderation",
           { cause },
         );
-        log.error("Failed to fetch additional material moderation", cause);
+        log.error("Failed to fetch lesson plan", cause);
```
mantagen, line 89: additional materials moderation

packages/api/src/router/additionalMaterials.ts
```diff
@@ -108,26 +76,22 @@ export const additionalMaterialsRouter = router({
...
         Sentry.captureException(TrpcError);
```
mantagen, line 90: Do we want to add any more context to the error that's sent to Sentry?

packages/api/src/router/additionalMaterials/helpers.ts
```diff
@@ -0,0 +1,183 @@
...
+import { isToxic } from "@oakai/core/src/utils/ailaModeration/helpers";
+import type { ModerationResult } from "@oakai/core/src/utils/ailaModeration/moderationSchema";
+import type { PrismaClientWithAccelerate } from "@oakai/db";
+import { aiLogger } from "@oakai/logger";
+
+import type { LooseLessonPlan } from "../../../../aila/src/protocol/schema";
+
+const log = aiLogger("additional-materials");
+
+interface GenerateAdditionalMaterialParams {
```
mantagen, line 19: I think generally we're preferring `type =` vs `interface`

mantagen: Although I could be wrong!

packages/api/src/router/additionalMaterials/helpers.ts
```diff
@@ -0,0 +1,183 @@
...
+    provider: "openai",
+    parsedInput: {
+      documentType: input.documentType,
+      context: input.context,
+    },
+    action: input.action,
+  });
+
+  if (!result) {
+    throw new Error("Failed to generate additional material");
```
mantagen, line 44: Is the error already logged->Sentry by this point? Otherwise looks like we could be losing information

packages/api/src/router/additionalMaterials/helpers.ts
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  const { resourceId, documentType } = input;
+  const version = additionalMaterialsConfigMap[documentType].version;
+
+  const interaction = await prisma.interaction.create({
+    data: {
+      userId,
+      config: {
+        resourceType: documentType,
```
mantagen, line 59: Might be good to consider consistent terminology here.

Perhaps a glossary in Notion? (AI could create a good one I'm sure)

packages/db/prisma/schema.prisma
JBR90: There is only input text from the user when a 1 shot Aila lesson is created in the background prior to the AR being created. The adaptations are also predefined options (lower reading age, etc), although there maybe be a custom option past MVP.

When we link from OWA there will be no input text as well as it will be based in that specific lesson.

apps/nextjs/src/app/additional-material-spike/AdditionalMaterials.tsx
JBR90: This is the spike code that will go after Tom's PR is merged.

packages/api/src/router/additionalMaterials/helpers.ts
JBR90: Yeah good idea, there is some work to align the naming of this feature and to differentiate from the lesson additional materials. It's being discussed tomorrow in product steering.

mantagen (approved):
LGTM, nice!

mikeritson-oak (changes requested):
Three possible issues seen during testing:

1. A sporadic issue on initial lesson generation, the lesson details are shown but then disappear following the moderation step and placeholders are shown instead. I think this may be dependant on whether a moderation result is returned at that point:
SCR-20250507-nqfd
2. Messages which trigger threat detection are shown the relevant warning message but these messages do not persist if the browser is refreshed. All messages are removed from the chat if they were detected.
3. Lesson output does not persist if the browser is refreshed during the moderation stage, prior to completion. This one seems to also happen in Production, I don't know if we would expect it to persist at this stage?

JBR90:
> Three possible issues seen during testing:
> 
> 1. A sporadic issue on initial lesson generation, the lesson details are shown but then disappear following the moderation step and placeholders are shown instead. I think this may be dependant on whether a moderation result is returned at that point:
>    SCR-20250507-nqfd
> 2. Messages which trigger threat detection are shown the relevant warning message but these messages do not persist if the browser is refreshed. All messages are removed from the chat if they were detected.
> 3. Lesson output does not persist if the browser is refreshed during the moderation stage, prior to completion. This one seems to also happen in Production, I don't know if we would expect it to persist at this stage?

Hey Mike, i think the PR is ready for another look now #657 . I've not actually changed any code but have updated decencies and seems to be working as expected
A sporadic issue on initial lesson generation, the lesson details are shown but then disappear following the moderation step and placeholders are shown instead. I think this may be dependant on whether a moderation result is returned at that point:
Not seen this one after re-building
Messages which trigger threat detection are shown the relevant warning message but these messages do not persist if the browser is refreshed. All messages are removed from the chat if they were detected.
Unrelated to the changes in this PR, i don't think we are persisting the messages yet
Lesson output does not persist if the browser is refreshed during the moderation stage, prior to completion. This one seems to also happen in Production, I don't know if we would expect it to persist at this stage?
Not sure about this one, if on production and new expected we should add a bug ticket

PR #668, opened 2025-05-06
feat: connect additional materials moderation and threat to modals and zustand

JBR90:
## Description

-  Connects moderation and threat detection to zustand store
- Adds draft modulation and threat copy - (awaiting final copy and designs)
- Threat - 
      -  modal
      - should clear state
- Moderation 
      - adds moderation message to lesson - modal if user clicks on more guidance
      - adds moderation message to material - modal id user clicks on more gudance

// How to test

- Aila basic functionality test
- AM work is still awaiting some designs and not ready for testing

## Checklist


apps/nextjs/src/components/DialogControl/ContentOptions/AdditionalMaterialsModeration.tsx
```diff
@@ -0,0 +1,39 @@
...
+      $width="100%"
+      $height="100%"
+      $flexDirection="column"
+      $justifyContent="space-between"
+    >
+      <OakP>
+        This content that needs additional guidance. Please check carefully
+        before using. Learn more about moderation and content guidelines.
+      </OakP>
+      {moderation?.categories.map((category) => <OakP>{category}</OakP>)}
```
mantagen, line 28: Does this need a key={} on the map?

apps/nextjs/src/stores/resourcesStore/types.ts
```diff
@@ -21,7 +22,7 @@ export type StepOneFormState = {
 };
 
 export type ResourcesState = {
-  id: string;
+  id: string | null;
```
mantagen, line 25: When is id null? Before it's been persisted? Wondering if it's beneficial to have a type with id and a type without? or overkill?

mantagen (approved):
Code looks good couple suggestions/questions

apps/nextjs/src/stores/resourcesStore/types.ts
JBR90: The id is created when the user creates a additional material, what would be the benefit of the two types ?

mantagen: Maybe I wasn't fully understanding the type here.

I was thinking more if you had a DB entity AdditionalMaterialResource or whatever, that would always have an ID ... whereas if you were creating a NewAdditionalMaterialResource, it would never have an ID (until it was persisted, at which point it would also have created_at updated_at etc).

This is non-blocking and more philosophical at this point!

PR #693, opened 2025-06-03
feat: add scheduled clerk user cleanup cron job - AI-1245

codeincontext:
## Problem

Clerk user limits require periodic cleanup of inactive test and demo accounts. We need automated cleanup to run on staging environment where test accounts accumulate, but Vercel cron jobs only work on production deployments.

## Solution

Created a GitHub Action for scheduled clerk user cleanup that:

- **Scheduled execution**: Runs daily at 2 AM UTC on staging environment
- **Manual triggering**: Can be triggered manually with configurable dry-run option
- **Deletion rules**: 
  - Test accounts (emails with + or numbers): deleted after 1 month inactive
  - Regular accounts: deleted after 3 months inactive
- **Safe operation**: Includes dry-run mode and detailed logging
- **Proper authentication**: Uses Bearer token auth and Vercel protection bypass headers

## Testing Results

✅ **Dry run test**: 
```
Clerk user cleanup completed: 155 users would be deleted, 317 users remaining
```

✅ **Live run test**:
```
Found 155 users to delete out of 472 total users
Clerk user cleanup completed: 155 users deleted, 317 users remaining
```

## Pre-merge Tasks

- [x] Update `CLERK_USER_CLEANUP_URL` secret back to main deployment

## Issue(s)

Fixes AI-1245

### DIFF
--- a/.github/workflows/clerk-user-cleanup.yml
+++ b/.github/workflows/clerk-user-cleanup.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: Run Clerk User Cleanup on Staging
         run: |
-          DRY_RUN="${{ github.event.inputs.dry_run || 'false' }}"
+          DRY_RUN="${{ github.event.inputs.dry_run || false }}"
           CLERK_USER_CLEANUP_URL="${{ secrets.CLERK_USER_CLEANUP_URL }}"
           CRON_SECRET="${{ secrets.CRON_SECRET }}"
           VERCEL_AUTOMATION_BYPASS_SECRET="${{ secrets.VERCEL_AUTOMATION_BYPASS_SECRET }}"


apps/nextjs/src/app/api/cron-jobs/clerk-user-cleanup/route.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  const lastActive = new Date(user.lastActiveAt ?? user.createdAt);
+  const created = new Date(user.createdAt);
+  const now = new Date();
+
+  const timeSinceLastActive = now.getTime() - lastActive.getTime();
+  const sameDay = lastActive.toDateString() === created.toDateString();
+
+  const hasTestCharacters =
+    email.includes("+") || OAK_EMAIL_WITH_NUMBER.test(email);
```
mantagen, line 30: I think some genuine Oak emails have numbers in them. I'm not sure if people use them but they are legitimate.

apps/nextjs/src/app/api/cron-jobs/clerk-user-cleanup/route.ts
```diff
@@ -0,0 +1,115 @@
...
+      deletedCount: 0,
+      remainingUsers: result.data.length,
+    };
+  }
+
+  let deletedCount = 0;
+
+  for (const user of usersToDelete) {
+    try {
+      await deleteUser(user, isDryRun);
```
mantagen, line 71: Guessing this doesn't gets rate limited?

apps/nextjs/src/app/api/cron-jobs/clerk-user-cleanup/route.ts
```diff
@@ -0,0 +1,115 @@
...
+    return;
+  }
+
+  await clerkClient.users.deleteUser(user.id);
+  log.info(`Deleted user (ID: ${user.id})`);
+}
+
+async function cleanupUsers(isDryRun: boolean) {
+  const result = await clerkClient.users.getUserList({
+    limit: 500,
```
mantagen, line 51: Is `limit` a non-optional param?

.github/workflows/clerk-user-cleanup.yml
```diff
@@ -0,0 +1,46 @@
...
+
+permissions: {}
+
+jobs:
+  clerk-user-cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Clerk User Cleanup on Staging
+        run: |
+          DRY_RUN="${{ github.event.inputs.dry_run || 'false' }}"
```
mantagen, line 23: is 'false' here a string? Wasn't sure whether it's casted at some point

packages/api/src/router/testSupport/userPreparation/clerkUsers.ts
```diff
@@ -10,10 +10,9 @@ export const deleteOldTestUser = async () => {
     limit: 500,
   });
 
-  const NUMBERS_USER = /\d{5,10}.*@/; // jim+010203@thenational.academy
   const testUsers = result.data.filter((u) => {
     const email = u.primaryEmailAddress?.emailAddress ?? "";
-    return email.startsWith("test+") || email.match(NUMBERS_USER);
+    return email.startsWith("test+");
```
mantagen, line 15: The numbers-in-email check is removed here, but logic is part of `hasTestCharacters`?

mantagen (commented):
Looks good! Couple comments. Also, as it stands would this be running on the production deployment too?

apps/nextjs/src/app/api/cron-jobs/clerk-user-cleanup/route.ts
codeincontext: Interesting insight, thank you. Would 3 numbers in a row rule those out?

packages/api/src/router/testSupport/userPreparation/clerkUsers.ts
codeincontext: Yeah, I think that the e2e test account should stick to their own logic of having a user pool with a finite length. They will cycle a lot quicker than the 1 month on manually created test accounts (hasTestCharacters).

codeincontext:
> Also, as it stands would this be running on the production deployment too?

That's the issue I'm trying to get around by using a github action. I originally used a vercel cron, but those only run in production!

As a scheduled github action, it will trigger only in the context of the main branch on github. To be honest that ends up not being too important as Github only has secrets for staging: both CLERK_USER_CLEANUP_URL and CRON_SECRET are specific to staging so it can't call the production endpoint

I think we should definitely put a check in the cron endpoint itself though. I'll do that now

apps/nextjs/src/app/api/cron-jobs/clerk-user-cleanup/route.ts
mantagen: Yep sounds good. I've only seen the number '1' in the email, so two or more digits should be safe

.github/workflows/clerk-user-cleanup.yml
codeincontext: I don't know the exact answer of how these are interpolated, but it does seem to work as expected. It's a good question

codeincontext: DRY_RUN is ultimately only passed forward into the URL concatenation, so I suppose it doesn't matter what the type is here in this case

mantagen (approved):

PR #702, opened 2025-06-06
refactor: create material id at step 1

JBR90:
## Description

- To implement a avo tracking plan we need a id from step 1 => step 4 to link the tracked events, we were creating id only at the final step when the material was generated. This PR introduces a session id at step one that is updated throughout the flow.

## How to test

Additional materials works the same as before

## Checklist

### DIFF
--- a/apps/nextjs/src/components/AppComponents/AdditionalMaterials/hooks/useStepSubmitLogic.ts
+++ b/apps/nextjs/src/components/AppComponents/AdditionalMaterials/hooks/useStepSubmitLogic.ts
@@ -1,14 +1,25 @@
+import { aiLogger } from "@oakai/logger";
+
 import * as Sentry from "@sentry/nextjs";
 
 import { useResourcesActions } from "@/stores/ResourcesStoreProvider";
 import { trpc } from "@/utils/trpc";
 
+const log = aiLogger("additional-materials");
 const useStepSubmitLogic = () => {
-  const { submitLessonPlan, setStepNumber, generateMaterial } =
-    useResourcesActions();
+  const {
+    submitLessonPlan,
+    setStepNumber,
+    generateMaterial,
+    createMaterialSession,
+  } = useResourcesActions();
 
   const generateLessonPlan =
     trpc.additionalMaterials.generatePartialLessonPlanObject.useMutation();
+  const updateMaterialSession =
+    trpc.additionalMaterials.updateMaterialSession.useMutation();
+  const createSession =
+    trpc.additionalMaterials.createMaterialSession?.useMutation?.();
 
   // Handle submit for step 2
   const handleSubmitLessonPlan = async (params: {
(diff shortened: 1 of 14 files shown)


apps/nextjs/src/components/AppComponents/AdditionalMaterials/hooks/useStepSubmitLogic.ts
```diff
@@ -57,10 +77,28 @@ const useStepSubmitLogic = () => {
...
+        try {
+          return await createSession.mutateAsync(documentType);
+        } catch (e) {
+          const error = e instanceof Error ? e : new Error(String(e));
+          Sentry.captureException(error);
+          throw error;
+        }
+      },
+    });
+  };
```
codeincontext, line 96: This change looks fine, but coming to the file the first time is a bit confusing. I'm not sure I see anything that would need it to be a hook rather than a store action?

apps/nextjs/src/stores/resourcesStore/actionFunctions/handleSubmitLessonPlan.ts
```diff
@@ -87,6 +93,32 @@ const updateStoreWithLessonPlan = (
...
+const updateMaterialSessionWithLessonId = async (
+  resourceId: string | null,
+  lessonId: string,
+  updateSessionMutateAsync?: UseMutateAsyncFunction<
+    { success: boolean },
+    Error,
+    { resourceId: string; lessonId: string }
+  >,
+) => {
+  if (resourceId && updateSessionMutateAsync) {
```
codeincontext, line 108: Is this a valid case? If so wouldn't we want to handle it or log?

If it's just for typescript then we can use an invariant. That way if it's being called incorrectly we will know befofe it gets to production

packages/api/src/router/additionalMaterials.ts
```diff
@@ -34,13 +38,112 @@ const log = aiLogger("additional-materials");
...
+                resourceType: parsedDocType,
+                resourceTypeVersion: version,
+              },
+            },
+          });
+
+        log.info("Material session created", { resourceId: interaction.id });
+        return { resourceId: interaction.id };
+      } catch (cause) {
+        const TrpcError = new Error("Failed to create material session", {
```
codeincontext, line 85: I think TrpcError is an instance so should start with a lowercase t?

mikeritson-oak (changes requested):
It seems that the build might have failed on this PR, no preview environment available. Can you take a look and send it back to me once sorted? 👍

apps/nextjs/src/components/AppComponents/AdditionalMaterials/hooks/useStepSubmitLogic.ts
JBR90: Yes there is a ticket to move this to the store, we couldn't use the same pattern with 'mutate' as we did with the other stores so we decided to prioritise getting the feature out for now as it didn't seem as straight forward. There are a few tech debt tickets to look at after including refactoring the selector pattern used in this store to match the others.

PR #705, opened 2025-06-06
Feat/improvements to glossary

JBR90:
## Description

- File name update for download - e.g -  `Spanish colonisation and its global impact - comprehension tasks - 336f5`

prompt improvements
- americanisms
- Added min and max amount for glossary -  5-15
- Include keywords from lesson in glossary 

## Issue(s)

Fixes #

## How to test

- downloads match [lesson title] - [doc type] - [id]
- no Americanisms
- min 5 glossary words (we were seeing 3 occasionally)
 

## Checklist

### DIFF
--- a/packages/additional-materials/src/documents/additionalMaterials/glossary/buildGlossaryPrompt.ts
+++ b/packages/additional-materials/src/documents/additionalMaterials/glossary/buildGlossaryPrompt.ts
@@ -1,5 +1,5 @@
 import type { ContextByMaterialType } from "../configSchema";
-import { getLessonDetails } from "../promptHelpers";
+import { getLessonDetails, language } from "../promptHelpers";
 import {
   type AllowedReadingAgeRefinement,
   readingAgeRefinementMap,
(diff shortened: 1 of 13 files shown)


packages/additional-materials/src/documents/additionalMaterials/glossary/buildGlossaryPrompt.ts
```diff
@@ -57,8 +57,11 @@ The definition should be no longer than 200 characters.
 
 **Rules**:
 - **Do not** include markdown in your response.
-- **Do not** include any americanisms.
 - definitions start with lower case (unless it is a known acronym or proper noun) and end with a full stop.
 - terms start with capital letter.
+- minimum of 10 terms, maximum of 15 terms.
```
codeincontext, line 62: This says minimum 10 terms, but the PR description says 5?

JBR90:
> One question from testing this. Is there any reason that the export naming convention differs from that which we use in Production? If there is no specific reason behind this, would it be best to align the naming conventions?
> 
> Production: [lesson title] - [8 char id] - [doc type capitalised] Preview: [lesson title] - [doc type lowercase] - [5 char id]
> 
> image

Good point, will change to match

PR #706, opened 2025-06-06
perf: add moderation timing tracking to analytics

codeincontext:
## Summary
- Add timing measurement around `performModeration()` call in `AilaModeration.ts`
- Include `duration_ms` property in existing `reportModerationToAnalytics()` method
- Enables monitoring of moderation performance bottlenecks through existing analytics infrastructure

## Test plan
- [ ] Verify type checking passes for aila package
- [ ] Test that analytics events include duration_ms property
- [ ] Monitor moderation timing in analytics dashboard after deployment

Relates to AI-1154

### DIFF
--- a/packages/aila/src/features/moderation/AilaModeration.ts
+++ b/packages/aila/src/features/moderation/AilaModeration.ts
@@ -106,11 +106,13 @@ export class AilaModeration implements AilaModerationFeature {
       throw new Error(errorText);
     }
 
+    const startTime = Date.now();
     const moderationResult: ModerationResult = await this.performModeration({
       messages,
       content,
       retries: 3,
     });
+    const moderationDuration = Date.now() - startTime;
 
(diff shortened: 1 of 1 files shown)


packages/aila/src/features/moderation/AilaModeration.ts
```diff
@@ -144,6 +146,7 @@ export class AilaModeration implements AilaModerationFeature {
   public reportModerationToAnalytics(
     moderationResult: ModerationResult,
     moderation: Moderation,
+    durationMs?: number,
```
mantagen, line 149: Is this method used elsewhere? I'm thinking why not make durationMs non-optional?

codeincontext: Yeah, my bad. I'm going to close this for now anyway and look at telemetry for this

PR #725, opened 2025-06-19
fix: preserve keyword definition casing [AI-1322]

mantagen:
## Description

- It's been noted that the keyword definition is always sentence case, however it should not be.
- The first letter of the first word in the definition should be lower case unless the first word is a proper noun or an acronym/initialism

## Issue(s)

[AI-1322](https://www.notion.so/oaknationalacademy/Keyword-def-lowercase-21726cc4e1b180d69e19c1aee9f22931?source=copy_link)

### DIFF
--- a/packages/aila/src/protocol/sectionToMarkdown.ts
+++ b/packages/aila/src/protocol/sectionToMarkdown.ts
@@ -120,7 +120,7 @@ function renderTwoKeyObjectArray(value: readonly unknown[]): string {
     .map((v) => {
       const header = v[key1] ?? "…";
       const body = v[key2] ?? "…";
-      return `### ${toSentenceCase(header)}\n\n${toSentenceCase(body)}`;
+      return `### ${toSentenceCase(header)}\n\n${body}`;
     })
     .join("\n\n");
 }


codeincontext (approved):

mantagen:
Thanks @mikeritson-oak 

I've taken a look and both of those lessons have the text capitalised when it comes back from the LLM (see below). In the agentic system, we can target prompt to ensure that LLM returns appropriate casing.

image

image

mikeritson-oak:
I thought the purpose of this ticket was to display keyword definitions in lowercase, as per the Notion ticket and your 'After' example in the first comment? That's what I wasn't seeing. @mantagen

mantagen:
Thanks @mikeritson-oak 

I've updated the ticket's name and description to clarify, sorry for any confusion!

https://www.notion.so/oaknationalacademy/Keywords-and-definitions-preserve-casing-from-LLM-21726cc4e1b180d69e19c1aee9f22931

mikeritson-oak (approved):
I think this PR now matches the desired outcome of the ticket. We have changed requirements slightly since the initial ticket. Now, instead of adjusting any of the returned LLM output the format is maintained. We will look at amendments to the output with agentic changes, so the expected values *should* be returned by the LLM, rather than manipulated in code later.

PR #732, opened 2025-06-27
fix: [AI-1381] prevent analytics tracking on useChat errors

codeincontext:
## Summary
- Fixes an invariant error that occurs when useChat encounters an error (network timeout, server error, etc.)
- The analytics tracking was attempting to process a completion event even when the API call failed
- This caused the tracking to fail when trying to access assistant message parts that don't exist

## Changes
- Added `streamingError` state to chat store to track when streaming fails
- Added `streamingFailed` action that sets the error flag in useChat's `onError` callback
- Updated analytics tracking to check for errors before attempting to track completion
- Clear error state when new requests start
- Added Sentry error logging for analytics tracking failures

This PR is in response to an error some users have encountered while streaming that leads to a full page error. As of 9th July I haven't been able to induce this behaviour so it's hard to say if this is the right fix


apps/nextjs/src/stores/chatStore/types.ts
```diff
@@ -58,6 +59,7 @@ export type ChatState = {
     append: (action: ChatAction) => void;
     stop: () => void;
     streamingFinished: () => void;
+    streamingFailed: () => void;
```
mantagen, line 62: Very minor (and just personal preference) but I like it when naming of a callback like this signifies that it's an action (e.g. onStreamingFailed, or a verb like 'stop..' or 'append..' etc), to help distinguish it from state properties when scanning.

mantagen (approved):
Code looks good though I've not tested the functionality. Minor point on naming, but no blocker

PR #758, opened 2025-07-31
fix: rag lesson schema - cycle3 optional and all without length

mantagen:
## Description

- Ingested lesson plans will now have cycle3 as optional and broadly speaking won't have length constraints
- This PR updates the RAG search function to reflect that
- It also generally makes Cycle 3 nullable, which I actually thought it should have been already (we have Slides templates for lessons with 1 or 2 cycles) 

_**Note:** it's this last point that is the most substantial change (potential to affect application beyond the maths/agentic capability)_ 

## Issue(s)

https://www.notion.so/oaknationalacademy/Fix-new-RAG-zod-schema-24126cc4e1b180738046ed321a0dbbc9?source=copy_link

### DIFF
--- a/packages/additional-materials/src/documents/additionalMaterials/promptHelpers.ts
+++ b/packages/additional-materials/src/documents/additionalMaterials/promptHelpers.ts
@@ -5,7 +5,10 @@ export const getLessonTranscript = (transcript: string) => {
   return transcript;
 };
 
-const renderCycle = (cycle: LooseLessonPlan["cycle1"], label: string) => {
+const renderCycle = (
+  cycle: LooseLessonPlan["cycle1"] | null,
+  label: string,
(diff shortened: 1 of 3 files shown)


packages/additional-materials/src/documents/additionalMaterials/promptHelpers.ts
```diff
@@ -5,7 +5,10 @@ export const getLessonTranscript = (transcript: string) => {
   return transcript;
 };
 
-const renderCycle = (cycle: LooseLessonPlan["cycle1"], label: string) => {
+const renderCycle = (
```
mantagen, line 8: @JBR90 this was the only 'additional' change i had to make due to making cycle3 nullable ... do you think it is safe?

PR #756, opened 2025-07-29
feat: add LaTeX to image rendering for exports

codeincontext:
## Summary
- Adds MathJax support for converting LaTeX expressions to SVG
- Implements PNG conversion for LaTeX expressions using resvg
- Starts infrastructure for processing LaTeX patterns in exported documents

## Implementation Details
- Static imports for both MathJax and resvg, with correct nextjs webpack setup
- Deterministic hash-based URLs for LaTeX images
- Parallel image uploads
- Test coverage with fixture-based testing

## Progress
- ✅ LaTeX pattern detection (`findLatexPatterns`)
- ✅ LaTeX to SVG conversion using MathJax
- ✅ SVG to PNG conversion using resvg
- ✅ Test fixtures and unit tests for conversion
- ✅ `latexToImageReplacements` function for markdown conversion
- ⏳ Integration with quiz/lesson export pipeline
- ⏳ Upload to GCS (currently using placeholder URLs)
- ⏳ Manual testing with actual quiz exports

## Next Steps
- Determine best integration point in export pipeline (likely in data prep functions)
- Implement actual cloud storage upload
- Test with real quiz data containing LaTeX expressions

PR #759, opened 2025-07-31
feat: implement LaTeX image dimensions and URL handling

codeincontext:
## Summary
Improves LaTeX image handling by extracting accurate dimensions from rendered PNGs and updating the filename format to include dimensions.

## Changes
- Extract dimensions from resvg's rendered PNG output instead of parsing SVG ex units
- Update GCS filename format to include dimensions: `latex/{hash}-{width}x{height}.png`
- Simplify `latexToSvg` to return only the SVG string (dimensions come from PNG)
- Remove width option from `svgToPng` - always render at natural size with 3x zoom
- Fix `getDimensions` regex parsing with optional chaining for type safety
- Clean up alt text generation by removing resolution info

## Why these changes?
- **Accurate dimensions**: SVG ex units are font-relative and hard to convert. Getting dimensions from the rendered PNG is more reliable.
- **Filename format**: Including dimensions in the filename allows the Google Docs API to size images correctly without additional metadata.
- **Simpler API**: Removing unnecessary options makes the code cleaner and easier to use.

## Testing
- [x] Type checking passes
- [x] Updated tests to match new API
- [x] Manual testing confirms dimensions are correctly extracted

## Related
- Builds on #757 (GCS infrastructure) and #756 (LaTeX rendering)

### DIFF
--- a/packages/exports/src/dataHelpers/prepQuizForDocs.ts
+++ b/packages/exports/src/dataHelpers/prepQuizForDocs.ts
@@ -1,11 +1,12 @@
 import type { Quiz } from "../schema/input.schema";
 import { filterToMcQuestions, processQuizAnswers } from "../utils";
+import { latexToImageReplacements } from "../gSuite/docs/latexToImageReplacements";
 
-function processQuizAnswersForQuiz(
+async function processQuizAnswersForQuiz(
   answers: string[] | undefined,
   distractors: string[] | undefined,
 ) {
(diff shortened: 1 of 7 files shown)


mantagen (approved):
Nice one, looks great!

Just a couple comments though nothing blocking.

General things:
1. Guessing rate limits aren't getting hit even LaTeX
2. Room potentially for a couple of tests for new behaviours

PR #767, opened 2025-08-14
feat: revise moderation rubric and prompt

markhodierne:
Complete restructure of moderation system from 6 grouped categories to 28 individual categories with independent 1-5 Likert scoring and individual justifications.

**Key Changes:**
- 28 individual categories (l1-l2, u1-u5, s1, p2-p5, e1, r1-r2, n1-n7, t1-t6)
- Individual category assessment with detailed prompts  
- New response format with abbreviated codes
- Transformation layer for API compatibility
- All existing functionality maintained via helper abstraction

**Files changed:** 19 files with moderation system restructure
**Testing:** All tests passing, zero breaking changes


packages/additional-materials/src/moderation/generateAdditionalMaterialModeration.ts
```diff
@@ -20,5 +23,17 @@ export const generateAdditionalMaterialModeration = async ({
     provider,
   );
 
-  return moderation;
+  // Transform the LLM response format to the internal ModerationResult format
+  // LLM returns: { scores, justifications, flagged_categories }
+  // Internal expects: { scores?, categories, justification? }
+  const result: ModerationResult = {
```
mantagen, line 29: Wondering if makes sense to version what gets persisted here rather than shoehorn it into the previous format?

mantagen: Looks like the idea is to make it backwards compatible to minimise required changes but is that just temporary?

packages/core/src/utils/ailaModeration/moderationPrompt.ts
```diff
@@ -1,46 +1,87 @@
...
+<category>
+Category Title: '${cat.title}'
+Output Score Code (Abbreviation): '${cat.abbreviation}'
+Original Full Code (for context): '${cat.code}'
+Description for Assessment: ${cat.llmDescription}
 Rating Criteria:
-- 5 ${categoryGroup.criteria5}
-- 1 ${categoryGroup.criteria1}
-</category-group>
+- 5: ${cat.criteria5}${
```
mantagen, line 27: Could be a `ratingCriteriaText` function with a `list.filter().map().join()` which might be a bit nicer than all the ternaries?

mantagen (approved):
Looks great! A couple comments/questions but nothing blocking so long as it gets QA review

PR #780, opened 2025-08-28
fix: "based on" retrieval - upgrade quizzes to v2 and use lengthless schema

codeincontext:
We have a recurring error in Sentry where the basedOn RAG lesson fails to parse with:

```
"path": ["starterQuiz"],
"message": "Expected object, received array"
```

https://oak-national-academy.sentry.io/issues/16562945/events/477a5d202e834d79ba34a14938a9ebb5/?project=4507089597300816 

This indicates that we're fetching the basedOn lesson with a v1 quiz (the v2 quiz schema is an object)

This PR:
- Uses the upgradeQuizzes function to bring the fetched lesson plan up to the v2 quiz schema
- Makes tryWithErrorReporting throw in non-production environments (note that in this case it will be caught by a higher try/catch

Second problem:
- We're parsing the RAG lesson with `LessonPlanSchemaWhilstStreaming`. This has length requirements for fields like checkForUnderstanding.answers and keywords, which the RAG content doesn't adhere to
This PR:
- Uses CompletedLessonPlanSchemaWithoutLength instead

### Testing

Does "based on" work?

### DIFF
--- a/packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
+++ b/packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
@@ -1,27 +1,14 @@
 import type { PrismaClientWithAccelerate } from "@oakai/db";
 
-import invariant from "tiny-invariant";
-
 import { tryWithErrorReporting } from "../../helpers/errorReporting";
-import type { LooseLessonPlan } from "../../protocol/schema";
+import type { PartialLessonPlan } from "../../protocol/schema";
 import { CompletedLessonPlanSchemaWithoutLength } from "../../protocol/schema";
-import { upgradeQuizzes } from "../../protocol/schemas/quiz/conversion/lessonPlanQuizMigrator";
-
-// asserts that the object has a lessonPlan property
-function hasLessonPlanKey(obj: unknown): obj is { lessonPlan: unknown } {
-  return (
-    typeof obj === "object" &&
-    obj !== null &&
-    "lessonPlan" in obj &&
-    typeof obj.lessonPlan === "object" &&
-    obj.lessonPlan !== null
-  );
-}
+import { migrateLessonPlan } from "../../protocol/schemas/versioning/migrateLessonPlan";
 
 export async function fetchLessonPlanContentById(
   id: string,
   prisma: PrismaClientWithAccelerate,
-): Promise<LooseLessonPlan | null> {
+): Promise<PartialLessonPlan | null> {
   const lessonPlanRecord = await prisma.lessonPlan.findFirst({
     where: {
       id,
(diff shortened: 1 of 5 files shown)


packages/aila/src/helpers/errorReporting/index.ts
```diff
@@ -54,6 +54,9 @@ export const tryWithErrorReporting = <T>(
       level,
       extra,
     });
-    return null;
+    if (process.env.NEXT_PUBLIC_ENVIRONMENT === "prd") {
+      return null;
```
mantagen, line 58: 🙃

packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
```diff
@@ -18,8 +19,24 @@ export async function fetchLessonPlanContentById(
   if (!lessonPlanRecord) {
     return null;
   }
+
+  const upgradeResult = await upgradeQuizzes({
```
mantagen, line 23: Would be quite nice to have a rolling updates pipeline which is always called ... though mostly will be a trivial noop.

packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
```diff
@@ -18,8 +19,24 @@ export async function fetchLessonPlanContentById(
...
+    persistUpgrade: null,
+  });
+
+  // upgradeQuizzes expects and returns an object with a lessonPlan property
+  const upgradedLessonPlan =
+    upgradeResult.data &&
+    typeof upgradeResult.data === "object" &&
+    "lessonPlan" in upgradeResult.data
+      ? upgradeResult.data.lessonPlan
+      : null;
```
mantagen, line 36: Is there a need to error handle in case we end up with null here?

mantagen (approved):
Nice

packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
codeincontext: Yep, I agree. I think we can start to think about that as we start generalising the schema updates. eg: quizv3 gives us more quiz schema versions, and then later we'll have more fields using schemas

packages/aila/src/utils/lessonPlan/fetchLessonPlanContentById.ts
codeincontext: Yeah you're right. Will fix

mikeritson-oak (approved):
I've done some checks on this change and all looks good: 

- Based on lesson relevant to the lesson input
- Based on lesson less relevant to the input
- No based on lesson

No issues seen during testing.

PR #784, opened 2025-08-31
feat: add 4-column horizontal layout for image quiz questions

codeincontext:
## Summary
- Add selective image sizing based on table context detection
- Implement 4-column horizontal layout for image-based multiple choice questions  
- Use constrained dimensions (120x90pt) only for images in 4-column quiz tables
- Preserve LaTeX image calculated dimensions regardless of context

## Test plan
- [ ] Verify image quiz tables render with correct image dimensions
- [ ] Confirm LaTeX images maintain calculated dimensions
- [ ] Test regular content images use standard stem dimensions


mikeritson-oak (changes requested):
Adam taking another look. Lesson looked OK but still vertical in the PDF.

packages/aila/src/protocol/schemas/quiz/conversion/rawQuizIngest.test.ts
```diff
@@ -179,4 +180,207 @@ describe("convertHasuraQuizToV2", () => {
...
+              answer_is_correct: true,
+            },
+          ],
+        },
+        feedback: "",
+        hint: "",
+        active: true,
+      },
+    ];
+    const result = convertHasuraQuizToV2(quizWithOnlyLetter);
```
mantagen, line 256: Should this now be v3 ?

mantagen: (I guess techincally doesn't matter if no image attrs?)

mantagen (approved):
Looks very good codewise! Just a question around the v2 image function contained therein

PR #791, opened 2025-09-03
fix: only fetch owa data when clerk is hydrated

JBR90:
## Description

Fixes an intermittent issue where tRPC requests would fail with:
`TRPCClientError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`

**When it happens**
On first load from OWA => Aila Teaching materials, if you reload the page after the errror it works fine. 

**What i think is causing this**

- User is already signed in OWA.
- They click a link to  labs.thenational.academy/aila/teaching-materials
- The page renders and auto-calls fetchOwaData() from the store before Clerk is hydrated with new session cookie it creates when you are logged in and switch domains.
- That triggers a tRPC call to /api/trpc, but since no session cookie exists yet, Clerk middleware on TM returns an HTML sign-in page.
- The tRPC client chokes trying to parse that as JSON.

**Fix**
I've added a check in the store to check is the user isLoaded before fetching owa data.
I've added callWithHandshakeRetry to try 3 times before error to allow time for the session cookie to update, if `Unexpected token '<', "<!DOCTYPE` in error refresh the user including auth token.

## How to test

We need to quick test that TM works as before from a link from OWA, this a very small change and doesnt need a full feature test.
We need to test on production once this is merged as I cant reproduce this locally

## Checklist


apps/nextjs/src/stores/resourcesStore/utils/callWithHandshakeRetry.ts
```diff
@@ -0,0 +1,163 @@
...
+  const { unauthorized, sessionNotReady } = errorAnalysis;
+
+  if (unauthorized) {
+    return baseDelayMs;
+  } else if (sessionNotReady) {
+    return attempt * 500; // 500ms, 1000ms, 1500ms for session propagation
+  } else {
+    return baseDelayMs * Math.pow(2, attempt - 1);
+  }
+}
```
codeincontext, line 71: We could avoid some of the logic in this file with a package like [p-retry](https://github.com/sindresorhus/p-retry)

codeincontext (approved):
Great to see this problem sorted!

I think that we could simplify the retry logic a little if we can look at the status code and retry with a package like [p-retry](https://github.com/sindresorhus/p-retry), but it's not a blocker

PR #809, opened 2025-09-17
fix: set 1200px size limit on RAG images

codeincontext:
We found a bug where OpenAI refused to describe an image as it was 20mb. Here's the offending image:
https://oaknationalacademy-res.cloudinary.com/image/upload/v1702545884/euy1sjtfakobqdfe4pyh.png

This PR adjusts the cloudinary URL to constrain images to 1200px in either dimension


mantagen (approved):
Looks great

mikeritson-oak (approved):
This looks good for the given example prompt and I've also tested out a few other lessons. No issues seen and large images, which caused the original issues, observed at reduced sizes.

PR #810, opened 2025-09-18
fix: agentic non maths old rag

mantagen:
## Description

- in the agentic flow, if the lesson plan is non-maths, use the old RAG system -- because new RAG doesn't have non-maths content in it yet

## Issue(s)

Fixes ai-1617

## How to test

1. Go to preview with agentic feature flag on
2. Generate a maths lesson e.g. "_ks4 maths lesson on circle theorems_" -> should see 'new' Oak maths lessons e.g.
3. 
image

4. Generate a non-maths lesson e.g. "_Create a lesson plan about the end of Roman Britain for key stage 3 history_" -> should see 'old' Oak lessons offered as RAG:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ yarn-error.log*
 .claude/env.local.json
 .claude/prompts
 *.local.md
+
+AGENT.md
--- a/.gitignore
+++ b/.gitignore
@@ -67,5 +67,3 @@ yarn-error.log*
 .claude/env.local.json
 .claude/prompts
 *.local.md
-
-AGENT.md
--- a/packages/aila/src/core/chat/AilaStreamHandler.ts
+++ b/packages/aila/src/core/chat/AilaStreamHandler.ts
@@ -3,7 +3,6 @@ import { prisma as globalPrisma } from "@oakai/db/client";
 import { aiLogger } from "@oakai/logger";
 import { getRagLessonPlansByIds } from "@oakai/rag";
 
-import { migrateLessonPlan } from "protocol/schemas/versioning/migrateLessonPlan";
 import type { ReadableStreamDefaultController } from "stream/web";
 import invariant from "tiny-invariant";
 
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -67,3 +67,5 @@ yarn-error.log*
 .claude/env.local.json
 .claude/prompts
 *.local.md
+
+AGENT.md
```
codeincontext, line 71: I think that we would want agent rules to be checked into the project, and personal rules in a locally gitignored file? For example, for claude there's a checked in claude.md, and I have a user level claude.md with my personal rules

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -212,6 +216,41 @@ export class AilaStreamHandler {
           return quiz;
         },
         fetchRagData: async ({ document }) => {
+          const { subject, keyStage, topic, title } = document;
+          invariant(title, "Document title is required to fetch RAG data");
+
+          const isMaths = subject?.toLowerCase().startsWith("math") ?? false;
+
+          if (!isMaths) {
+            // temporaryily use only RAG if subject is not maths -- we only have maths in the new RAG
```
codeincontext, line 225:
```suggestion
            // temporarily use only RAG if subject is not maths -- we only have maths in the new RAG
```

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -212,6 +216,41 @@ export class AilaStreamHandler {
...
+              starterQuiz: {
+                version: "v2",
+                questions: [],
+                imageAttributions: [],
+              },
+              exitQuiz: {
+                version: "v2",
+                questions: [],
+                imageAttributions: [],
+              },
```
codeincontext, line 251: I know this is WIP, but it might be worth a comment to explain why we don't need quizzes in the rag

.gitignore
mantagen: Whoops you're right, didn't mean to commit!

codeincontext (approved):

PR #811, opened 2025-09-22
fix: perssit non-maths aila agentic RAG data

mantagen:
## Description

- Previously RAG data was not being persisted in non-maths agentic flow
- This PR stores the RAG lesson metadata agent the chat session


codeincontext (approved):
`this._chat.relevantLessons =` sure looks like a code smell, but I agree that it's the right thing to do for now 👍

PR #813, opened 2025-09-24
fix: always migrate chat when fetching from database

mantagen:
## Description

- In some places we fetch the chat without using the Aila persistence adapter, and in these places we're not migrating which has led to errors being thrown and the user not able to complete their journey E.G. https://oak-national-academy.sentry.io/issues/65444362
- This PR ensures that everywhere we fetch lesson chats, that we're migrating the data
- **To reproduce:** old lessons (with v2 quizzes) erroneously cause a 404 if navigating to the /downloads page directly (without going to the chat page first, which migrates the data on load)

## To test

- Find an old lesson (with v2 quizzes) and navigate straight to /downloads (don't open the lesson chat page) 
- The page should no longer 404


codeincontext (approved):
Impressive work. And wow, that's a big oversight. Thank you

PR #943, opened 2026-02-06
fix: [AI-2072] shorten zip filenames to avoid Windows path length limit

codeincontext:
## Summary
- Windows users couldn't extract downloaded zips when lesson titles were long, exceeding the 260 char path limit
- Zip filename simplified: `aila: {long title} all resources.zip` → `{truncated title}.zip`
- Internal filenames simplified: `{title} - {id} - Lesson plan.docx` → `Lesson plan.docx` (the folder already has the title)
- Added `truncateFilename()` utility that cuts at word boundaries (default 50 chars)
- Kept `sanitizeFilename()` unchanged (strips special chars only)

Fixes AI-1629

## Test plan
- [x] Download all resources for a lesson with a long title
- [x] Verify zip filename is short and descriptive
- [x] Verify internal files are named by type only (e.g. "Lesson plan.docx")
- [x] Verify the zip extracts correctly on Windows


JBR90 (approved):

assadk88:
For any reviewers, we are now sticking with this heuristic based truncation approach. I've merged main and added some additional test cases. Re-validated and tested the approach on MacOS and it works fine, however Windows is the issue and it will need to be re-tested on that @mikeritson-oak.

@neil-hughes FYI this PR is scoped to filenames only. If we want curated short titles for the browser in the future, the shortTitle LLM approach in [AI-2072](https://www.notion.so/oaknationalacademy/Generate-shortTitle-in-title-agent-for-filenames-and-page-titles-32526cc4e1b180ae827efe83d174f4c9?v=c76315f391b74cf2bd6722b8bdcbc3e6&source=copy_link) would be the way to go but that's a heavier change (as discussed). As it stands browser tab titles are untouched.

neil-hughes (approved):
LGTM @assadk88 - seems like the smartest approach to fix the specific Windows bug. We can review browser tab titles when we revisit this area in the upcoming period.

mikeritson-oak (approved):
I rechecked this and all looks good, as per previous testing. It's been noted that we should also follow up with a ticket to look at how exports are handled on teaching materials.
Good to merge. 👍

PR #981, opened 2026-03-10
feat: content guidance chat banner and modal - AI-2051

codeincontext:
## Summary

Replace the old blue moderation chat bubble (which opened a feedback form modal via a React context chain) with a severity-aware content guidance banner and informational modal.

### What changed

- Chat banner: blue bg-videoBlue → yellow bg-lemon30 with info icon. Text and link vary by severity level:
    - content-guidance (l/, u/, s/, p/): "…may require additional consideration…" + "View content guidance."
    - enhanced-scrutiny (r/): "…requires enhanced scrutiny." + "View details."                                                                          
    - heightened-caution (e/): "…requires heightened professional caution." + "View details."
  - Content guidance modal: new OakModalCenter-based modal with OakP body text, safety guardrails link, and "Continue" button. Single category shows the
   long description; multiple categories use the spec copy ("This lesson has been flagged under multiple content guidance categories…"). Replaces the   
  old ModerationFeedbackForm which collected user comments via a Radix dialog.                                                                          
  - Unified locking moderation modal: renamed toxic-content-modal → locking-moderation-modal and made it accept heading and body props so it can serve  
  both toxic and highly-sensitive moderation. ChatModerationDisplay now renders a single LockingModerationModal with text driven by getSafetyResult()   
  instead of branching to separate view components. Closing the modal redirects to /aila.
  - Deleted redundant components: removed ToxicModerationView, HighlySensitiveModerationView, ToxicContentModal (replaced by LockingModerationModal),   
  ModerationFeedbackForm, ModerationFeedbackModal, ChatModerationContext.                                                                               
  - RHS panel: removed GuidanceRequired banner from lesson plan display (component kept — still used by share page).
  - Simplified architecture: the old flow was ChatModerationProvider (context) → useModerationModal (hook) → ModerationFeedbackModal (Radix dialog) →   
  ModerationFeedbackForm. This existed to pass chatId to the feedback form for survey submission. The new informational modal doesn't need chatId, so we
   use local useState in ModerationMessage instead. The locking moderation views were two near-identical wrapper components differing only in
  heading/body text — now consolidated into one.           

### Data changes

- `moderationCategoriesOakService.json`: added `severityLevel` field to all 8 category groups (`content-guidance`, `enhanced-scrutiny`, `heightened-caution`, `highly-sensitive`, `toxic`)
- `helpers.ts`: new `getDisplayCategories()`, `getHighestSeverity()` (shared by `ModerationMessage` and modal), `severityPriority`, `SeverityLevel`, `DisplayCategory`

### Follow-up

- Share page `GuidanceRequired` update

### Files

| File | Change |
|------|--------|
| `moderationCategoriesOakService.json` | Add `severityLevel` to all groups |
| `helpers.ts` | Add `getDisplayCategories()`, `getHighestSeverity()`, `severityPriority`, `SeverityLevel`, `DisplayCategory` |
| `helpers.test.ts` | **New** — 13 tests for `isToxic`, `getSafetyResult`, `getDisplayCategories` |
| `ModerationMessage.tsx` | Severity-aware banner, info icon, local dialog state, uses `getHighestSeverity()` |
| `ModerationMessage.stories.tsx` | Stories per severity level + multiple categories |
| `layout.tsx` | `moderation` role: `bg-videoBlue` → `bg-lemon30` |
| `content-guidance-modal.tsx` | **New** — `OakModalCenter` modal with `OakP` body, spec copy for multiple categories |
| `content-guidance-modal.stories.tsx` | **New** — stories derive data from `getDisplayCategories()` |
| `ChatModeration.tsx` | Simplified to toxic gate + passthrough |
| `chat-lessonPlanDisplay.tsx` | Remove `GuidanceRequired` from RHS |
| `index.stories.tsx` | Remove `ChatModerationProvider` decorator |
| `ModerationFeedbackForm.tsx` | **Deleted** |
| `ModerationFeedbackModal.tsx` | **Deleted** |
| `ChatModerationContext.tsx` | **Deleted** |

## Test plan

There is a doc for copy in the ticket

[FIGMA](https://www.figma.com/design/N08ZrlKXfvFHXTOex5orUq/Moderation?node-id=791-1409&m=dev)
There ares some slight differences to the design due to Oak components behaving slightly different to designs.
- Text area has build in bottom margin for label that cannot be overidden
- You cant hide the logo on feedback confirmation modal

These could be follow up Oak components tickets if needed, but beyond the scope of this PR

- [ ] `mod:sen` trigger → yellow banner with info icon + severity text + "View content guidance." link
- [ ] Click link → `OakModalCenter` with category title/description, "Continue" dismisses
- [ ] `mod:tox` → locking modal with "Potential misuse of Aila detected" heading, feedback form, redirect to `/aila` on close
- [ ] `mod:hs` → locking modal with "Highly sensitive topic" heading, feedback form, redirect to `/aila` on close    
- [ ] RHS panel has no moderation banner
- [ ] Unit tests: `pnpm --filter @oakai/core test -- --testPathPatterns ailaModeration --no-coverage`
- [ ] Storybook: check all ModerationMessage and ContentGuidanceModal story variants render correctly


apps/nextjs/src/components/AppComponents/Chat/chat-message/ModerationMessage.tsx
```diff
@@ -29,38 +42,71 @@ export function getModeration(
   return null;
 }
 
+const severityDisplay: Record<SeverityLevel, { banner: string; link: string }> =
```
neil-hughes: Extremely petty suggestion, but I'd prefer more precise names like `bannerText` and `linkText` here - both `banner` and `link` are suggestive of more complex objects on first glance.

apps/nextjs/src/components/AppComponents/Chat/chat-message/ModerationMessage.tsx
```diff
@@ -29,38 +42,71 @@ export function getModeration(
...
 }: Readonly<{ forMessage: ParsedMessage }>) {
   const persistedModerations = useModerationStore((state) => state.moderations);
   const moderation = getModeration(forMessage, persistedModerations);
-  const { moderationModalHelpers } = useChatModeration();
+  const [modalOpen, setModalOpen] = useState(false);
 
   if (!moderation) {
     return null;
   }
+  console.log("Moderation component rendering with moderation:", moderation);
```
neil-hughes: Stray `console.log` here which I think would be best deleted rather than converted to using the logger.

apps/nextjs/src/components/AppComponents/Chat/Chat/ChatModerationDisplay.tsx
```diff
@@ -1,12 +1,28 @@
...
 import type { PersistedModerationBase } from "@oakai/core/src/utils/ailaModeration/moderationSchema";
 
-import { Flex } from "@radix-ui/themes";
+import { useRouter } from "next/navigation";
 
-import HighlySensitiveModerationView from "../highly-sensitive-moderation-view";
-import ToxicModerationView from "../toxic-moderation-view";
+import { LockingModerationModal } from "../locking-moderation-modal";
+
+const getLockingModerationModalText = (safety: SafetyResult) => {
```
neil-hughes: What are the other possible SafetyResults? I guess at the moment we have "highly-sensitive" and, basically, "everything else"... but now that this modal text has multiple options I wonder if this function should check specifically for "highly-senstive" and [whatever-the-misuse-flags-are], and then show a default message. Even better would be a compile-time type failure, so if we add a new SafetyResult in future we are forced to create the new modal text immediately.

apps/nextjs/src/components/AppComponents/Chat/content-guidance-modal.tsx
```diff
@@ -0,0 +1,91 @@
...
+  OakHeading,
+  OakIcon,
+  OakModalCenter,
+  OakP,
+  OakPrimaryButton,
+  OakSecondaryLink,
+  OakSpan,
+} from "@oaknational/oak-components";
+
+function getTitle(categories: DisplayCategory[]): string {
```
neil-hughes: Lint warning on unused function here

packages/core/src/utils/ailaModeration/helpers.ts
```diff
@@ -106,6 +106,53 @@ export function getCategoryGroup(category: string) {
   );
 }
 
+export const severityPriority = [
```
neil-hughes: is "helpers" the best possible file for all of this? It feels like we could be a bit more organised here with some more disciplined file naming/refactoring. Perhaps a "severityLevel" module, or multiple modules? I'm wary of "helpers" becoming a general bucket of functionality.

apps/nextjs/src/components/AppComponents/Chat/chat-message/ModerationMessage.tsx
JBR90: Opps - will remove

apps/nextjs/src/components/AppComponents/Chat/content-guidance-modal.tsx
JBR90: Will remove

packages/core/src/utils/ailaModeration/helpers.ts
```diff
@@ -106,6 +106,53 @@ export function getCategoryGroup(category: string) {
...
+  severityLevel: string;
+};
+
+export function getDisplayCategories(
+  result: ModerationBase,
+): DisplayCategory[] {
+  const slugs = categorySlugs(result);
+
+  return slugs.flatMap((slug) => {
+    const category = findOakServiceCategory(slug);
```
neil-hughes: `getDisplayCategories` currently drops any non-Oak-service category. 

Is it worth keeping the legacy fallback here as well, since `moderationGuidanceText` (line 88) and its tests still explicitly support historical/legacy moderation codes? Without that, a persisted legacy category like `l/strong-language` would produce an empty modal.

A simple fallback would be: if \findOakServiceCategory` misses, try `findLegacyCategory` and map it to something like:

- shortDescription: legacy.title

- longDescription: "Contains ${legacy.userDescription}. Check content carefully."

- severityLevel: "content-guidance"

apps/nextjs/src/components/AppComponents/Chat/locking-moderation-modal.tsx
```diff
@@ -0,0 +1,122 @@
...
+            </OakLink>
+            .
+          </OakP>
+        )}
+
+        {showFeedback && !hasSubmitted && (
+          <OakMultilineText
+            $height={"spacing-120"}
+            charLimit={500}
+            initialValue={comment}
```
neil-hughes, line 87: Curious about this - it looks like the textbox only takes the comment value when it first appears? Existing feedback is loaded later, so is it possible that the textbox can stay blank even though there is already a saved comment in state? i.e. if a user already left feedback, the form state could get updated but the textbox doesn't, so the UI would look empty but the hidden comment would get submitted!

neil-hughes (changes requested):

apps/nextjs/src/components/AppComponents/Chat/content-guidance-modal.tsx
JBR90: Yep that's gone now, I noticed that as your were reviewing i think

apps/nextjs/src/components/AppComponents/Chat/chat-message/ModerationMessage.tsx
JBR90: That's gone. II noticed that as you were reviewing.

mikeritson-oak (changes requested):
I've added some notes and tagged you in Notion @JBR90. 👍

apps/nextjs/src/components/AppComponents/Chat/locking-moderation-modal.tsx
JBR90: Added in a isLoading from the form hook so we don't render the text box unless previous comments have first been loaded

apps/nextjs/src/components/AppComponents/Chat/Chat/ChatModerationDisplay.tsx
JBR90: I have created a new type that just includes the two locking moderation types, isToxix and highlySensitive and have added a type safe map for the messages.

packages/core/src/utils/ailaModeration/helpers.ts
JBR90: Have broken these out into individual files rather than having them all in one helper's file. Agreed. It was getting far too big.

packages/core/src/utils/ailaModeration/helpers.ts
JBR90: Have it have added in a check for legacy slugs if none are found for the new version.

JBR90:
> I've added some notes and tagged you in Notion @JBR90. 👍

I've added some notes into the doc. Spoke with Anna about this and in general we are going to go for consistency with oak components. I have added a couple of follow-up housekeeping tickets for the other issues.

neil-hughes (approved):
@JBR90 This looks great code-wise, much cleaner with the refactoring - really great work!

Only comment I have is the UI when `isLoading` is true on the form - is it worth adding a little spinner or something to show there's some background activity for better UX? It's a minor change if you decide to add it - will approve the PR and leave that for you to decide.

Nice job!

mikeritson-oak:
The work from this PR is also in #1002 and being tested over there.

PR #987, opened 2026-03-17
chore: switch to new aila auth key for curriculum api

JBR90:
## Description

Details here - https://www.notion.so/oaknationalacademy/Curriculum-DB-Update-hasura-credentials-instructions-31926cc4e1b180faac50d71843a7e9b1

- Each client now has its own api key for each environment
- Api keys are passed as authorisation headers rather than custom headers
- We use the same prod key in dev , staging and prod (read access only for Aila)

## How to test

1. Go to deployemnt
Go to an Oak lesson overview page - https://www.thenational.academy/teachers/programmes/geography-secondary-ks3/units/tectonic-hazards-why-is-the-earth-restless/lessons/inside-earth

2. Click on 'Create more with Ai'

3. Extract this part off of the url or use this example at the end of the preview build URL `aila/teaching-materials?lessonSlug=inside-earth&programmeSlug=geography-secondary-ks3&docType=additional-glossary`

4. You should see a teaching material generate from the OWA lesson with no errors.

## Checklist


mikeritson-oak (changes requested):
First few attempts at this are returning ‘Internal server error’ messages in the UI, along with 500 errors on the tRPC calls in the console.

Image
Image

mikeritson-oak (approved):
Retested and all now working as expected. The issue with errors was caused by a stale root domain being used, rather than the branch URL from Vercel. All looks good. 👍

neil-hughes (approved):
Looks good to me @JBR90 - only minor note is the spelling of "CURRIC" appears to generate a warning in the linter. Would it be better to use "CURRICULUM" like the existing variables? Or perhaps we can add an exception for "CURRIC" as an allowed abbreviation in cspell?

PR #991, opened 2026-03-23
fix: handle agentic threat detection before generation init

neil-hughes:
## Description

In agentic mode, Aila was not correctly handling detected threats on first prompt due to the order in which generation was initialised and errors were handled.

This patch:

- Fixes agentic Aila threat handling so a detected threat can be processed before generation has been initialised
- Adds test coverage for the agentic first-prompt threat path
- Adds test coverage for AilaChat.generationFailed() when handling threat vs non-threat errors

## Issue(s)

Fixes #AI-2078

## How to test

1. Open Aila chat with agentic Aila enabled
2. Send a prompt that is detected as a threat
3. You should see the normal threat-detected UI response instead of a hanging UI (in the console this hanging UI would register as an internal "Generation not initialised" failure and the UI would not be updated)

### DIFF
--- a/packages/aila/src/core/Aila.test.ts
+++ b/packages/aila/src/core/Aila.test.ts
@@ -3,8 +3,12 @@ import type { Polly } from "@pollyjs/core";
 import { setupPolly } from "../../tests/mocks/setupPolly";
 import type { AilaCategorisation } from "../features/categorisation";
 import { MockCategoriser } from "../features/categorisation/categorisers/MockCategoriser";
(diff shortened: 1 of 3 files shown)


codeincontext (approved):
Sensible fix. Thank you

mikeritson-oak (approved):
All looked good in the preview branch. Threat detection was active on initial prompts and subsequent requests. A couple of questions came from testing but these can be picked up at a later point.

PR #992, opened 2026-03-23
fix: update example lesson key stage from 3 to 2 (AI-2093)

codeincontext:
## Summary
- The example prompt on the chat start screen ("The end of Roman Britain") is meant to demonstrate RAG-based lesson generation, where Aila finds relevant Oak lessons and offers them as a basis
- This content has moved from KS3 to KS2 in the Oak curriculum, so the current example returns no RAG results — the user sees no lessons to base theirs on
- Updated the example to KS2 to match where the content now lives

## Test plan
- [x] Click the example on the chat start screen
- [x] Confirm Aila finds relevant Oak lessons and offers them as options

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/chat-start.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/chat-start.tsx
@@ -28,9 +28,9 @@ const log = aiLogger("chat");
 
 export const exampleMessages = [
   {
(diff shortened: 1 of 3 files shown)


mikeritson-oak (approved):
Veni, vidi, verified! 👍 
This change marched through QA without resistance — all roads lead to green checks. ✅

neil-hughes (approved):

PR #1004, opened 2026-04-02
chore: post-threat detection switch code cleanup

neil-hughes:
## Description

This PR cleans up some accumulated tech debt surfaced during the threat detection switch. In particular:

- Unify threat detection provider selection: previously Aila chat & Teaching Materials independently loaded separate primary threat detection providers. This PR creates a unified place in code to define the threat protector, making it easier to change provider in future, and ensures that chat/TM remain in sync.
- Remove unused BasicThreatDetector: This was dead code, a duplicated unused class which served as an example threat detector implementation
- Clean up Prisma wildcard export. This fixes a type issue which made the dev logs extremely noisy, slowing down development. Some additional unhelpful log messages were removed to allow easier focus on important log messages in future which developing.
- Fix a minor next.js issue with a link receiving an empty href in the footer: This clears up an error in next.js during development, which will help surface real errors in future.

## How to test

There are no user-facing changes so site behaviour should be entirely unchanged. Model Armor should continue to provide threat detection.

1. Go to https://oak-ai-lesson-assistant-website-8n2j650x5.vercel.thenational.academy
2. Verify essential functions are working as usual: Aila chat, threat detection, teaching material generation


packages/db/index.ts
```diff
@@ -1,4 +1,33 @@
-export * from "@prisma/client";
+export {
```
JBR90, line 1: Currious why these are explicitly defined rather than exporing everything ?

apps/nextjs/src/components/Footer.tsx
```diff
@@ -213,7 +213,7 @@ const FooterButton = ({
   const element = href ? Link : "button";
   return (
     <StyledOakLink
-      href={href ?? ""}
+      href={href}
```
JBR90, line 216: Nice one! Been meaning to sort this

JBR90 (approved):

JBR90 (approved):
Looks good to me

JBR90 (approved):

packages/db/index.ts
neil-hughes: Doing a full re-export from `@prisma/client` was causing some extremely noisy console warnings (and being more specific about what we're exposing leads to a cleaner surface too - but mostly trying to clean up the warnings)

mikeritson-oak (approved):
I've done general testing, threat detection and moderation checks on lesson planner and teaching materials. No issues seen. 👍

PR #1007, opened 2026-04-07
feat(core): [AI-2109] replace inngest with background tasks

neil-hughes:
## Description

This PR removes the dependency on Inngest, replacing it with a Vercel background task scheduler using `waitUntil`.

The six functions in Inngest which send Slack notifications have been brought into a simpler function set up in the `core` package. The Slack notifications, destinations and content are themselves unchanged - all user-facing behaviour should be identical to before.

## Issue(s)

Fixes #AI-2109

## How to test

1. Go to {deployment_url}
2. Trigger Slack notifications through:
   a.  threat detection in Teaching Materials
   b. threat detection in Aila Chat
   c. Moderation in Teaching Materials
   d. Moderation in Aila Chat
   e. User Rate Limit hit
   f. User banned
 
Each notification should appear in the proper channel with the same content as previously.

## Checklist

- [X] Manually tested


packages/core/src/models/transcript.ts
```diff
@@ -81,9 +81,8 @@ export class Transcripts {
         snippetId: createdSnippet.id,
       });
 
-      // await inngest.send({
-      //   name: "app/snippet.embed",
-      //   data: { snippetId: createdSnippet.id },
+      // await scheduleSnippetEmbedding({
```
JBR90, line 84: Is this intentional commenting out ?

neil-hughes: It was already commented out - I just converted it to the new format in case we ever re-enable it. Probably excessive paranoia - arguably a better approach would have been to delete the comment, but was trying to be thorough in removing references to Inngest in case I missed anything.

JBR90 (approved):
Great work Neil, so good to get the inngest out

JBR90 (approved):

packages/core/src/models/transcript.ts
JBR90: Yeah, I thought that was probably the case. Just double checking.

mikeritson-oak (approved):
I have tested all notification functions as detailed, plus also some general checks around Aila. No issues seen, looks good. 👍

PR #1021, opened 2026-04-22
fix: [AI-2148] tighten Agentic-mode error handling

neil-hughes:
## Description 

This PR tightens Agentic Aila error handling ahead of rollout, while preserving partial progress when some section updates have already been applied.

The main change is to distinguish between:

- true hard failures, where no section updates were successfully applied
- partial successes, where one or more section updates were applied before a later failure

With this change:

- zero-progress agentic turns now fail cleanly and do not persist
- turns with applied section updates now persist those updates and finish with a fallback user-facing message if a later step fails
- failed agentic turns emit an explicit failed-turn marker for frontend handling and tracking
- successful partial turns continue through the normal persistence path

This PR also makes the agentic turn flow easier to follow by introducing explicit phase outcomes and refactoring plan-step execution.

Separately, it fixes a bug in `serverSideFeatureFlag()` where cached boolean true values were being treated as false on cache hits. That fix is shared infrastructure and is not specific to Agentic Aila.

## Production impact

The agentic turn-handling changes remain behind the Agentic Aila path and are not active in the current production non-agentic chat flow.

The only shared runtime change is the server-side feature flag cache fix.

## Other changes

- Adds fault-injection hooks and manual-testing notes for agentic failure paths so engineers can test failures locally
- Makes failed-turn detection use the parsed protocol comment instead of string matching
- Suppresses completion tracking for true failed agentic turns while allowing persisted partial successes
- chore: Makes Prisma query logging opt-in via `LOG_DB_QUERIES=true`

## Testing

Validated with targeted Jest suites for:

- agentic turn execution and stream completion
- failed-turn frontend/store handling
- lesson-plan tracking behavior
- server-side feature flag cache regression

## Testing Plan

No code changes have been made to non-agentic Aila, so the only testing requirement for this PR is to verify that non-Agentic Aila continues to function as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,8 @@ The `apps/nextjs` and `packages/aila` projects have Jest tests. Run all tests us
 pnpm test
 ```
 
+For local agentic failure-injection commands, see [packages/aila/src/lib/agentic-system/MANUAL_TESTING.md](packages/aila/src/lib/agentic-system/MANUAL_TESTING.md).
+
 ### End-to-end tests
 
 Ensure the dev server is running with `pnpm dev`. Use the Playwright UI to select and run individual tests interactively.
--- a/apps/nextjs/src/stores/chatStore/__tests__/messageStates.ts
+++ b/apps/nextjs/src/stores/chatStore/__tests__/messageStates.ts
@@ -1,5 +1,14 @@
 /* eslint-disable @cspell/spellchecker */
-export const createMessageStates = (fixedDate: Date) => ({
+import type { AiMessage } from "../types";
+
+type MessageStates = {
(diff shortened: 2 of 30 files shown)


assadk88 (approved):
LGTM 👍

mikeritson-oak (approved):
I've ran through the basic tests that we do for every release candidate - sign-up, loging, lesson plan, teaching materials, exports, etc. - and all worked as expected.
If merged this PR will also be re-checked during the release testing as well.
Looks good to merge. 👍

neil-hughes:
Thanks for finding time for this one @mikeritson-oak 🙌

PR #1024, opened 2026-04-23
chore: remove unused final state - AI-2151

JBR90:
## Description

Details in - https://www.notion.so/oaknationalacademy/Decide-whether-the-client-should-reconcile-with-the-final-server-state-34826cc4e1b180b5ac3aec412c1e7a29

> The agentic backend emits a final full lesson state, but the client does not appear to use it. 

In the lesson plan store we refetch the lesson plan and update from the database every time messageFinished is called. This action is called in onFinish ( aisdk useChat). We can remove the unused client passing function and ‘final’ state as we sync with database after every message.

```
 messageFinished: () => {
        log.info("Message finished");
        set({ isAcceptingChanges: false, ...initialPerMessageState });
        void handleRefetch(set, get, trpcUtils)();
      },

```

-remove final state from `ailaTurnCallbacks.ts`
-remove unused `findLatestServerSideState`

- useDemoLocking was using ` type: "state" , reasoning: "final" `to work out when a lesson had finish to limit the amounts of edits after lesson completion. This was broken as it was treating i`sLoading` as a boolean rather than streaming statuses. It was returning early and never limited the edits for demo users. This has been refactored to replay the lesson by looping through patches and extract lessons parts to determine a completed lesson before counting edits.

## How to test

- Full lesson flow works as before
- demo users have a limit to edits after lesson completion (3)
image

## follow ups
- Investigate conditional refresh rather than fetching after every message

## Checklist

### DIFF
--- a/apps/nextjs/src/hooks/useDemoLocking.test.ts
+++ b/apps/nextjs/src/hooks/useDemoLocking.test.ts
@@ -1,0 +1,269 @@
+import type { PartialLessonPlan } from "@oakai/aila/src/protocol/schema";
+
+import type { AiMessage } from "@/stores/chatStore";
+
+import {
+  findFirstCompleteAssistantMessage,
+  replayLessonPlanFromMessages,
+} from "./useDemoLocking";
+
+const fixedDate = new Date("2023-01-01T12:00:00.000Z");
+
+// Helper to create a message with patches
+const createAssistantMessageWithPatches = (
+  id: string,
+  patches: Array<{ op: "add" | "replace"; path: string; value: unknown }>,
+): AiMessage => {
+  const patchContent = patches
+    .map(
+      (patch) =>
+        `{\n  "type": "patch",\n  "reasoning": "test",\n  "value": ${JSON.stringify(patch)},\n  "status": "complete"\n}`,
+    )
+    .map((patch) => `␞\n${patch}\n␞`)
+    .join("\n");
+
+  return {
+    id,
+    role: "assistant",
+    content: patchContent,
+    createdAt: fixedDate,
+  };
+};
+
+const createUserMessage = (id: string, content: string): AiMessage => {
+  return {
+    id,
+    role: "user",
+    content,
+    createdAt: fixedDate,
+  };
+};
+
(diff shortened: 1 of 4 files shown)


apps/nextjs/src/hooks/useDemoLocking.test.ts
```diff
@@ -0,0 +1,269 @@
...
+        messages,
+        isLessonCompleteMock,
+      );
+
+      expect(result?.id).toBe("a1");
+      // Verify that a2 is ignored
+      expect(result?.id).not.toBe("a2");
+    });
+  });
+});
```
neil-hughes, line 470: Great to see this hook getting tested, but could we add some hook-level tests too? e.g.
- Demo user, no complete lesson: returns false
- Demo user, first complete message found, fewer than `NEXT_PUBLIC_DEMO_MESSAGES_AFTER_COMPLETE` later user messages: returns false.
- Same setup, at threshold: returns true.
- Non-demo user: returns false.

apps/nextjs/src/hooks/useDemoLocking.ts
```diff
@@ -32,56 +18,106 @@ const DEMO_MESSAGES_AFTER_COMPLETE = parseInt(
...
+ * are complete, where N is NEXT_PUBLIC_DEMO_MESSAGES_AFTER_COMPLETE.
+ *
+ * Completeness is determined by replaying patches from stable messages in order
+ * and checking when isLessonComplete flips to true for the first time.
+ */
+export function useDemoLocking() {
+  const demo = useDemoUser();
+  const stableMessages = useChatStore((state) => state.stableMessages);
 
   if (!demo.isDemoUser) {
```
neil-hughes, line 111: The rest of this function after the early return is quite expensive - replaying/parsing all assistant patches to find the first complete message and counting user messages afterwards - and this happens on every single render.

We could add a `useMemo` here keyed on `stableMessages` to wrap the rest of the function from line 104 onwards so this only happens once for a given set of data?

neil-hughes (changes requested):

neil-hughes (approved):

mikeritson-oak (approved):
Tested for standard accounts and demo accounts, looks good. 👍

PR #1026, opened 2026-04-27
fix: cache lesson plan exports per content guidance

assadk88:
## Description

The fix removes regressions which introduced by PR #1010, raised on Sentry: [1](https://oak-national-academy.sentry.io/issues/114173238/events/b8ab5f6336494b31bc28ab78fc2ca96c?environment=production&project=4507089597300816&referrer=alert-rule-issue-list) [2](https://oak-national-academy.sentry.io/issues/114286617/events/cf330b13655f49bbab8da1981acdbe50?environment=production&project=4507089597300816&referrer=alert-rule-issue-list&seerDrawer=true)

PR #1010 wrote `expiredAt` on existing `lesson_exports` rows to invalidate the cache when content guidance changed. But `aila-download` filters `expiredAt IS NULL` because that column means "the Drive file has been deleted", breaking any download link a teacher already had loaded.

This PR introduces changes which ensure the lesson snapshot hash now folds in a string derived from active moderations. Existing rows are left untouched, so previously generated links continue to work until the daily cleanup cron retires them after 14 days.

These changes are only scoped to lesson plan exports. Slides, worksheets, quizzes, and additional materials don't render content guidance, so they are unaffected.

## Issue(s)

[AI-2159](https://www.notion.so/oaknationalacademy/Lesson-plan-export-cache-invalidation-bug-34b26cc4e1b180259f97cc5c7fa53570?v=c76315f391b74cf2bd6722b8bdcbc3e6&source=copy_link)

## How to test

Go to https://oak-ai-lesson-assistant-website-glbrddbm2.vercel.thenational.academy. 

### 1. Chat with no moderation should result in normal cache hit

1. Open a chat with a finished lesson, no active moderations.
2. Generate the lesson plan and click "Download lesson plan (.docx)", download the file and note the Drive URL (**URL-A**).
    - Right-click "Download lesson plan (.docx)" and copy the link address. Paste it somewhere this is URL-A. The important part is fileId=<…> in the URL; that's what changes when a fresh export is generated.
4. Click it again immediately, it should return the same URL-A download instantly (check the URL).

This demonstrates the same URL is present on both clicks and that the link works. Take note of URL-A.

### 2. Moderation presence should invalidate the cache and result in a fresh download link

1. Go back to the same chat, send a refinement message containing the magic string `oak-sen`. A content guidance banner should appear in the chat once Aila has finished.
2. Return to the download page (hard refresh by pressing Cmd+Shift+R.) and click "Download lesson plan (.docx)" > note the new URL (**URL-B**). It should differ from URL-A because the cache key now includes guidance.
3. Click it again, it should resolve instantly to URL-B (cache hit on the guidance-keyed snapshot).
5. Open URL-B in a fresh tab; it should download. The document should contain a content guidance section.

Demonstrates that URL-B is distinct from URL-A and repeat clicks return the same URL. Take note of URL-B.

### 3. Changing moderation does not break old links

This test is key as it shows that PR #1010's regression is fixed: changing moderation must NOT break old download links.

1. With URL-A and URL-B saved from the previous tests, send another refinement message containing `oak-sen` plus a request that meaningfully changes the lesson (e.g. *"Add a learning cycle on regional differences. oak-sen"*).
2. Wait for Aila to update the lesson, then return to the download page (hard refresh).
3. Click "Download lesson plan (.docx)" > note the new URL (**URL-C**). It should differ from both URL-A and URL-B.
4. **Critical:** open URL-B in a fresh browser tab. It must still download.
6. Open URL-A in a fresh browser tab. It must still download.

URL-A and URL-B should both still resolve after URL-C is generated.

> @mikeritson-oak URL-A and URL-B  are replaced by URL-C, but the underlying Drive files still exist so direct links will still keep working. FYI the job of this fix is to ensure the export reflects the current lesson plan based on lesson content and moderation state, not to revoke previously distributed links. Cleaning them up is the cron's job responsibility (which retires Drive files older than 14 days on a daily basis). 

### 4. Other exports are unaffected

1. In the chat from Test 3 (active moderations present), click "Download lesson slides" > note the URL (**URL-S**).
2. Send another `oak-sen` message to keep moderation active.
3. Return to the download page (hard refresh) and click "Download lesson slides" again, it should resolve to the same URL-S.

Demonstrates that the slides exporter ignores moderation changes. This is correct, because slides docs don't render content guidance.

## Screenshots

N/A, backend fix only. No UI changes.

## Checklist

### DIFF
--- a/packages/api/src/export/exportHelpers.ts
+++ b/packages/api/src/export/exportHelpers.ts
@@ -25,9 +25,11 @@ function contentGuidanceCacheKey(
     return undefined;
   }
 
+  // Don't swap this for localeCompare — the output is hashed, and the sort
+  // needs to be the same on every machine.
   return categories
     .map((c) => `${c.code}:${c.severityLevel}`)
-    .sort()
+    .sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
     .join("|");
 }
 
--- a/packages/api/src/export/exportHelpers.ts
+++ b/packages/api/src/export/exportHelpers.ts
@@ -1,4 +1,4 @@
-import { LessonSnapshots } from "@oakai/core";
+import { LessonSnapshots } from "@oakai/core/src/models/lessonSnapshots";
 import type { Snapshot } from "@oakai/core/src/models/lessonSnapshots";
 import { Moderations } from "@oakai/core/src/models/moderations";
 import type { DisplayCategory } from "@oakai/core/src/utils/ailaModeration/severityLevel";
--- a/packages/api/src/export/exportHelpers.ts
+++ b/packages/api/src/export/exportHelpers.ts
@@ -1,5 +1,8 @@
 import { LessonSnapshots } from "@oakai/core";
 import type { Snapshot } from "@oakai/core/src/models/lessonSnapshots";
+import { Moderations } from "@oakai/core/src/models/moderations";
+import type { DisplayCategory } from "@oakai/core/src/utils/ailaModeration/severityLevel";
+import { getDisplayCategories } from "@oakai/core/src/utils/ailaModeration/severityLevel";
 import type {
   LessonExportType,
   LessonSnapshot,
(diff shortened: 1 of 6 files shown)


neil-hughes (approved):
Approach looks sound to me. There are a couple of rare edge cases that could lead to misalignment between the cache and the contents (e.g. if the content guidance text changes independently of the severity level, which I don't think will happen in practice) and we _could_ ensure that the `contentGuidanceCategories` passed to the renderer are sorted in a strict order too... but the risk is low in practice I think. This fixes the issue in production so we'll keep an eye on the potential for hardening the caching layer later. Nice one @assadk88

mikeritson-oak (approved):
Tested as per notes, all looks good. 👍

PR #1035, opened 2026-05-06
fix: retry Oak Moderation Service once

Dedekind561:
## Context

The intention behind this PR is to make the Oak Moderation Service safer to promote from shadow mode to the primary Aila moderator.

 The current primary inline OpenAI moderator already has a very lightweight resilience pattern: if the first moderation call throws, it logs the failure and tries once more. The Oak Moderation Service path did not have the equivalent protection, so a transient 5xx, timeout, or short-lived network failure would immediately fail the primary moderation step once we switch over.

This PR intentionally mirrors the existing inline moderator behaviour rather than introducing a broader retry framework. It keeps the change small and easy to reason about: call the Oak service once, log on failure, then retry once. If the second attempt also fails, the existing error handling still applies.

The goal is not to mask persistent moderation service failures, but to avoid user-facing failures for one-off transient errors during the primary moderator migration.

## Summary
- add a lightweight one-retry wrapper around Oak Moderation Service moderation calls
- keep Oak service error wrapping unchanged
- add a focused unit test covering retry after a transient failure

## Testing
- pnpm --filter @oakai/aila test -- --runTestsByPath src/features/moderation/moderators/OakModerationServiceModerator.test.ts
- pnpm --filter @oakai/aila type-check

### DIFF
--- a/packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
+++ b/packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
@@ -66,6 +66,7 @@ export class OakModerationServiceModerator extends AilaModerator {
       return await this._moderate(input, context);
     } catch (error) {
       log.error("Oak Moderation Service moderation error: ", error);
+      log.warn("Retrying Oak Moderation Service call");
       return await this._moderate(input, context);
     }
   }
--- a/packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
+++ b/packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
@@ -33,14 +33,10 @@ export class OakModerationServiceModerator extends AilaModerator {
     this.config = config;
   }
 
-  async moderate(
+  private async _moderate(
     input: string,
     context?: AilaModeratorContext,
   ): Promise<ModerationResult> {
-    log.info("Calling Oak Moderation Service", {
-      contentLength: input.length,
-    });
-
     try {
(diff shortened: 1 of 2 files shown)


neil-hughes (approved):
LGTM @Dedekind561 - just one suggestion, might it be worth adding an additional `log.warn` before the retry (line 57, OkaModerationServiceModerator.ts), just to make it clear in the logs that not only did an error occur but we are retrying?

packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
```diff
@@ -49,4 +45,17 @@ export class OakModerationServiceModerator extends AilaModerator {
...
+  async moderate(input: string): Promise<ModerationResult> {
+    log.info("Calling Oak Moderation Service", {
+      contentLength: input.length,
+    });
+
+    try {
+      return await this._moderate(input);
+    } catch (error) {
+      log.error("Oak Moderation Service moderation error: ", error);
+      return await this._moderate(input);
```
mantagen: Are we happy with just one retry?

Dedekind561: We're just bringing this in line with the current moderator behaviour - I'd suggest leaving as-is for now and if we need more defence then follow up afterwards. But the shadow deploy's got 0% error rate at the moment following Margaux's retry fix yesterday morning.

mantagen (approved):

PR #1036, opened 2026-05-06
fix: pass Aila moderation identifiers to Oak service

Dedekind561:
## Context

The underlying intent here is traceability across systems without introducing extra PII into the moderation service.

Aila already has its own chat/session records and message records, while the Oak Moderation Service creates its own moderation records.

The PR passes only the operational linkage identifiers the moderation service needs:

  - session_id: the Aila chat/session ID
  - message_id: the assistant message ID being moderated

It intentionally does not pass user_id, because that is PII and is not required for linking moderation records back to the Aila event/message context.

## Summary
- pass Aila chat/session and assistant message identifiers through moderation calls
- include only session_id and message_id in Oak Moderation Service context; user_id is intentionally not sent
- cover the Oak service request body and Aila moderator context propagation in tests

### DIFF
--- a/packages/aila/src/features/moderation/index.test.ts
+++ b/packages/aila/src/features/moderation/index.test.ts
@@ -6,11 +6,16 @@ import type { PrismaClientWithAccelerate } from "@oakai/db";
 import { AilaModeration } from ".";
 import { Aila } from "../../core/Aila";
 import type { Message } from "../../core/chat";
+import type { LLMService } from "../../core/llm/LLMService";
 import type { AilaPlugin } from "../../core/plugins";
 import type { AilaChatInitializationOptions } from "../../core/types";
 import type { PartialLessonPlan } from "../../protocol/schema";
 import type { AilaModerator } from "./moderators";
 import { MockModerator } from "./moderators/MockModerator";
+
+jest.mock("@oakai/db", () => ({
+  prisma: {},
+}));
 
 // Ensure that no tests start relying on prisma
 const prismaMock = {} as PrismaClientWithAccelerate;
@@ -38,9 +43,20 @@ const setUpModeration = ({
     document,
     chat,
     options: {
-      useModeration: true,
+      useModeration: false,
       usePersistence: false,
       useAnalytics: false,
+    },
+    services: {
+      chatLlmService: {
+        name: "mock-llm-service",
+        createChatCompletionStream: jest.fn(),
+        createChatCompletionObjectStream: jest.fn(),
+      } as unknown as LLMService,
+      chatCategoriser: {
+        categorise: jest.fn(),
+      },
+      oakModerator: moderator,
     },
     moderator,
     prisma: prismaMock,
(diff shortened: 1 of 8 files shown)


packages/aila/src/features/moderation/moderators/MockModerator.ts
```diff
@@ -13,7 +13,7 @@ export class MockModerator extends AilaModerator {
     this._mockedResults = results;
   }
 
-  async moderate(input: string): Promise<ModerationResult> {
+  async moderate(input: string, _context?: unknown): Promise<ModerationResult> {
```
neil-hughes: I think this `unknown` type should use the `AilaModeratorContext`, nullable for if we want to ignore it: `_context?: AilaModeratorContext`

packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
```diff
@@ -29,7 +33,10 @@ export class OakModerationServiceModerator extends AilaModerator {
     this.config = config;
   }
 
-  async moderate(input: string): Promise<ModerationResult> {
+  async moderate(
+    input: string,
+    context?: AilaModeratorContext,
```
neil-hughes, line 38: This one is correctly typed but is named inconsistently with the others which have an underscore.

packages/aila/src/features/moderation/moderators/OpenAiModerator.ts
```diff
@@ -168,7 +168,7 @@ export class OpenAiModerator extends AilaModerator {
     };
   }
 
-  async moderate(input: string): Promise<ModerationResult> {
+  async moderate(input: string, _context?: unknown): Promise<ModerationResult> {
```
neil-hughes: This one is missing the strong typing again: `_context?: AilaModeratorContext`

neil-hughes (changes requested):
This looks good @Dedekind561 although I think there's some inconsistency in the types/parameters used which would be good to tighten before we merge.

packages/aila/src/features/moderation/index.test.ts
```diff
@@ -38,10 +43,21 @@ const setUpModeration = ({
     document,
     chat,
     options: {
-      useModeration: true,
+      useModeration: false,
```
mantagen: Not clear why the test is now running with useModeration: false?

packages/aila/src/features/moderation/moderators/OakModerationServiceModerator.ts
Dedekind561: I've tightened up the typing on those other parameter declarations but left this as-is because `context` is being read in the function body

packages/aila/src/features/moderation/index.test.ts
Dedekind561: Yeah fixed that - think it was originally there to bypass reading some env var locally

PR #1049, opened 2026-05-18
feat(agentic): add per-section British English corrector agent

assadk88:
## Description

I've attempted to resolve the issue of Americanisms by introducing prompt-side prevention (`sectionAgents/shared/identityAndVoice.ts` and `plannerAgent.instructions.ts`) and per-section LLM correction via a new `britishEnglishCorrectorAgent`.

- Strengthens British English guidance in two prompts (`identityAndVoice.ts` for section agents, `plannerAgent.instructions.ts` for the planner).
- Adds a per-section corrector that runs between section-agent output and `commitStepUpdate`. After each section is generated, `AilaAmericanisms.findAmericanisms()` scans the output; if actionable spelling/phrasing are detected, a dedicated `britishEnglishCorrectorAgent` rewrites that section before the patch is emitted.
- Refactored scoring harness to wire into ailaTurn and collect stats directly (previous way parsed logs)

The corrector is a best-effort fallback. When it fails, it defaults to outputting the original section with Americanisms.

### Scoring results

Scoring results on the merged branch (`SCORE_RUNS=5`, 2 scenarios, 10 total runs) confirm the Americanism issue is resolved:

| Scorer | full-lesson | no-rag-lesson |
|---|---|---|
| `americanisms-spelling` | pass: 5/5 | pass: 5/5 |
| `americanisms-phrasing` | pass: 5/5 | pass: 5/5 |
| `americanisms-meanings` | flag: 5/5 (advisory, intentionally bypassed) | flag: 5/5 (advisory, intentionally bypassed) |

### Performance cost

@neil-hughes regarding your concerns around performance, the corrector agent adds one LLM call per section if Americanisms are flagged (in that section). The scoring harness measures this with:

  - 2 scenarios total (RAG and non-RAG)
  - 5 runs per scenario (`SCORE_RUNS=5`) = **10 runs total**
  - 3 turns per run (deterministic for these scenarios) = **30 turns total**
  - Each turn produces multiple sections = **111 total section across all runs**

The corrector fired on 3 of those 111 sections, or 2.7%. That translates to about 0.1 extra LLM calls per turn or 0.3 per complete lesson. All 3 firings were in learning-cycle sections (`cycle1`, `cycle2`, `cycle3`) of the no-RAG scenario; the RAG-enabled scenario was clean across all 5 runs. So the performance cost does seem negligible, but the two scenarios the scorer runs on may not be entirely representative of what's typically encountered in a production environment.

## Issue(s)

Fixes [AI-2150](https://www.notion.so/oaknationalacademy/Add-British-English-americanism-checks-to-the-agentic-flow-34826cc4e1b1801d8fb5df6f9256fc43?v=c76315f391b74cf2bd6722b8bdcbc3e6&source=copy_link)

## How to test

1. Run the agentic scoring harness:
   ```
   cd packages/aila && SCORE_RUNS=5 pnpm with-env npx jest --testMatch="**/scoring/scoreAgenticIssues.ts" --testTimeout=600000
   ```
`americanisms-spelling` and `americanisms-phrasing` should both report `pass: 5` per scenario. `americanisms-meanings` will still flag but these are advisory.

2. Verify CI's freshness check passes: `Agentic scoring report is up to date`.

3. Optional manual smoke: `pnpm dev`, create a lesson. Latency on the common path should feel unchanged. The corrector only fires on sections that flag Americanisms (~3% of sections in scoring scenarios). You can do this using a prompt that contains explicit Americanisms (e.g. recognize, eraser, etc.) to see if Aila reproduces or corrects them:

```
Create a Year 9 art and design lesson titled "Color Theory: How Painters Mix Their Palettes". In the learning outcome, mention that pupils will recognize the difference between primary and secondary colors, practice using an eraser to lighten pencil shading, and analyze how famous painters organized their work. Include four learning cycles, key learning points, vocabulary, and a six-question exit quiz.
```

## Screenshots

N/A. No UI change.

## Checklist

- [ ] Manually tested across browsers / devices: N/A (agentic backend only)
- [ ] Considered impact on accessibility: N/A (no UI)


packages/aila/src/features/americanisms/AilaAmericanisms.ts
```diff
@@ -13,24 +13,30 @@ import {
 
 export type AilaDocumentContent = Record<string, unknown>;
 
+// Phrases the translator dictionary flags as Americanisms but which are
+// correct British English in Aila's educational contexts — e.g. "gas" as a
+// state of matter, which the dictionary maps to "petrol". Filtered so the
+// corrector never fires on them and the scorers don't flag them.
+const FILTER_OUT_PHRASES: ReadonlySet<string> = new Set([
```
neil-hughes, line 20: Good idea. Do we want to add words like `line` and `triangle` here? (These both flag in maths lessons as "queue" and "setsquare" which is not intended!) There are probably some more but I imagine there's diminishing returns from adding too many filter words here.

packages/aila/src/lib/agentic-system/agents/britishEnglishCorrectorAgent/britishEnglishCorrectorAgent.instructions.ts
```diff
@@ -0,0 +1,18 @@
+import { britishEnglishRules } from "../shared/britishEnglishRules";
+
+export const britishEnglishCorrectorAgentInstructions = `# Identity
+You correct British English in a single section of an Oak National Academy lesson plan written for UK teachers and pupils.
```
neil-hughes: This prompt looks tight but I'm curious about if it's worth clarifying the opening? "You correct British English" reads ambiguously to me. Would something like this be more clear?

`"You rewrite a single section of an Oak National Academy lesson plan to remove Americanisms and use standard British English suitable for UK teachers and pupils."`

packages/aila/src/lib/agentic-system/agents/sectionAgents/shared/identityAndVoice.ts
```diff
@@ -15,6 +17,17 @@ You will be given a specific task and the current state of the lesson plan. You
...
-## Language
-Use British English spelling and vocabulary (e.g. colour not color, centre not centre, rubbish not trash) unless the user sets a different primary language. This reflects our UK teacher audience.
+## Language — British English (mandatory)
+Write **British English** throughout the lesson plan: spelling, vocabulary, and phrasing. This is mandatory because our audience is UK teachers and pupils. The only exception is if the user has explicitly set a different primary language for the lesson.
+
+${britishEnglishRules}
+
+### Self-check before emitting content
+Before returning your section, scan your output for these specific patterns and correct them:
+1. Any word ending in **-ize**, **-izes**, **-ized**, **-izing**, **-yze**, **-yzes** → change to the **-ise** / **-yse** form.
```
neil-hughes, line 28: This prompt seems solid to me but it's triggering the linter. Some of the lint warnings we can add to the `cspell` global list. I'm not sure of the cleanest way to resolve the partial words that we don't want to add as global exceptions - perhaps local exceptions can be added to this file?

packages/aila/src/lib/agentic-system/scoring/scores.yaml
```diff
@@ -14,8 +14,7 @@ summary:
     title-no-gerund:
       pass: 3
     americanisms-spelling:
-      flag: 1
-      pass: 2
+      pass: 3
```
neil-hughes, line 17: 🥳  Nice!

packages/aila/src/lib/agentic-system/execution/applyBritishEnglishCorrection.ts
```diff
@@ -0,0 +1,76 @@
...
+  content,
+  responseSchema,
+}: {
+  context: AilaExecutionContext;
+  sectionKey: SectionKey;
+  content: unknown;
+  responseSchema: z.ZodTypeAny;
+}): Promise<unknown> {
+  const detected = americanisms.findAmericanisms({ [sectionKey]: content });
+  const sectionIssues = detected[0]?.issues ?? [];
```
neil-hughes: This depends on the insertion order staying the same - if we change the architecture to look at multiple sections in future this could break. We could instead:

```
  const sectionIssues =
    detected.find((d) => d.section === sectionKey)?.issues ?? [];
```

packages/aila/src/lib/agentic-system/execution/executePlanSteps.ts
```diff
@@ -85,7 +86,15 @@ async function executeGenerateStep(
...
-  const validated = sectionSchema.parse(result.data);
+
+  const corrected = await applyBritishEnglishCorrection({
+    context,
+    sectionKey: step.sectionKey,
+    content: result.data,
+    responseSchema: sectionSchema,
+  });
+
+  const validated = sectionSchema.parse(corrected ?? result.data);
```
neil-hughes: Very minor but `applyBritishEnglishCorrection` already calls `safeParse` and returns parsed data so we could avoid reparsing here if `corrected` isn't null/undefined.

packages/aila/src/lib/agentic-system/agents/britishEnglishCorrectorAgent/britishEnglishCorrectorAgent.instructions.ts
```diff
@@ -0,0 +1,18 @@
...
+
+# Task
+You receive a single lesson-plan section as JSON, plus a list of Americanisms detected in that section with their British alternatives. Return the section corrected into British English, preserving structure and meaning.
+
+# Constraints
+- Do not change keys or structure. Return the same shape you received.
+- Do not change the educational meaning of the content.
+- Preserve case: "Recognize" → "Recognise", "RECOGNIZE" → "RECOGNISE".
+- Apply British English consistently across the section, not only to the listed phrases — if you notice closely related Americanisms in the same section, fix those too.
```
neil-hughes: I'm curious about the breadth of "if you notice.... fix those too" - is there a chance it could make broader changes than intended? I guess the instruction is fairly clearly intended to only mean Americanisms, but would "... convert those to British English too" be a tighter instruction?

packages/aila/src/lib/agentic-system/agents/britishEnglishCorrectorAgent/createBritishEnglishCorrectorAgent.ts
```diff
@@ -0,0 +1,70 @@
...
+        content: `Section key: ${sectionKey}`,
+      },
+      { role: "developer" as const, content: issuesBlock(issues) },
+      {
+        role: "user" as const,
+        content: `Original section JSON:\n${JSON.stringify(content)}`,
+      },
+    ],
+    modelParams: {
+      ...DEFAULT_AGENT_MODEL_PARAMS,
```
neil-hughes: This spread doesn't do anything, right? So it could just be `modelParams: DEFAULT_AGENT_MODEL_PARAMS`

.claude/skills/score-agentic/SKILL.md
```diff
@@ -13,10 +13,35 @@ The harness runs the agentic lesson generation pipeline multiple times and score
...
-cd packages/aila && SCORE_RUNS=3 pnpm with-env npx jest --testMatch="**/scoring/scoreAgenticIssues.ts" --testTimeout=600000
+cd packages/aila && DEBUG=ai:aila:agents SCORE_RUNS=3 pnpm with-env npx jest --testMatch="**/scoring/scoreAgenticIssues.ts" --testTimeout=600000 2>&1 | tee /tmp/score-agentic.log
 ```
 
-This updates `packages/aila/src/lib/agentic-system/scoring/scores.yaml`.
+This updates `packages/aila/src/lib/agentic-system/scoring/scores.yaml` and captures `aila:agents` debug logs (including the British English corrector firing counter) to `/tmp/score-agentic.log` for further analysis.
+
+## Summarise corrector firings
+
+After the harness completes, summarise how often the British English corrector fired:
```
neil-hughes: I really like adding the performance check to the harness output! Nice one.

.claude/skills/score-agentic/SKILL.md
```diff
@@ -29,8 +54,10 @@ packages/aila/src/lib/agentic-system/scoring/check-scoring-freshness.sh
...
 2. Run the scoring harness (command above)
 3. Review the report for regressions
-4. Commit `scores.yaml` alongside your code changes
+4. Review the corrector firing summary if you changed prompt guidance, the corrector, or `AilaAmericanisms` filters
+5. Commit `scores.yaml` alongside your code changes
 
 ## Configuration
 
 - `SCORE_RUNS=N` — number of runs per scenario (default 3). Higher values give more confidence but take longer.
+- `DEBUG=ai:aila:agents` — enables `aiLogger` output for the agentic system, including the corrector firing counter. Without this, the corrector summary above will be empty.
```
neil-hughes: Why do we need this additional variable to be set? Would we ever want to run the harness without it? I'm also not sure about the name `DEBUG=ai:aila:agents` for something that measures performance of the Americanisms corrector - can we make it more clear what it is, or just remove it if we would always like it switched on? (Or is this something that already exists and it's just being documented here now?)

neil-hughes (changes requested):
This looks really good! I've added some comments for you to consider, but overall this seems super solid. Great work.

packages/aila/src/lib/agentic-system/agents/sectionAgents/shared/identityAndVoice.ts
assadk88: I've added these to an inline ignore

packages/aila/src/lib/agentic-system/agents/britishEnglishCorrectorAgent/britishEnglishCorrectorAgent.instructions.ts
assadk88: I've tightened this

packages/aila/src/lib/agentic-system/execution/executePlanSteps.ts
assadk88: Good spot

.claude/skills/score-agentic/SKILL.md
assadk88: The scorer doesn't need `DEBUG`. This env var is only needed for routing the corrector's own `fired / skipped / threw / errored` log lines into the log file so the grep in the skill can count them and return a summary. If I remove `DEBUG` we'd lose log output and the firing summary, which was added to give a rough view of performance / LLM calls.

Tracking corrector firing was intended to be separate from the core scoring as I didn't think routine runs would need it. The idea was that you'd set `DEBUG` when you specifically want to use an agent to see how the corrector behaved (firing rate, which sections, any failures).

Also `aila:agents` isn't corrector-specific. It's the pre-existing logger namespace for the agentic system. Happy to either (a) tighten the `SKILL.md` wording to spell out that the namespace covers the whole agentic system and that `DEBUG` isn't required for scoring, or (b) move the corrector logs to a dedicated `aila:agents:corrector` sub-namespace so the env var reads more specifically (`DEBUG=ai:aila:agents:corrector`)?

neil-hughes: Ah, I understand that more now. I personally think that adding a more detailed scoring summary is useful and there's no harm in it being generated every time without needing a variable.

I think if we keep the env var it would be good to rename it to something like `DEBUG_AGENTS` or similar, partially for clarity, but mainly because `DEBUG` has a high chance of clashing with other env vars that are also named `DEBUG` in any third party dependency we use.

Or maybe even call it `OUTPUT_SUMMARY=corrector,some_other_agent,some_other_other_agent`? 

But I feel like having a detailed summary makes the harness more useful in general, so my suggestion is to either:
1. make it default, remove the env var, and then we can choose to add helpful output from other agents to the summary in future
2. rename the env var itself to `OUTPUT_SUMMARY`

Do feel free to ignore or push back if you think these are bad ideas or I'm missing something!

assadk88: I think you're right, it might be sensible to make it the default. This would mean moving the firing summary out of the skill, tracking stats directly and outputting them as a `corrector:` block in the `scores.yaml` file. Is that what had in mind? 

If so I'm thinking of the block taking either of the following shape:

```
corrector:
  fired: 1
  skipped: 33
  failed: 0
  firings:
    cycle3: 1
```

alternatively we could also include the fire and failure rates:

```
corrector:
  fired: 1
  skipped: 33
  failed: 0
  fire_rate: "3.0%"
  failure_rate: "0.0%"
  firings:
    cycle3: 1
```

Sound about right?

neil-hughes: That seems like an even cleaner idea - definitely agree it's the best approach. Of the two, I think the one with the percentages is better, although I think we should make sure the labels are super clear - in a few weeks time I won't remember what "fire_rate" and "failure_rate" mean. Can we call it (just in the scores.yaml) "americanisms_corrector", and make sure each sub-label is as obvious at a glance as is reasonable?

assadk88: Sounds good!

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -142,7 +144,8 @@ export class AilaStreamHandler {
 
       if (this._chat.aila.options.useAgenticAila) {
         await this.span("start-agent-stream", async () => {
-          agenticTurnOutcome = await this.startAgentStream();
+          const outcome = await this.startAgentStream();
+          agenticTurnStatus = outcome.status;
```
neil-hughes: Why are we duplicating the state here - is it not cleaner just to declare an `agenticTurnOutcome` outside of the try and access the `.status` property on it whenever we need a status? This direct status variable feels like a leftover artifact of the old structure that hasn't been cleanly updated to the new shape for turn outcomes.

neil-hughes (changes requested):

packages/aila/src/core/chat/AilaStreamHandler.ts
```diff
@@ -194,9 +195,6 @@ export class AilaStreamHandler {
         this._chat.id,
       );
     } catch (e) {
-      if (this._chat.aila.options.useAgenticAila) {
```
neil-hughes, line 209: I left a comment already but github seems to have swallowed it so putting it again - sorry if it gets duplicated somewhere!

I don't understand why these three lines were removed as part of the refactoring cleanup - it seems like a logic change to no longer mark an error as failure.

(Sidenote: I have an upcoming ticket to refactor this file so don't worry too much about getting it all perfect, but changing the logic here does seem like a potential regression, unless I'm missing something.)

neil-hughes (changes requested):

packages/aila/src/core/chat/AilaStreamHandler.ts
assadk88: The local var now holds the whole outcome object instead of just a status string. I change this in response to your earlier comment which made the catch block assignment redundant because the `shouldComplete` check further on collapses "no outcome" and "failed outcome" to false, and chat.complete() is skipped just as before.

neil-hughes (approved):
Looks good to me @assadk88 - really great work! As it's a fairly large diff I'm going to tag in Copilot for a final scan just in case there's any final minor polish worth fixing. I think we're good to start testing it and you can move onto something new 🥳

packages/aila/src/lib/agentic-system/execution/applyBritishEnglishCorrection.ts
```diff
@@ -0,0 +1,84 @@
...
+  context,
+  sectionKey,
+  content,
+  responseSchema,
+}: {
+  context: AilaExecutionContext;
+  sectionKey: SectionKey;
+  content: unknown;
+  responseSchema: z.ZodTypeAny;
+}): Promise<unknown> {
```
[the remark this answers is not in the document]
assadk88: @neil-hughes so I went with the generic (`<S extends z.ZodTypeAny>` + `Promise<z.infer<S> | null>`) rather than the literal `Promise<unknown | null>`. FYI TS collapses `unknown | null` to just `unknown`, so that suggestion just seems to be syntactic sugar? The generic I added gives a signature improvement (corrected/validated typed as the section union instead of `unknown`), but the `draft as Record<string, unknown>` cast widens it back two lines later (modifying that is a bit involved). So this change is only a marginal gain for the function signature. Happy to revert to `Promise<unknown>` if you'd rather skip the eslint-disable, or open to other suggestions?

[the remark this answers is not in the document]
neil-hughes: I think this comment from Copilot is ultimately not that important and shouldn't lead to any rabbit holes - your time would be much better spent on other tickets. Your generic version of this function looks like an improvement on the function signature so nice one, let's keep that and move on.

mikeritson-oak (approved):
I've taken a look over this, ran some test lesson and had no problems. Speed of generation was normal and any possible Americanisms within the lesson were correctly changed to British English.

PR #1052, opened 2026-05-19
fix: [AI-2162] stop empty moderation requests and fix stream completion edge cases

neil-hughes:
## Description

- **Prevents empty moderation requests**: `performModeration()` now skips the moderation service call when `document.content` is `{}`. The service was receiving empty requests and responding invalidly. Mock override codes (`oak-hs`, `oak-tox`) are unaffected as they are checked before this guard.

- **Prevents non-agentic stream errors calling `complete()`**: `generationFailed()` is now called from the error path, ensuring `complete()` (and moderation) don't run after a stream failure

- **Stops client abort being persisted as `SUCCESS`**: `readFromStream` now sets `skipCompletion = true` immediately after a client aborts (e.g. by closing a tab during a stream), preventing an empty document from being persisted

- **Prevents unhandled `controller.error()` rejection**: The `.catch()` in `startStreaming` could throw`ERR_INVALID_STATE` if the controller was already closed by the `finally` block - added a try/catch to prevent this behaviour.

- **Agentic silent failure**: When `ailaTurn` throws (rather than returning `{ status: "failed" }`), the client received no error message. Now one is enqueued explicitly.

## Automated test plan

- [X] `pnpm --filter @oakai/aila test -- --testPathPatterns="AilaStreamHandler|moderation/index"` — 23 tests pass

## Issue(s)

Fixes #AI-2162

## How to test

1. No manual test plan for the specific changes. All fixes cover rare error paths that are not reproducible through the UI. Verification is via the unit test suite (23 tests) and code review. 
2. We should verify the happy path is unaffected - Aila lesson generation should operate as usual for typical scenarios.

### DIFF
--- a/packages/aila/src/core/chat/AilaStreamHandler.test.ts
+++ b/packages/aila/src/core/chat/AilaStreamHandler.test.ts
@@ -245,6 +245,21 @@ describe("AilaStreamHandler", () => {
     });
   });
 
+  it("enqueues an error message when ailaTurn throws (agentic)", async () => {
+    mockedAilaTurn.mockRejectedValue(new Error("ailaTurn threw unexpectedly"));
+
+    const chat = createMockChat({ useAgenticAila: true });
+    const handler = new AilaStreamHandler(chat as never);
+
+    await consumeStream(handler.startStreaming());
+
+    expect(chat.complete).not.toHaveBeenCalled();
+    expect(chat.enqueue).toHaveBeenCalledWith({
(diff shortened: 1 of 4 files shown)


mikeritson-oak:
Smoke tested on the preview branch across multiple lesson generations, no issues seen. 👍

mikeritson-oak (approved):

JBR90 (approved):

PR #1059, opened 2026-05-21
feat: [AI-2214] add dev/staging debug overlay

neil-hughes:
## Description

Adds a collapsible debug overlay to the Aila chat page, visible only in dev/staging and only to admin users. 

The primary motivation is making it easy to confirm whether agentic mode is active for your session without digging into network traffic or server logs.

To support this, `serverSideFeatureFlag` has been moved from apps/nextjs/src/utils into packages/api/src/utils, and a new `getAgenticAilaEnabled` helper wraps it with the flag name and production guard. This is now the single source of truth used by both the chat handler and the new debug tRPC router.

No changes to existing tests, stories, or component props — the streaming status test hook is preserved as a screen-reader-only element so test IDs keep working for e2e tests.

## Issue(s)

Fixes #AI-2214

## How to test

0. All Playwright e2e and standard unit tests should pass
1. Go to https://oak-ai-lesson-assistant-website-ec8ldurtd.vercel.thenational.academy
2. As an admin account, click on Aila chat
3. You should see a new debug interface 
4. Click to expand/hide the main body of the overlay
5. Log in as a non-admin account and verify that the overlay doesn't appear

## Screenshots

Screenshot 2026-05-21 at 14 51 15

Screenshot 2026-05-21 at 14 51 20


mikeritson-oak (approved):
This works as expected as per the ticket and the PR. If we were going to do any more iterations of this work I would have some suggestions for improvements/changes, but for now this is good to go. 👍

neil-hughes:
> This works as expected as per the ticket and the PR. If we were going to do any more iterations of this work I would have some suggestions for improvements/changes, but for now this is good to go. 👍

@mikeritson-oak Would definitely like to hear suggestions - this was just a quick initial version to make sure the upcoming agentic testing is simpler. We can certainly improve to it as we go.

JBR90 (approved):
Looks good to me, I wonder if it we be useful to expand this to active feature flags? Not for this PR tho

PR #1074, opened 2026-06-10
feat: configure moderation shadow independently

Dedekind561:
## Summary

Implements Phase 0 of the [candidate moderator shadowing plan in Aila](https://github.com/oaknational/oak-ai-moderation-service/blob/docs/candidate-moderator-shadowing-spec/docs/superpowers/specs/2026-06-09-candidate-moderator-shadowing-design.md).

- keeps primary moderator selection controlled by the existing `OAK_MODERATION_V1_PRIMARY` flag
- decouples shadow moderator selection from primary moderator selection
- adds explicit candidate-shadow env vars: `MODERATION_SHADOW_ENABLED`, `MODERATION_SHADOW_API_URL`, and `MODERATION_SHADOW_API_BYPASS_SECRET`
- stops automatically wiring the legacy inline `OpenAiModerator` as a shadow when Oak Moderation Service is primary
- adds focused factory tests for primary selection and the new shadow env combinations

## Behaviour change / subtlety

Before this change, Aila used a paired primary/shadow setup:

- `OAK_MODERATION_V1_PRIMARY=true`: primary was Oak Moderation Service, shadow was the legacy inline `OpenAiModerator`
- `OAK_MODERATION_V1_PRIMARY!=true`: primary was the legacy inline `OpenAiModerator`, shadow was Oak Moderation Service

After this change, primary and shadow are independent:

- `OAK_MODERATION_V1_PRIMARY=true`: primary is Oak Moderation Service
- `OAK_MODERATION_V1_PRIMARY!=true`: primary is still the legacy inline `OpenAiModerator`, so we can still switch production back to the old inline moderator if needed
- shadow is absent by default
- shadow is only created when `MODERATION_SHADOW_ENABLED=true` and `MODERATION_SHADOW_API_URL` is set
- that explicit shadow is always an Oak Moderation Service URL, intended for a candidate preview deploy

So this does not remove the inline moderator as a possible primary fallback. It removes the old automatic pairing where the inline moderator was shadowed by default, and it removes the implicit ability to toggle into "inline primary + production Oak service shadow" just by changing `OAK_MODERATION_V1_PRIMARY`. To shadow an Oak service now, configure the explicit `MODERATION_SHADOW_*` env vars.

## Verification

- `corepack pnpm --filter @oakai/aila test-file -- src/core/AilaFeatureFactory.test.ts`
- `corepack pnpm exec eslint packages/aila/src/core/AilaFeatureFactory.ts packages/aila/src/core/AilaFeatureFactory.test.ts`

Package type-check was attempted with `corepack pnpm --filter @oakai/aila type-check`, but it failed on existing/worktree dependency type-resolution issues outside this change.


mantagen (approved):
Looks good to me.

Would it be possible to add instructions for how QA would test once it is release candidate, for instance?

Dedekind561:
## QA instructions

Once this PR is merged, 
just a few things to check for on the Aila release candidate deploy:
Try planning a lesson on the chat workflow and then:
- [ ] Aila still consumes the Oak Moderation Service 
Aila logs on Vercel should _still_ show:
Screenshot 2026-06-15 at 15 12 55

- [ ] Double-check moderation requests are still being sent to
[Braintrust Moderation service production project](https://www.braintrust.dev/app/Oak%20National%20Academy/p/Moderation%20-%20production/logs)

- [ ] The shadow is off
Can no longer see this in the Aila Vercel logs on `/api/chat`: 
Screenshot 2026-06-15 at 15 22 28

- [ ] No changes in the user facing behaviour in the chat worklow on Aila

@mikeritson-oak

mikeritson-oak:
This work was checked previously, as per Mitch's testing comments and all was as expected. Aila was processing messages and moderations, Vercel logs showed relevant entries and Braintree logs showed all moderations.

mikeritson-oak (approved):

PR #1078, opened 2026-06-16
feat: agentic e2e tests

JBR90:
## Description

  Adds end-to-end test coverage for the agentic lesson generation pipeline, mirroring the existing megaprompt fixture system.

  - FixtureOpenAIProxy — wraps the OpenAI client via JS Proxy to intercept responses.parse() calls, recording or replaying numbered JSON fixtures per agent turn
  - getAgenticFixtureConfig — reads x-e2e-aila-system and fixture headers, gated behind AILA_FIXTURES_ENABLED to prevent header abuse in production
  - chatHandler.ts — uses x-e2e-aila-system to override pipeline selection in tests, bypassing the feature flag DB lookup; skips the old FixtureReplayLLMService when in agentic mode to
  avoid ENOENT
  - New persona modify-lesson-plan-agentic — gives the modify-lesson agentic spec its own isolated user so it can run in parallel with the megaprompt spec
  - Three new E2E specs: full Romans flow (desktop + mobile), modify lesson + add materials

  Testing instructions

  Existing megaprompt E2E tests are unaffected — applyLlmFixtures defaults to ailaSystem: "megaprompt".

  The new @agentic tagged tests will run automatically in CI against the preview build in replay mode using the committed fixtures in tests-e2e/recordings/agentic/.

  To re-record fixtures, change FIXTURE_MODE to "record" in the relevant test file and run against a local dev server with a real OpenAI key. Switch back to "replay" and commit the updated
  JSON files before pushing.
## Issue(s)

Fixes #

## How to test

1. Go to https://oak-ai-lesson-assistant-website-izruy5h7r.vercel.thenational.academy
e2e tests for agentic passing, check report below 

## Checklist


apps/nextjs/src/app/api/chat/chatHandler.ts
```diff
@@ -84,7 +85,27 @@ async function setupChatHandler(req: NextRequest) {
...
+        ? e2eAilaSystem === "agentic"
+        : await getAgenticAilaEnabled();
+
+      log.info(
+        "Chat handler: e2eAilaSystem=%s useAgenticAila=%s",
+        e2eAilaSystem ?? "null",
+        String(useAgenticAila),
+      );
+
+      const agenticFixture = getAgenticFixtureConfig(req.headers);
```
neil-hughes, line 110: May be tempted to add a check here that e2e fixture mode has the data needed, something like:

```
const agenticFixture = getAgenticFixtureConfig(req.headers);

if (
  process.env.AILA_FIXTURES_ENABLED === "true" &&
  e2eAilaSystem === "agentic" &&
  !agenticFixture
) {
  throw new Error(
    "Agentic E2E fixture mode requires x-e2e-fixture-name and a valid x-e2e-fixture-mode.",
  );
}
```

apps/nextjs/src/app/api/chat/fixtures/index.ts
```diff
@@ -86,3 +87,33 @@ export function getFixtureOakModerator(headers: Headers) {
...
+
+  if (headers.get("x-e2e-aila-system") !== "agentic") {
+    return undefined;
+  }
+
+  const fixtureMode = headers.get("x-e2e-fixture-mode");
+  const fixtureName = headers.get("x-e2e-fixture-name");
+
+  if (!fixtureName || (fixtureMode !== "record" && fixtureMode !== "replay")) {
+    return undefined;
```
neil-hughes: This feels a little dangerous - would it be better to throw on an invalid fixture?

Something like

```
if (!fixtureName) {
  throw new Error("Missing agentic fixture name");
}

if (fixtureMode !== "record" && fixtureMode !== "replay") {
  throw new Error(`Invalid agentic fixture mode: ${fixtureMode ?? "missing"}`);
}

if (
  fixtureName.startsWith("/") ||
  fixtureName.includes("..") ||
  fixtureName.includes("\\")
) {
  throw new Error(`Invalid agentic fixtureName: ${fixtureName}`);
}
```

apps/nextjs/src/app/api/chat/fixtures/index.ts
```diff
@@ -86,3 +87,33 @@ export function getFixtureOakModerator(headers: Headers) {
...
+  }
+  if (
+    fixtureName.startsWith("/") ||
+    fixtureName.includes("..") ||
+    fixtureName.includes("\\")
+  ) {
+    log.warn("Ignoring invalid agentic fixtureName: %s", fixtureName);
+    return undefined;
```
neil-hughes: See above

neil-hughes (changes requested):

PR #1069, opened 2026-06-04
feat(aila): add feedback link beside the beta tag

assadk88:
## Description

Adds a feedback link next to the Beta label on the Aila starter and lesson pages. The top nav is also updated to match the new design.

- `beta-tag.tsx`: added a `BetaTagWithFeedback` component (the Beta pill plus the feedback link) and a shared `AILA_FEEDBACK_FORM_URL` constant. The link opens the feedback form in a new tab.
- `aila-start/index.tsx` (starter page): moved the Beta tag into the page body with the feedback link, and added the `Aila, Oak's AI lesson assistant` heading and subheading.
- `chat-start.tsx` (lesson page): added the Beta tag and feedback link above the "Hello" greeting.
- `header.tsx`: removed the Beta tag from the nav and added the `Oak's AI lesson assistant` tagline with a divider beside the logo.

The link points to the existing Oak feedback Google form.

## Issue(s)

Fixes [AI-2218](https://app.notion.com/p/oaknationalacademy/Beta-feedback-link-36626cc4e1b18017997ff9ca401deac0?v=c76315f391b74cf2bd6722b8bdcbc3e6)

## How to test

1. Go to https://oak-ai-lesson-assistant-website-18tf9xvrz.vercel.thenational.academy/aila
2. The nav should read `Aila | Oak's AI lesson assistant`, and a `Beta` pill with "Feedback will help us improve Aila" should sit above the `Aila, Oak's AI lesson assistant` heading.
3. Click the feedback link. The Oak feedback form should open in a new tab.
4. Go to https://oak-ai-lesson-assistant-website-dfocxxgk3.vercel.thenational.academy/aila/lesson. The same `Beta` pill and feedback link should sit above the "Hello, {name}" greeting.
5. Narrow the window to mobile width. The tagline drops out of the nav; the Beta pill and feedback link stay.

## Screenshots

How it used to look (delete if n/a):
image
image

How it should now look:

image
image

## Checklist

- [ ] Manually tested across browsers / devices
- [ ] Considered impact on accessibility (new heading is an `h1`; link has descriptive text)
- [ ] ~~Does this PR update a package with a breaking change~~ (no package changes)

PR #1079, opened 2026-06-16
fix(aila): point beta feedback link to hubspot form

assadk88:
## Description

Moves the Aila feedback link to the live HubSpot survey, and makes the URL configurable instead of hard-coded.

- `beta-tag.tsx`: the shared `AILA_FEEDBACK_FORM_URL` constant now reads from a `NEXT_PUBLIC_AILA_FEEDBACK_FORM_URL` env var, falling back to the HubSpot survey URL if the var is unset. The link can then be changed in Doppler (plus a redeploy) rather than via a code change.
- `faqs/index.tsx`: pointed the three FAQ feedback links at the same `AILA_FEEDBACK_FORM_URL` constant, replacing the old Google form they still used.

The original feedback link (added in #1069, now part of this base branch) always pointed at the Google form as a stopgap, with the plan to switch to HubSpot once Kath's form was set up. That form is now published and tested.

## Issue(s)

Fixes [AI-2218](https://app.notion.com/p/oaknationalacademy/Beta-feedback-link-36626cc4e1b18017997ff9ca401deac0?source=copy_link)

## How to test

1. Check out this branch and run the app.
2. Go to `/aila` and `/aila/lesson`.
3. The Beta pill and "Your feedback will help us improve Aila" link sit above the heading / greeting as before.
4. Click the feedback link. The **HubSpot** survey should open in a new tab (not the Google form).
5. Go to `/faqs`. Three feedback links there now point at the same HubSpot survey: "Your feedback is essential" (in the beta-phase intro), "feedback or suggestions" (under "Can I provide feedback or suggest improvements for Aila?"), and "give us your feedback!" (under "Can users suggest features for future updates?"). Click each and confirm it opens the HubSpot survey, not the Google form.

## Checklist

- [x] Manually tested across browsers / devices
- [x] Considered impact on accessibility

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
@@ -7,7 +7,7 @@ import {
 } from "@oaknational/oak-components";
 
 export const AILA_FEEDBACK_FORM_URL =
-  "https://docs.google.com/forms/d/e/1FAIpQLSf2AWtTtr4JISeMV4BY5LCMYhDFPz0RPNdXzmy_vjk4BmM69Q/viewform";
+  "https://survey.hsforms.com/184ua72b5RtOD9SDPq4BKKQbvumd";
 
 export function BetaTagWithFeedback({
   feedbackHref,
--- a/apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
@@ -7,6 +7,7 @@ import {
 } from "@oaknational/oak-components";
 
 export const AILA_FEEDBACK_FORM_URL =
+  process.env.NEXT_PUBLIC_AILA_FEEDBACK_FORM_URL ??
   "https://survey.hsforms.com/184ua72b5RtOD9SDPq4BKKQbvumd";
 
 export function BetaTagWithFeedback({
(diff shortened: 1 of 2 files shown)


apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from "@oaknational/oak-components";
 
 export const AILA_FEEDBACK_FORM_URL =
-  "https://docs.google.com/forms/d/e/1FAIpQLSf2AWtTtr4JISeMV4BY5LCMYhDFPz0RPNdXzmy_vjk4BmM69Q/viewform";
+  "https://survey.hsforms.com/184ua72b5RtOD9SDPq4BKKQbvumd";
```
JBR90, line 11: I'm wondering whether this should be stored in Doppler. Then any future changes can be made there

JBR90 (changes requested):

apps/nextjs/src/components/AppComponents/Chat/beta-tag.tsx
assadk88: Good shout, done. The URL now reads from a NEXT_PUBLIC_AILA_FEEDBACK_FORM_URL env var, so future changes are a Doppler edit. It's exported once from `beta-tag.tsx`, and I kept the current HubSpot link as a fallback so it doesn't break if the var isn't set. I also pointed the three FAQ feedback links at the same constant, since they were still on the old Google form.

JBR90 (approved):

mikeritson-oak (approved):
All looks good to merge. 👍

assadk88:
@JBR90 this is ready to merge into your branch

PR #1082, opened 2026-06-17
fix: upgrade posthog-js to 1.388.1

mantagen:
## Description

- Bump `posthog-js` from `^1.201.0` to `^1.388.1` in `apps/nextjs`.

uBlock Origin started blocking requests to `/flags` when `ip=` appears in the query string, regardless of domain. They have since fixed this on their side, but anyone on the **Lite** version of uBlock will still hit the issue.

Older versions of `posthog-js` included `ip=0` on every request, which trips this filter. Upgrading to `1.387.0+` removes `ip=0` on every request.

Context: [Slack thread](https://oaknationalacademy.slack.com/archives/C03SQBG3L78/p1781700733101719)

## Issue(s)

Fixes #

## How to test

1. Go to https://oak-ai-lesson-assistant-website-itdieqgux.vercel.thenational.academy
2. Open DevTools → Network and inspect a request to `/flags` (or `/decide`/`/array`).
3. Confirm `ip=0` is no longer appended to the query string, and that PostHog analytics / feature flags still load correctly with uBlock Origin enabled.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/nextjs/package.json
+++ b/apps/nextjs/package.json
@@ -109,7 +109,7 @@
     "openai": "^4.96.0",
     "p-limit": "^7.2.0",
     "partial-json-parser": "^1.2.2",
-    "posthog-js": "^1.201.0",
+    "posthog-js": "^1.388.1",
     "posthog-node": "^5.24.7",
     "pptxgenjs": "^3.12.0",
     "ramda": "^0.30.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -302,8 +302,8 @@ importers:
         specifier: ^1.2.2
         version: 1.2.2
       posthog-js:
-        specifier: ^1.201.0
-        version: 1.201.0
+        specifier: ^1.388.1
+        version: 1.388.1
       posthog-node:
         specifier: ^5.24.7
         version: 5.24.10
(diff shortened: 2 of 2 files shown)


assadk88 (approved):

mikeritson-oak (changes requested):
I've tsaken a look at this update and, generally, things seemed to be OK but a couple of questions came up during testing:

1. Production currently shows `ip=1` rather than the stated `ip=0`, most likely not an issue but double checking
2. On the preview branch, a query for the URL parameters as stated, flags/array/decide, brings back results which do not contain any reference to `ip=` but a query for `ip=` does return results to other paths/endpoints

Aside from these two points, testing with both ublock on and off showed events being recorded in Posthog.

Production shows the version as 1.201.0 (as expected) but includes the parameter `ip=1` rather than the stated `ip=0`
￼
Image

Preview with the filter `/flags|array|decide/` applied in the network tab, shows the version as 1.388.1 and non of the requests contain any reference to `ip=`

Image
￼
Preview with the filter `ip=` still shows the updated version but also with results for requests showing that include `ip=0` in the request URL. This includes some GET requests for surveys but primarily POST requests containing the /e/ path (possibly a posthog event endpoint?)


Image

￼

PR #1087, opened 2026-06-22
feat: targeted edits for keywords and misconceptions

assadk88:
## Description

Adds targeted single-item editing to the `keywords` and `misconceptions` sections. Previously, editing one item (e.g. "remove the second keyword") regenerated the whole section, so untouched items came back reworded or dropped. Both now route through the same generic dispatcher, so only the targeted items change and the rest stay byte-identical.

## Issue(s)

Fixes [AI-2219](https://app.notion.com/p/oaknationalacademy/Targeted-editing-for-list-based-lesson-sections-36626cc4e1b180539965f809c5611c02?source=copy_link)

## How to test

On the [deployment](https://oak-ai-lesson-assistant-website-git-feat-ai-2219-targete-c38da9.vercel.thenational.academy/), open a lesson with populated `keywords` and `misconceptions` sections (generate one if needed), then exercise each through the chat.

**Keywords**

1. Ask to remove the second keyword. Expect: only that keyword is removed; the others are unchanged, word for word.
2. Ask to add a keyword about a given topic. Expect: a new keyword is added and the existing ones are untouched.
3. Ask to rewrite the second keyword. Expect: only the second keyword changes; the rest are identical.

**Misconceptions**

1. Ask to remove the second misconception. Expect: only that misconception is removed; the others are unchanged, word for word.
2. Ask to add a misconception about a given topic, e.g. "Add a point about <...> to the misconceptions section". Expect: a new misconception is added and the existing ones are untouched.
3. Ask to rewrite the second misconception. Expect: only the second misconception changes; the rest are identical.

Edge cases (either section): asking to "remove a keyword" without naming which one replies asking which, and leaves the list unchanged; asking to remove the only remaining keyword or misconception is declined with a note.

Ensure there are no regressions to the previous quiz behaviour.

## Screenshots

N/A, no UI change.

## Checklist

- [x] Manually tested across browsers / devices
- [ ] ~~Considered impact on accessibility~~ (no UI change)
- [ ] ~~Does this PR update a package with a breaking change~~ (no package changes)


packages/aila/src/core/chat/PatchEnqueuer.ts
```diff
@@ -23,6 +40,22 @@ export class PatchEnqueuer {
...
   }
 
+  private safeEnqueue(encoded: string): void {
+    if (!this.controller) {
+      throw new Error("Controller not set");
+    }
+    try {
+      this.controller.enqueue(encoded);
+    } catch (error) {
+      if (isControllerClosed(error)) {
```
neil-hughes, line 52: Probably worth looking at `isControllerClosed` more closely as part of this PR.

It catches all TypeErrors, not just closed-controller ones. A TypeError thrown for an unrelated reason (e.g. a type coercion failure inside enqueue) would be silently swallowed as a disconnect, losing data without any error surface. 

Perhaps the message string "Invalid state: Controller is already closed" should also be checked inside `isControllerClosed`:

```
return (
  error instanceof TypeError &&
  (error.message.includes("Controller is already closed") ||
    (error as { code?: unknown }).code === "ERR_INVALID_STATE")
);
```

packages/aila/src/lib/agentic-system/quizOperations/quizOperationDispatcher.ts
```diff
@@ -1,133 +1,43 @@
...
-        ...questions.slice(insertAt - 1),
-      ],
+  const result = await sectionListOperationDispatcher<LatestQuizQuestion>(
+    currentQuiz.questions,
+    intent,
+    runSingleQuestion,
+    {
+      itemNoun: "question",
+      min: 0,
+      max: Number.POSITIVE_INFINITY,
```
neil-hughes: I feel for the pupils who are given an infinitely-long quiz as homework!

packages/aila/src/lib/agentic-system/execution/executePlanSteps.ts
```diff
@@ -183,6 +258,73 @@ async function executeQuizDispatchStep(
...
+async function executeSectionListDispatchStep(
+  context: AilaExecutionContext,
+  step: PlanStep,
+  sectionKey: SectionKey,
+  intent: StructuralItemIntent,
+  config: ListSectionConfig,
+): Promise<AilaTurnPhaseOutcome> {
+  const existing = context.currentTurn.document[sectionKey];
+  const sectionWasAbsent = existing == null;
+  const currentItems = (existing ?? []) as unknown[];
```
neil-hughes: Can we avoid the `unknown[]` casts here?

packages/aila/src/lib/agentic-system/execution/executePlanSteps.ts
```diff
@@ -183,6 +258,73 @@ async function executeQuizDispatchStep(
...
+      const agent = context.runtime.sectionAgents[agentId];
+      if (!agent) return null;
+      const agentResult = await agent.handler(context);
+      if (agentResult.error) return null;
+      const parsed = config.arraySchema.safeParse(agentResult.data);
+      if (!parsed.success) return null;
+      // The add/change agents must return exactly one item; if more come back,
+      // decline rather than silently using the first (which for a change could
+      // land an unrelated item in the target slot).
+      const data = parsed.data as unknown[];
```
neil-hughes: And here? Is it possible that ListSectionConfig could be narrowed to carry the actual Zod output type?

packages/aila/src/lib/agentic-system/execution/executePlanSteps.ts
```diff
@@ -183,6 +258,73 @@ async function executeQuizDispatchStep(
...
+      const item = data[0];
+      if (!config.validateItem(item)) return null;
+      // Correct only this new item, wrapped as a single-element section, so
+      // existing items are never sent to the corrector and stay untouched.
+      const correctedItems = (await applyBritishEnglishCorrection({
+        context,
+        sectionKey,
+        content: [item],
+        responseSchema: config.arraySchema,
+      })) as unknown[] | null;
```
neil-hughes: Again, curious if this is strictly a necessary `unknown` or if we can be a bit tighter.

neil-hughes (changes requested):
Overall looks really good, really like the more generic editing approach, and the code is clean and well-tested.

Two overall thoughts and some individual comments:

1. There's a potential test you may want to add as currently there's no test for REGENERATE_SECTION on keywords/misconceptions (i.e. when `itemIntent` is absent or action is REGENERATE_SECTION, verify it falls through to `runSectionAgent`). The quiz tests cover this for starterQuiz, so it's likely fine, but a brief test confirming the keywords/misconceptions fallthrough would make the contract explicit.

2. Is there any persisted data in production that still references `quizIntent`? i.e. if I load an old lesson generated with `quizIntent` is it still valid or does it error? If we do need to support this, then we should add some temporary migration code that allows for loading old persistent data as the new name is definitely more correct.

packages/aila/src/lib/agentic-system/quizOperations/quizOperationDispatcher.ts
assadk88: Ha, fair! It's Infinity because the old dispatcher never capped questions either, so this was just preserving that. Capping it is a behaviour change for the quiz path. Happy to do that if what's you want?

assadk88:
> Overall looks really good, really like the more generic editing approach, and the code is clean and well-tested.
> 
> Two overall thoughts and some individual comments:
> 
> 1. There's a potential test you may want to add as currently there's no test for REGENERATE_SECTION on keywords/misconceptions (i.e. when `itemIntent` is absent or action is REGENERATE_SECTION, verify it falls through to `runSectionAgent`). The quiz tests cover this for starterQuiz, so it's likely fine, but a brief test confirming the keywords/misconceptions fallthrough would make the contract explicit.
> 2. Is there any persisted data in production that still references `quizIntent`? i.e. if I load an old lesson generated with `quizIntent` is it still valid or does it error? If we do need to support this, then we should add some temporary migration code that allows for loading old persistent data as the new name is definitely more correct.

@neil-hughes There's no migration needed. quizIntent is never persisted: it only exists in memory while the planner runs a single turn, then is discarded. The one place quizIntent still exists is the e2e recordings (always quizIntent: null, which is also why e2e is passing atm). Replay ignores it since the code reads itemIntent and regenerates the whole section. I could re-record this as part of this PR or in a future one?

neil-hughes:
> > Overall looks really good, really like the more generic editing approach, and the code is clean and well-tested.
> > Two overall thoughts and some individual comments:
> > 
> > 1. There's a potential test you may want to add as currently there's no test for REGENERATE_SECTION on keywords/misconceptions (i.e. when `itemIntent` is absent or action is REGENERATE_SECTION, verify it falls through to `runSectionAgent`). The quiz tests cover this for starterQuiz, so it's likely fine, but a brief test confirming the keywords/misconceptions fallthrough would make the contract explicit.
> > 2. Is there any persisted data in production that still references `quizIntent`? i.e. if I load an old lesson generated with `quizIntent` is it still valid or does it error? If we do need to support this, then we should add some temporary migration code that allows for loading old persistent data as the new name is definitely more correct.
> 
> @neil-hughes There's no migration needed. quizIntent is never persisted: it only exists in memory while the planner runs a single turn, then is discarded. The one place quizIntent still exists is the e2e recordings (always quizIntent: null, which is also why e2e is passing atm). Replay ignores it since the code reads itemIntent and regenerates the whole section. I could re-record this as part of this PR or in a future one?

Good catch @assadk88 

I think it makes sense to update the associated e2e test data along with these changes so they're always in sync with the branch state - what do you think?

packages/aila/src/lib/agentic-system/quizOperations/quizOperationDispatcher.ts
assadk88: @neil-hughes should I cap this?

neil-hughes: I wondered that when I left the comment but can't think of what a more meaningful cap should be - if I remember rightly in practice this max doesn't really matter? If you think it would benefit from a cap of some sort then feel free to. Maybe 99 or something - a number way higher than would ever be practical but not so high that it would cause problems if we ever actually used it? 🤔

assadk88: Sounds good 👍

mikeritson-oak (approved):
No issues seen during testing. Looks good to merge. 👍

neil-hughes (approved):
Looks good, nice work @assadk88

PR #1100, opened 2026-07-02
fix: [AI-2313] only show content guidance when categories change

neil-hughes:
## Description

This PR fixes repeated Aila content-guidance banners by only showing a warning when the visible guidance category set changes.

## Issue(s)

Fixes #AI-2313

## How to test

1. Go to https://oak-ai-lesson-assistant-website-faps4bek4.vercel.thenational.academy
2. Generate a lesson in Aila chat which requires content guidance - you should see it as normal.
3. Further messages should not show the banner again, unless/until the categories change.
4. In more detail:

#### Scenarios to Verify

1. First content-guidance result shows a banner.
2. Repeated messages with the same guidance categories do not show another banner.
3. A later message with added, removed, or different guidance categories shows a new banner.
4. If a safe moderation happens between guidance results, the next guidance result shows a banner again - this is a rare situation in real usage, but something like "generate a lesson with a practical activity, so the `practical-activities` guidance appears, then remove the activity, then add it back again - the content guidance should reappear. 
5. Locking moderation (toxic or highly sensitive content) should still opens the blocking modal as before - no changes are expected for this behaviour.

### DIFF
--- a/apps/nextjs/src/components/AppComponents/Chat/chat-list/index.tsx
+++ b/apps/nextjs/src/components/AppComponents/Chat/chat-list/index.tsx
@@ -1,12 +1,16 @@
 "use client";
 
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+import type { PersistedModerationBase } from "@oakai/core/src/utils/ailaModeration/moderationSchema";
 
 import scrollIntoView from "scroll-into-view-if-needed";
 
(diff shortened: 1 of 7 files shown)


mikeritson-oak (approved):
Tested across a sample of Aila lessons on various topics. No issues seen with the moderation messaging.

Messages appeared when expected, were suppressed when there were no updates, and appeared again when the content changed. I also confirmed the usual sensitive, highly sensitive and toxic messaging and actions behaved as expected.

This looks good to merge. 👍

JBR90 (approved):

PR #1103, opened 2026-07-07
feat: [AI-2309] persist agentic generations with single-source prompt templates

neil-hughes:
## Description

Agentic Aila didn't record generations in the `Generation` table the way non-agentic Aila does. This PR adds that feature, plus the prompt-versioning plumbing it needs.

### What it does

- Persists **one `Generation` row per agent LLM call** (planner, each section agent, corrector, messageToUser), captured at the single execution choke point and written in one batch at the end of a turn.
- Each row links to a **real `prompts.id`**, resolved from a per-agent prompt template. Quiz agents resolve a distinct template per build mode (`fullRegen` / `addOne` / `rewriteOne`).
- Skipped in fixture mode; no-ops safely when there's no `userId`.

### Notes

- Added a **single source of truth for prompts.** Each agent's static prompt (instructions + voice) is defined once in `AGENTIC_PROMPT_TEMPLATES` and rendered by a shared `staticPromptParts` builder — used by *both* the runtime prompt and the stored/versioned template, so they can't drift. (This fixed a latent bug where 10/14 section templates recorded the wrong voice.)
- **Content-stable versioning.** A new prompt version is created only when a template's content actually changes, not on every deploy (opt-in `stableAcrossDeploys` flag; non-agentic keeps its per-deploy behaviour).
- **Concurrency/rollback-safe.** Prompt creation is now an upsert on the unique hash, so concurrent requests and rollbacks re-use the existing row instead of hitting a unique-constraint violation.

### Reviewer note

Touches shared core (`packages/core/.../promptVariants.ts`). Non-agentic behaviour is unchanged in the normal path — the hashing is byte-identical (the new flag defaults off) and the upsert only changes a path that previously *threw*. Split into its own commit for isolatability.

@JBR90 It would be useful for you to review this branch locally and inspect the generated contents in the `prompts` and `generations` tables matches your expectation across multiple key stages and activities like adding/removing quiz questions.

### Testing
- Manually verified: prompts table populated on first run, no duplicates on re-run, a new version appears when a prompt changes (and reverts cleanly back to the prior version), generations persist and link to the correct current prompt.
- Automated: full agentic + persistence unit tests pass. All prompt snapshots were unchanged, which confirms runtime prompts are identical after the refactor.

## Issue(s)

Fixes #AI-2309

## How to test

This PR is mostly invisible to users as it is focused on recording rows in the database. However, it touched a lot of code and involved re-visiting some recent merges so regression tests on Aila chat would still be useful:

1. Go to https://oak-ai-lesson-assistant-website-ollauerub.vercel.thenational.academy
2. Generate a full lesson end-to-end in both agentic/non-agentic modes: the result should be within the normal range of outputs
3. Generate a full lesson end-to-end for a _different key stage_ in both agentic/non-agentic modes and verify the output appears within the normal range
4. Generate a starter/exit quiz - you should see valid questions as normal.
5. Add/change/rewrite/remove questions should still work as before


mikeritson-oak (approved):
Checked as per testing instructions, plus a few general tests, and no issues seen.
Looks good to merge. 👍

packages/aila/src/lib/agentic-system/agents/sectionAgents/createSectionAgent.ts
```diff
@@ -4,13 +4,101 @@ import type { ResponseCreateParamsNonStreaming } from "openai/resources/response
...
+      case "addOne":
+        return {
+          text: instructionsByMode.addOne(keyStage),
+          variant: `addOne:${normalisedKeyStage}`,
+          promptInputs: { keyStage, buildMode: "addOne" },
+        };
+      case "rewriteOne":
+        return {
+          text: instructionsByMode.rewriteOne(mode.position, keyStage),
+          variant: `rewriteOne:${mode.position}:${normalisedKeyStage}`,
```
JBR90: Do we need a new variant for each re-write position ?

neil-hughes: Good catch @JBR90 - my initial version of this treated position as a parameter, but in the complex merge with the prompt caching branch I lost track of this nuance. Will re-add it now.

assadk88:
LGTM:

- Two lessons at once: each prompt saved once, nothing crashed
- Interrupting a lesson: refreshed mid-generation, finished agent calls were still saved, and the lesson carried on after reload
- Starter quiz prompt edit and revert: editing the prompt created v2 with generations linked to it. Reverting made v1 current again with no duplicate row or crash
- Exit quiz edits: add and rewrite each save under their own prompt; rewriting Q2 vs Q5 reuses the same one. Deleting makes no LLM call so nothing is saved.
- Generation to message links: every generation row pointed at the right assistant message across a six-turn lesson
- Non-agentic Aila works as before, writes nothing new to the prompts table

PR #1129, opened 2026-08-20
fix: prevent automatic preload on homepage videos

neil-hughes:
## Description

- Set `preload="none"` on all Mux players (via `.attrs` on the shared `StyledMuxPlayer`, so it applies to all three and any future ones)
- Render the hero player once instead of one copy per breakpoint: the mobile copy sat behind `display: none`, which still mounts and still downloads

Together these stop video being fetched for visitors and bots who never press play. Expect Mux views to roughly halve and the fatal error rate to drop from ~14.5% toward ~1%: both intended, not a regression.

## Issue(s)

Fixes #ADAPT-60

## How to test

1. Go to https://oak-ai-lesson-assistant-website-1c6uoe01c.vercel.thenational.academy
2. Open Network tab, filter for `stream.mux`
3. You should see no activity until pressing play


mantagen (approved):
LGTM

PR #1131, opened 2026-08-28
feat: add a PostHog-switchable service status banner

assadk88:
## Description

Adds a service status banner that can be switched on from PostHog during an incident, so nobody has to deploy to tell users Aila is having problems. Its wording comes from the flag's JSON payload, which meant extending the flag bootstrap (`getAllFlags` becomes `getAllFlagsAndPayloads`) rather than reading the flag client-side: `posthog-js` never initialises for users who declined cookie consent. Existing flag values are returned unchanged; the banner renders nothing until `aila-status-banner` is enabled, and that flag is currently off in both staging and production.

## Issue(s)

Fixes #

## How to test

1. Go to https://oak-ai-lesson-assistant-website-qbbnoigkt.vercel.thenational.academy
2. You should see no banner. The flag is off by default.
3. In PostHog, enable the `aila-status-banner` flag for this project. Allow up to 60 seconds for the flag definition cache to refresh.
4. Reload. You should see a strip at the top of the page reading "We're currently experiencing some issues, see our status page for more information", with `our status page` as a link.
5. Check it also appears on `/aila`, `/aila/teaching-materials` and `/sign-in`.
6. Click the link. It should open `https://status.thenational.academy/` in a new tab.
7. Edit the flag's payload in PostHog to change the wording, wait 60 seconds, reload. You should see the new wording with no deploy.
8. Switch the flag off and reload. The banner should disappear.

Edge cases worth checking separately:

9. With the flag on, decline cookie consent, then reload. The banner should still appear. This is the reason the flag is read server-side rather than in the browser, so it is the most valuable check here.
10. Put a deliberate mistake in the payload JSON (for example change `href` to `not-a-url`), reload. You should see the default message rather than no banner at all.
11. Tab to the link with the keyboard. It should take focus with a visible focus indicator, and the banner should not obscure it.

[Posthog feature flag page](https://eu.posthog.com/project/8837/feature_flags/261563)

## Screenshots

N/A

How it should now look:
image

## Checklist

- [ ] Considered impact on accessibility (`role="status"` with `aria-live="polite"`)
- [ ] ~~Does this PR update a package with a breaking change~~ (no dependency changes)


mikeritson-oak (changes requested):
Tested across Chrome, Safari, Edge and Firefox - the banner message and link displayed correctly.

I am however seeing some inconsistency with the feature flag itself. Across changes and page refreshes I sometimes get the configured custom message and sometimes what appears to be the generic fallback message. I’ve also had the flag disabled in PostHog for some time and continued to see the banner after refreshing in different browsers.

So the banner functionality looks good, but the reliability/propagation of the PostHog flag seems worth investigating. Possibly just aa quirk of the feature flags but may be related to how this particular one has been implemented.

neil-hughes:
@mikeritson-oak We can re-test this now - should be more reliable with the Posthog flags.

Worth noting that before we merge this we should make sure the flag is left in a valid state and off in production.

neil-hughes (approved):

mikeritson-oak (approved):
Looks good to release. 👍 
Feature flags and payloads, on both Staging and Production, are as per original deployment and toggled to 'off' until needed.

